PR #29, opened 2020-09-07
Added JSON parsing, changed plot and added scientific notation for plots in eemumu_AV

TaranSinghania:
For running on consumer cards:

1. Needed to use `#define MGONGPU_FPTYPE_FLOAT 1` in `madgraph4gpu/examples/gpu/eemumu_AV/src`.

2. Matching the CUARCHNUM with the nvidia architecture being used. (In my case: Pascal -> sm_61)

3. Tiny error in `madgraph4gpu/examples/gpu/eemumu_AV/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc`. In line 466: needed to change `isnan` to `std::isnan`.

4. Added another command line argument flag `--json|-j` which accepts the `date` and `run` and sends the output to a json located in `madgraph4gpu-1/examples/gpu/eemumu_AV/SubProcesses/P1_Sigma_sm_epem_mupmum/perf/data`. This will make it easier to parse while generating graphs. This is similar to how it is in `madgraph4gpu/examples/gpu/eemumu` and the files will be named similarly as well.

An example would be `./gcheck.exe -p 64 32 1 -j 20200404 1` and this would generate `20200404-perf-test-run1.json`.
If you run the same command again, the information will be appended to the json file. 

5. The file `20200404-perf-test-run1.json` was generated by running:
 `./gcheck.exe -p 128 64 1 -j 20200404 1`
 `./gcheck.exe -p 64 32 1 -j 20200404 1`

In eemumu, the `readdata.py` which was used to parse the output is now redundant in eemumu_AV, there is an extra function defined in `Class perf()` in the `madgraph4gpu/examples/gpu/eemumu_AV/SubProcesses/P1_Sigma_sm_epem_mupmum/perf/perf.py` file to parse the values containing "sec" or "GeV" to float.

6. The y axis was changed from `TotalTimeInWaveFuncs` to `TotalTimeInWaveFuncs/TotalEventsComputed` and tthe scientific notation was added as well.

Do let me know if any changes are to be made :)


valassi:
Hi @TaranSinghania thanks a lot for your patches!
I am just resuming work on this project now so I am only going through your changes now.

One question: are date and run optional or mandatory when adding "-j"? Presently they are effectively optional, so if you run "-j" without adding date and run, this creates 0-perf-test-run0.json (taking date=0 and run=0 - where actually these variables are not initialised). If (as I imagine?) date and run should actually be mandatory, can you make them mandatory in the code?

Anyway, for the moment I just added the initialisation of date/run, I also changed names to jsondate and jsonrun to make it more explicit. See https://github.com/madgraph5/madgraph4gpu/commit/d1e29cada8eb953f94903dad3b6b7ccfc8376166

Thanks!
Andrea

TaranSinghania:
Hi @valassi , thanks for bringing it up! :)
Alternatively, would you want the date to be the current date and only the run to be optional?

PR #37, opened 2020-10-19
adding Kokkos Port

jtchilders:
I've added a port of the gg_ttgg_auto process to Kokkos.


examples/kokkos/gg_ttgg_auto/src/HelAmps_sm.cu
```diff
@@ -0,0 +1,808 @@
...
+#include <complex> 
+#include <cmath> 
+#include <iostream> 
+#include <cstdlib> 
+#include <thrust/complex.h> 
+
+namespace MG5_sm 
+{
+
+__device__ void ixxxxx(constdouble * pvec, double fmass, int nhel, int nsf,
```
tjburch, line 18: I was being nosy and looking through your conversion as I was working through mine, noticed `constdouble` here is probably a typo

PR #42, opened 2020-11-02
Reorder timer steps: move stats calculation upfront

valassi:


examples/gpu/eemumu_AV/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
```diff
@@ -318,21 +318,17 @@ int main(int argc, char **argv)
     // 3b. Evaluate MEs on the device
     // 3c. Copy MEs back from device to host
 
-
```
lfield, line 321: Was the removal of this blank line intentional?

examples/gpu/eemumu_AV/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
```diff
@@ -590,130 +594,120 @@ int main(int argc, char **argv)
...
+    }
+    else{
+      //deleting the last bracket and outputting a ", "
+      std::string temp = "truncate -s-1 " + perffile;
+      const char *command = temp.c_str();
+      system(command);
+      jsonFile << ", " << std::endl;
+    }
 
-      jsonFile << "{" << std::endl
```
lfield, line 624: Looks like just the indention changed.

valassi: Yes this is one of the main changes, now the "if(json)" is at the same level as, and outsode of, "if(perf)", while it was inside before. The logic of the json flag was not enforced very clearly (it needed -p too to do something useful). Now I redefined -p to print to screen and -j to dump to json, but the stats are computed all the time (previosuly they were computed only within "if(perf)").

examples/gpu/eemumu_AV/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
valassi: Yes, Doing some aesthetic cleanup (my own personal definition). At some point it would be nice to have som ecoding styles recommended, or even enforced by tools. For the moment it's "have a look every now and then and make it look good"!

PR #47, opened 2020-11-10
[eemumu_AV] Use unique_ptr and RAII to automate memory management.

hageboeck:
I wanted to test if memory / resource management can be automated. With unique_ptr and a few helper functions, a lot of `ifdef` and `delete[] / free`s can be removed. Is this helpful for us?

### DIFF
--- a/examples/gpu/eemumu_AV/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
+++ b/examples/gpu/eemumu_AV/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
@@ -4,6 +4,7 @@
 #include <fstream>
 #include <iomanip>
 #include <iostream>
+#include <memory>
 #include <numeric>
 #include <string>
 #include <unistd.h>
@@ -40,6 +41,36 @@ int usage(char* argv0, int ret = 1) {
   std::cout << "The '-d' flag only controls if nan's emit warnings" << std::endl;
   return ret;
 }
+
+#ifdef __CUDACC__
+template<typename T = fptype>
+struct CudaDevDeleter {
+  void operator()(T* mem) {
+    checkCuda( cudaFree( mem ) );
+  }
+};
+template<typename T = fptype>
+std::unique_ptr<T, CudaDevDeleter<T>> devMakeUnique(std::size_t N) {
+  T* tmp = nullptr;
+  checkCuda( cudaMalloc( &tmp, N * sizeof(T) ) );
+  return std::unique_ptr<T, CudaDevDeleter<T>>{ tmp };
+}
+template<typename T = fptype>
+struct CudaHstDeleter {
+  void operator()(T* mem) {
+    checkCuda( cudaFreeHost( mem ) );
+  }
+};
+template<typename T = fptype>
+std::unique_ptr<T[], CudaHstDeleter<T>> hstMakeUnique(std::size_t N) {
+  T* tmp = nullptr;
+  checkCuda( cudaMallocHost( &tmp, N * sizeof(T) ) );
+  return std::unique_ptr<T[], CudaHstDeleter<T>>{ tmp };
+};
+#else
+template<typename T = fptype>
+std::unique_ptr<T[]> hstMakeUnique(std::size_t N) { return std::unique_ptr<T[]>{ new T[N] }; };
+#endif
 
 int main(int argc, char **argv)
 {
@@ -131,6 +162,16 @@ int main(int argc, char **argv)
   //std::cout << "Calling cudaFree... " << std::endl;
   checkCuda( cudaFree( 0 ) ); // SLOW!
   //std::cout << "Calling cudaFree... done" << std::endl;
+
+  // --- Book the tear down at the end of main:
+  struct CudaTearDown {
+    CudaTearDown(bool print) : _print(print) { }
+    ~CudaTearDown() {
+      if ( _print ) std::cout << "Calling cudaDeviceReset()." << std::endl;
+      checkCuda( cudaDeviceReset() ); // this is needed by cuda-memcheck --leak-check full
+    }
+    bool _print{false};
+  } cudaTearDown(debug);
 #endif
 
   // --- 0a. Initialise physics process
@@ -159,71 +200,41 @@ int main(int argc, char **argv)
   using mgOnGpu::np4;
   using mgOnGpu::nparf;
   using mgOnGpu::npar;
+  using mgOnGpu::ncomb; // Number of helicity combinations
   const int nRnarray = np4*nparf*nevt; // (NB: ASA layout with nevt=npagR*neppR events per iteration)
-
-#ifdef __CUDACC__
-  const int nbytesRnarray = nRnarray * sizeof(fptype);
-  fptype* devRnarray = nullptr; // AOSOA[npagR][nparf][np4][neppR] (NB: nevt=npagR*neppR)
-  checkCuda( cudaMalloc( &devRnarray, nbytesRnarray ) );
-#if defined MGONGPU_CURAND_ONHOST or defined MGONGPU_COMMONRAND_ONHOST
-  fptype* hstRnarray = nullptr; // AOSOA[npagR][nparf][np4][neppR] (NB: nevt=npagR*neppR)
-  checkCuda( cudaMallocHost( &hstRnarray, nbytesRnarray ) ); // pinned memory
-#endif
-#else
-  fptype* hstRnarray = nullptr; // AOSOA[npagR][nparf][np4][neppR] (NB: nevt=npagR*neppR)
-  hstRnarray = new fptype[nRnarray]();
-#ifdef MGONGPU_COMMONRAND_ONHOST
-  const int nbytesRnarray = nRnarray * sizeof(fptype);
-#endif
-#endif
-
   const int nMomenta = np4*npar*nevt; // (NB: nevt=npagM*neppM for ASA layouts)
-  fptype* hstMomenta = nullptr; // AOSOA[npagM][npar][np4][neppM] (previously was: lp)
-#ifdef __CUDACC__
-  const int nbytesMomenta = nMomenta * sizeof(fptype);
-  checkCuda( cudaMallocHost( &hstMomenta, nbytesMomenta ) );
-  fptype* devMomenta = nullptr; // (previously was: allMomenta)
-  checkCuda( cudaMalloc( &devMomenta, nbytesMomenta ) );
-#else
-  hstMomenta = new fptype[nMomenta]();
-#endif
-
-#ifdef __CUDACC__
-  using mgOnGpu::ncomb;
-  const int nbytesIsGoodHel = ncomb * sizeof(bool);
-  bool* hstIsGoodHel = nullptr;
-  checkCuda( cudaMallocHost( &hstIsGoodHel, nbytesIsGoodHel ) );
-  bool* devIsGoodHel = nullptr;
-  checkCuda( cudaMalloc( &devIsGoodHel, nbytesIsGoodHel ) );
-#endif
-
   const int nWeights = nevt;
-  fptype* hstWeights = nullptr; // (previously was: meHostPtr)
+  const int nMEs     = nevt;
+
+#if defined MGONGPU_CURAND_ONHOST or defined MGONGPU_COMMONRAND_ONHOST or not defined __CUDACC__
+  auto hstRnarray   = hstMakeUnique<fptype>( nRnarray ); // AOSOA[npagR][nparf][np4][neppR] (NB: nevt=npagR*neppR)
+#endif
+  auto hstMomenta   = hstMakeUnique<fptype>( nMomenta ); // AOSOA[npagM][npar][np4][neppM] (previously was: lp)
+  auto hstIsGoodHel = hstMakeUnique<bool  >( ncomb );
+  auto hstWeights   = hstMakeUnique<fptype>( nWeights ); // (previously was: meHostPtr)
+  auto hstMEs       = hstMakeUnique<fptype>( nMEs ); // (previously was: meHostPtr)
+
 #ifdef __CUDACC__
+  auto devRnarray   = devMakeUnique<fptype>( nRnarray ); // AOSOA[npagR][nparf][np4][neppR] (NB: nevt=npagR*neppR)
+  auto devMomenta   = devMakeUnique<fptype>( nMomenta ); // (previously was: allMomenta)
+  auto devIsGoodHel = devMakeUnique<bool  >( ncomb );
+  auto devWeights   = devMakeUnique<fptype>( nWeights ); // (previously was: meDevPtr)
+  auto devMEs       = devMakeUnique<fptype>( nMEs ); // (previously was: meDevPtr)
+
+#if defined MGONGPU_CURAND_ONHOST or defined MGONGPU_COMMONRAND_ONHOST
+  const int nbytesRnarray = nRnarray * sizeof(fptype);
+#endif
+  const int nbytesMomenta = nMomenta * sizeof(fptype);
+  const int nbytesIsGoodHel = ncomb * sizeof(bool);
   const int nbytesWeights = nWeights * sizeof(fptype);
-  checkCuda( cudaMallocHost( &hstWeights, nbytesWeights ) );
-  fptype* devWeights = nullptr; // (previously was: meDevPtr)
-  checkCuda( cudaMalloc( &devWeights, nbytesWeights ) );
-#else
-  hstWeights = new fptype[nWeights]();
-#endif
-
-  const int nMEs = nevt;
-  fptype* hstMEs = nullptr; // (previously was: meHostPtr)
-#ifdef __CUDACC__
   const int nbytesMEs = nMEs * sizeof(fptype);
-  checkCuda( cudaMallocHost( &hstMEs, nbytesMEs ) );
-  fptype* devMEs = nullptr; // (previously was: meDevPtr)
-  checkCuda( cudaMalloc( &devMEs, nbytesMEs ) );
-#else
-  hstMEs = new fptype[nMEs]();
 #endif
 
-  double* genrtimes = new double[niter]();
-  double* rambtimes = new double[niter]();
-  double* wavetimes = new double[niter]();
-  fptype* matrixelementALL = new fptype[nevtALL](); // FIXME: assume process.nprocesses == 1
-  fptype* weightALL = new fptype[nevtALL]();
+  std::unique_ptr<double[]> genrtimes( new double[niter] );
+  std::unique_ptr<double[]> rambtimes( new double[niter] );
+  std::unique_ptr<double[]> wavetimes( new double[niter] );
+  std::unique_ptr<fptype[]> matrixelementALL( new fptype[nevtALL] ); // FIXME: assume process.nprocesses == 1
+  std::unique_ptr<fptype[]> weightALL( new fptype[nevtALL] );
 
   // --- 0c. Create curand or common generator
   const std::string cgenKey = "0c GenCreat";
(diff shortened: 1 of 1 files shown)


roiser:
In general I like the approach as it reduces the numbers of ifdefs. I have two questions on the CudaSetUpTearDown struct. 

Where is it initialised? 

If its initialised it will go out of scope probably only at the very end of the main. But the idea is to also dump the information how long the reset took in the     timermap.dump(); (very bottom of the main). Also the numbering would change 9c <-> 9d but that's a minor one.

Not even sure we need to time the cuda reset? @valassi ?

hageboeck:
> In general I like the approach as it reduces the numbers of ifdefs. I have two questions on the CudaSetUpTearDown struct.
> 
> Where is it initialised?
> 
Directly where it is declared:
https://github.com/hageboeck/madgraph4gpu/blob/7a7b0f27285cb08a0b946cfe149d8523d7e02ae6/examples/gpu/eemumu_AV/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc#L182

> If its initialised it will go out of scope probably only at the very end of the main. But the idea is to also dump the information how long the reset took in the timermap.dump(); (very bottom of the main). Also the numbering would change 9c <-> 9d but that's a minor one.

Very correct. I thought that once you saw it clean up once, you have seen it all.

It can obviously be reverted by
-[ ] Reverting to explicit manual cleanup.
-[ ] Putting the cuda part into a block.
-[ ] By resetting a resource manager, e.g. `cudaDeviceManager = nullptr`, `cudaDeviceManager.reset()` or similar.

I did it because it's exactly the same logic as the memory management, and cuda the frees *after* a device reset just lead to a crash. I've grown to love these `RAII`s, because you never have to worry about "forgot to clean up" or "what was the cleanup order again?".
Please choose one of the above. Warning, though, that if you want revert to explicit manual we need to kill all device memory before we kill the device. If the block + indent everything is OK, that's the easiest way, but it will create ugly diffs.

roiser:
> > In general I like the approach as it reduces the numbers of ifdefs. I have two questions on the CudaSetUpTearDown struct.
> > Where is it initialised?
> 
> Directly where it is declared:
> https://github.com/hageboeck/madgraph4gpu/blob/7a7b0f27285cb08a0b946cfe149d8523d7e02ae6/examples/gpu/eemumu_AV/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc#L182
> 

ah sorry didn't see that, ok

> > If its initialised it will go out of scope probably only at the very end of the main. But the idea is to also dump the information how long the reset took in the timermap.dump(); (very bottom of the main). Also the numbering would change 9c <-> 9d but that's a minor one.
> 
> Very correct. I thought that once you saw it clean up once, you have seen it all.
> 
> It can obviously be reverted by
> 
>     * [ ]  Reverting to explicit manual cleanup.
> 
>     * [ ]  Putting the cuda part into a block.
> 
>     * [ ]  By resetting a resource manager, e.g. `cudaDeviceManager = nullptr`, `cudaDeviceManager.reset()` or similar.
> 
> 
> I did it because it's exactly the same logic as the memory management, and cuda frees _after_ a device reset just lead to a crash. I've grown to love these `RAII`s, because you never have to worry about "forgot to clean up" or "what was the cleanup order again?".
> Please choose one of the above. Warning, though, that if you want revert to explicit manual we need to kill all device memory before we kill the device. If the "block + indent everything" is OK, that's the easiest way, but it will create ugly diffs.

I guess another way of solving it would be to put the dump into the destructor of the struct, then it's guaranteed that it will always go last (as long as there is only one such struct).

valassi:
Thanks @hageboeck for requesting a review.

My personal taste: some of these things could be useful, some look like an overhead. And there are a few subtleties to keep in mind.
- I would avoid CudaSetUpTearDown, simply because cudaFree(0) is not RAII where cudaFree is initialization and devicereset is termination. You could actually have only the first call, or only the second, or neither. It is not a new/delete. So I would not use that.
- About the hstMakeUnique, I do not like the fact that CUDA only has cudaMallocHost and C++ only has new. You can ALSO have new in CUDA. The point is that you have a choice whether you want pinned memory or not. In case, I think that hstMakeUnique should use new in both, while if you want cudaMallocHost you could have a second additional hstMakeUniquePinned, for instance.
- The destroy generator, I am really not sure. It's cool to also have a deleter, but in the end you are just adding the deletion in the line immediately after create generator, rather than at the end of the file. If you really want to have a 'unique pointer' type of design, somehoe you should encapsulate the create generator call within a unique pointer type of object for a generator instead. As it is now, I prefer to keep create generator and an explicit destroy generator, I find that easier to read and understand.
- In general instead I agree with replacing all those explicit free at the end by unique pointers, for the simple array type of objects (whether they are dev or host, and new or cudamalloc or cudamallochost).

So all in all, nice suggestion, but I would keep only the unique pointers for simple arrays, and would make sure there are two separate functions for cudamallochost and new on the host in CUDA.

valassi:
PS Sorry forgot to mention, this has now some conflicts after I merged PR #48

valassi:
Hi both, thanks for the discussion. Just for the record and so I do not forget, @hageboeck pointed out that the problem with adding unique pointers and not adding a struct with a destructor calling cudaDeviceReset is that cudaDeviceReset must be called AFTER all cudaMalloc. A code like
```
  cudaDeviceReset();
}
```
would instead first call cudaDeviceReset() and later call the destructors of unique pointers that go out of scope. I am ok to keep Stephan's suggested struct then (with or without cudaFree).

As discussed, this is any case temporary playground code, not production code. Eventually the memory allocations will be in MadGraph methods called by main, rather than by main. Another alternative I could suggest is to restructure check.cc so that it actually contains something like
```
int MG_main( argc, argv, timermap )
{
  // do everything, including memory allocations
}
int main( argc, argv )
{
    mgOnGpu::TimerMap timermap;
    int status = MG_main( argc, argv, timermap );
    cudaDeviceReset();
    return status;
}
```
but it is quite ugly. I do not mind, lets keep CudaSetUpTearDown as it is and move on with adding the missing functionality to the rest of the code.

hageboeck:
@roiser @valassi 
New version is up. I *think* I addressed everything we talked about.

examples/gpu/eemumu_AV/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
```diff
@@ -41,6 +42,38 @@ int usage(char* argv0, int ret = 1) {
...
+  return std::unique_ptr<T, CudaDevDeleter<T>>( tmp );
+}
+template<typename T = fptype>
+struct CudaHstDeleter {
+  void operator()(T* mem) {
+    checkCuda( cudaFreeHost( mem ) );
+  }
+};
+template<typename T = fptype>
+std::unique_ptr<T[], CudaHstDeleter<T>> hstMakeUnique(std::size_t N) {
```
valassi, line 65: I would still like to see hstMakeUnique and hstMakeUniquePinned, the former without cudaMallocHost, the latter with cudaMallocHost

examples/gpu/eemumu_AV/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
```diff
@@ -160,70 +200,42 @@ int main(int argc, char **argv)
...
-  fptype* hstRnarray = nullptr; // AOSOA[npagR][nparf][np4][neppR] (NB: nevt=npagR*neppR)
-  checkCuda( cudaMallocHost( &hstRnarray, nbytesRnarray ) ); // pinned memory
-#endif
-#else
-  fptype* hstRnarray = nullptr; // AOSOA[npagR][nparf][np4][neppR] (NB: nevt=npagR*neppR)
-  hstRnarray = new fptype[nRnarray]();
-#ifdef MGONGPU_COMMONRAND_ONHOST
-  const int nbytesRnarray = nRnarray * sizeof(fptype);
-#endif
+  auto devRnarray = devMakeUnique( nRnarray ); // AOSOA[npagR][nparf][np4][neppR] (NB: nevt=npagR*neppR)
```
valassi, line 203: This I would keep protecting with #ifdef CUDA

examples/gpu/eemumu_AV/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
```diff
@@ -160,70 +200,42 @@ int main(int argc, char **argv)
...
+#if defined MGONGPU_CURAND_ONHOST or defined MGONGPU_COMMONRAND_ONHOST or not defined __CUDACC__
+  auto hstRnarray = hstMakeUnique( nRnarray ); // AOSOA[npagR][nparf][np4][neppR] (NB: nevt=npagR*neppR)
```
valassi, line 205: Here I would still make the explicit distinction between hstMakeUnique in c++ and hstMakeUniquePinned in CUDA

examples/gpu/eemumu_AV/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
```diff
@@ -274,18 +286,18 @@ int main(int argc, char **argv)
     const std::string rngnKey = "1b GenRnGen";
     timermap.start( rngnKey );
 #ifdef MGONGPU_COMMONRAND_ONHOST
-    std::vector<double> commonRandomNumbers = commonRandomPromises[iiter].get_future().get();
-    assert( nRnarray == static_cast<int>(commonRandomNumbers.size()) );
+    std::vector<double> commonRnd = commonRandomPromises[iiter].get_future().get();
+    assert( nRnarray == static_cast<int>(commonRnd.size()) );
     // NB (PR #45): memcpy is strictly needed only in CUDA (copy to pinned memory), but keep it also in C++ for consistency
-    memcpy( hstRnarray, commonRandomNumbers.data(), nbytesRnarray );
+    std::copy( commonRnd.begin(), commonRnd.end(), hstRnarray.get() );
```
valassi, line 292: any particular reason why std::copy is better? I think memcpy makes it more obvious that we know we are copying a memy block of adjacent stuff... but maybe I am wrong

examples/gpu/eemumu_AV/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
```diff
@@ -548,43 +560,6 @@ int main(int argc, char **argv)
...
-
-#ifdef __CUDACC__
-  // --- 9c. Finalise cuda
-  const std::string cdrsKey = "9c CudReset";
-  timermap.start( cdrsKey );
-  checkCuda( cudaDeviceReset() ); // this is needed by cuda-memcheck --leak-check full
-#endif
-
   // --- 9d Dump to screen
   const std::string dumpKey = "9d DumpScrn";
```
valassi, line 564: 9d and 9e muct change to 9b and 9c.

examples/gpu/eemumu_AV/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
```diff
@@ -7,6 +7,7 @@
 #include <numeric>
 #include <string>
 #include <unistd.h>
+#include <memory>
```
valassi, line 10: personally I like to have headers ordered alphabetically, it makes the code more tidy

examples/gpu/eemumu_AV/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
valassi: Hm maybe I take this back now. I see that you are using hstMakeUnique here while in line 234 below you use explicitly a unique pointer. Looks better like you have done it, sorry for the noise. We should just remember to call this only for stuff that becomes pinned memory in cuda. Not sure, maybe we will have a mixture of pinned and not pinned in the future? But fo rthe momet ok like you are doing it.

examples/gpu/eemumu_AV/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
```diff
@@ -41,6 +42,38 @@ int usage(char* argv0, int ret = 1) {
...
+  }
+};
+template<typename T = fptype>
+std::unique_ptr<T[], CudaHstDeleter<T>> hstMakeUnique(std::size_t N) {
+  T* tmp = nullptr;
+  checkCuda( cudaMallocHost( &tmp, N * sizeof(T) ) );
+  return std::unique_ptr<T[], CudaHstDeleter<T>>( tmp );
+};
+#else
+template<typename T = fptype>
```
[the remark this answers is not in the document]
valassi: This I confirm, I'd remove devMakeUnique in c++, it has no meaning

examples/gpu/eemumu_AV/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
```diff
@@ -41,6 +42,38 @@ int usage(char* argv0, int ret = 1) {
...
+std::unique_ptr<T> devMakeUnique(std::size_t) { return nullptr; };
+template<typename T = fptype>
+std::unique_ptr<T[]> hstMakeUnique(std::size_t N) { return std::unique_ptr<T[]>( new T[N] ); };
```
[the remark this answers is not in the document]
valassi: Ok I take this one back too. Ok to have two different definitions of hstMakeUnique in cuda and c++. Until/unless one day we need to mix both pinned and unpineed memory for the data arrays that we pass in between kernels.

examples/gpu/eemumu_AV/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
```diff
@@ -131,6 +164,13 @@ int main(int argc, char **argv)
   //std::cout << "Calling cudaFree... " << std::endl;
   checkCuda( cudaFree( 0 ) ); // SLOW!
   //std::cout << "Calling cudaFree... done" << std::endl;
+
+  // --- Book the tear down at the end of main:
+  struct CudaTearDown {
+    ~CudaTearDown() {
+      checkCuda( cudaDeviceReset() ); // this is needed by cuda-memcheck --leak-check full
```
valassi, line 171: Maybe add an std::cout commented out about calling this cudadevicereset, may be useful for later debugging

examples/gpu/eemumu_AV/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
valassi: I take this back as discussed above

valassi (changes requested):
Thanks @hageboeck 

The main thing I suggest to change is to remove all appearance of devMakeUnique on c++ rathe than define a no-op one.

This morning we had discussed having hstMakeUnique and hstMakeUniquePinned, which are still not there. Initially I commented that we should have them, but then I took it back. So it's ok to leave these as they are, sorry for some hesitations...

Then the other are minor suggestions, but have a look, eg alphabetically ordered headers, why not memcpy, and add a printout on cudadevicereset.

Thanks! Andrea

examples/gpu/eemumu_AV/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
hageboeck: Of course. :-)
`nbytesRnarray` triggers a warning if specific branches of the `ifdef`s are active. It's because it was used both for cudaMemcpy and the on-host copy.
I could think of three ways to solve it:
- cast it to void in all branches where it is defined but not used. (Not nice)
- Hide behind more ifdefs. (I try to avoid as many ifdefs as I can, because you see errors only when you reconfigure. In fact, I saw the warning only when I did my last checks and cycled through the three options. I found that annoying / unsafe.)
- Use std::copy for the host copy, and cudamemcpy for a cuda memcopy. This way, `nbytesRnarray` is only used in one place.

Option 3 seemed the most obvious in C++

valassi: ah :-) not obvious, I thought you preferred std::copy to memcpy for some other reasons

I would go for option 2, which is what was used in the previous code and was working fine without any error or warning - ie redefine rnsrray excatly in the same way behind another ifdef

i do not see any problem with having pieces of code used only when the relevant define is defined.... this is the same we discussed this morning

examples/gpu/eemumu_AV/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
hageboeck: Let's make a deal:
I can order headers alphabetically if we agree to order includes from specific to general.

https://blog.knatten.org/2010/07/01/the-order-of-include-directives-matter/

examples/gpu/eemumu_AV/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
hageboeck: > ah :-) not obvious, I thought you preferred std::copy to memcpy for some other reasons

I do that as well :-)
It doesn't ask for a size, and it checks that the types are compatible.

examples/gpu/eemumu_AV/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
valassi: Interesting :-)

I do not agree with his conclusions "This all works out nicely until someone else wants to include geometry.hpp without including <vector> first, and get something like geometry.hpp:1: error: ‘vector’ is not a member of ‘std’. You have now forced all the users of your geometry library to include <vector> before including geometry.hpp."

What I would conclude is: I have forced the person who wrote geometry.hpp to fix their _BUG_ and include vector inside geometry.hpp ;-)

There are several different opinions, again. Google have their own, different from mine, different from yours: https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

Anyway, on this one (unlike the #ifdefs!) I have not such strong opinions. It does not matter. 

Maybe let's see what Stefan or others think. As long as we have a convention and stick with that, fine by me. (I am sure that each convention will include alphabetical ordering within each block, even if different conventions order blocks in different orders!)

hageboeck: > What I would conclude is: I have forced the person who wrote geometry.hpp to fix their BUG and include vector inside geometry.hpp ;-)

That's exactly the point. He phrases this wrongly, but including in the order
```
#include "MyHeader.h"

#include <other stuff>
```
uncovers bugs.

examples/gpu/eemumu_AV/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
hageboeck: Should I add this now? That's maybe good if `-d` is active, but it shouldn't print in general, should it?

hageboeck: I did add it, but only in `-d` mode.

examples/gpu/eemumu_AV/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
valassi: Boh, I am not convinced but I see your point. Pick your convention on includes - but not on #ifdefs, lets keep memcpy there.

I suggest the google convention, IIUC it fits your requirements, and is a standard. Else choose...
```
In dir/foo.cc or dir/foo_test.cc, whose main purpose is to implement or test the stuff in dir2/foo2.h, 
order your includes as follows:

    dir2/foo2.h.
    A blank line
    C system headers (more precisely: headers in angle brackets with the .h extension), e.g., <unistd.h>, <stdlib.h>.
    A blank line
    C++ standard library headers (without file extension), e.g., <algorithm>, <cstddef>.
    A blank line
    Other libraries' .h files.
    Your project's .h files.

Separate each non-empty group with one blank line.

With the preferred ordering, if the related header dir2/foo2.h omits any necessary includes, 
the build of dir/foo.cc or dir/foo_test.cc will break.
...
Within each section the includes should be ordered alphabetically. 
```

examples/gpu/eemumu_AV/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
valassi: thanks  - you can even ompit the _print and just comment it out, both are ok

examples/gpu/eemumu_AV/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
valassi: Thanks for going back to memcpy and the original #ifdef :-)
I think memcpy with the explicit size makes it much clearer that these are fixed size C-like arrays, much more similar to cuda.

valassi (approved):

PR #50, opened 2020-11-18
[eemm cuda] Add functionality to dump a file with 4-momenta and matrix elements.

### LINKED ISSUE #38: eemumu_AV depends on CUDA for Host Only CPU when CUDA is not defined
Even when the __CUDACC__ is not defined, eemumu_AV depends on CUDA for Host Only CPU

hageboeck:
In order to have automated tests, implement the following:
- `[g]check.exe -f <dumpFileName>` can dump files with 4-momenta and matrix elements

### DIFF
--- a/epoch1/cuda/ee_mumu/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
+++ b/epoch1/cuda/ee_mumu/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
@@ -33,7 +33,7 @@ bool is_number(const char *s) {
 
 int usage(char* argv0, int ret = 1) {
   std::cout << "Usage: " << argv0
-            << " [--verbose|-v] [--debug|-d] [--performance|-p] [--json|-j]"
+            << " [--verbose|-v] [--debug|-d] [--performance|-p] [--json|-j] [--file|-f <file to dump events to>]"
             << " [#gpuBlocksPerGrid #gpuThreadsPerBlock] #iterations" << std::endl << std::endl;
   std::cout << "The number of events per iteration is #gpuBlocksPerGrid * #gpuThreadsPerBlock" << std::endl;
   std::cout << "(also in CPU/C++ code, where only the product of these two parameters counts)" << std::endl << std::endl;
@@ -86,6 +86,7 @@ int main(int argc, char **argv)
   int jsonrun = 0;
   int numvec[5] = {0,0,0,0,0};
   int nnum = 0;
+  std::string dumpFileName;
 
   for (int argn = 1; argn < argc; ++argn) {
     if (strcmp(argv[argn], "--verbose") == 0 || strcmp(argv[argn], "-v") == 0)
@@ -99,6 +100,14 @@ int main(int argc, char **argv)
     else if (strcmp(argv[argn], "--json") == 0 ||
              strcmp(argv[argn], "-j") == 0)
       json = true;
+    else if (strcmp(argv[argn], "--file") == 0 ||
+             strcmp(argv[argn], "-f") == 0)
+    {
+      ++argn;
+      if (argn >= argc || is_number(argv[argn]))
+        return usage(argv[0]);
+      dumpFileName = argv[argn];
+    }
     else if (is_number(argv[argn]) && nnum<5)
       numvec[nnum++] = atoi(argv[argn]);
     else
@@ -141,6 +150,12 @@ int main(int argc, char **argv)
   {
     std::cout << "ERROR! #threads/block should be <= " << ntpbMAX << std::endl;
     return usage(argv[0]);
+  }
+
+  std::ofstream dumpFile;
+  if ( !dumpFileName.empty() )
+  {
+    dumpFile.open(dumpFileName, std::ios::trunc);
   }
 
   const int ndim = gpublocks * gputhreads; // number of threads in one GPU grid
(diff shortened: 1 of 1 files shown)


epoch1/cuda/ee_mumu/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
```diff
@@ -421,30 +436,51 @@ int main(int argc, char **argv)
 
     for (int ievt = 0; ievt < nevt; ++ievt) // Loop over all events in this iteration
     {
-      if (verbose)
-      {
-        // Display momenta
-        const int ipagM = ievt/neppM; // #eventpage in this iteration
-        const int ieppM = ievt%neppM; // #event in the current eventpage in this iteration
-        std::cout << "Momenta:" << std::endl;
+      auto getMomentum = [&](std::size_t evtNo, int particle, int component) {
```
valassi, line 439: I would stick to the same naming scheme used throughout the rest of the code:
- particle -> ipar (index up to npar)
- component -> ip4 (index up to np4)
- page -> ipagM (page for Momenta)
- evtNo: I would use ievt, but then it may be better to move this outside the loop on ievt to avoid human confusion

epoch1/cuda/ee_mumu/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
```diff
@@ -421,30 +436,51 @@ int main(int argc, char **argv)
...
+        assert(component < mgOnGpu::np4);
+        assert(particle  < mgOnGpu::npar);
+        const auto page  = evtNo / neppM; // #eventpage in this iteration
+        const auto ieppM = evtNo % neppM; // #event in the current eventpage in this iteration
+        return hstMomenta[page * npar*np4*neppM + particle * neppM*np4 + component * neppM + ieppM];
```
valassi, line 444: I would add back the comment  // AOSOA[ipagM][ipar][ip4][ieppM]

epoch1/cuda/ee_mumu/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
```diff
@@ -421,30 +436,51 @@ int main(int argc, char **argv)
...
+      };
+      auto dumpParticles = [&](std::ostream& stream, std::size_t evtNo, unsigned precision) {
```
valassi, line 446: consider moving this outside the ievt loop too for consistency

epoch1/cuda/ee_mumu/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
```diff
@@ -421,30 +436,51 @@ int main(int argc, char **argv)
...
+        const auto width = precision + 8;
```
valassi, line 447: I trust you on this one :-)
I am not sure if there is a need to generalise, but maybe you need it in the file dump

epoch1/cuda/ee_mumu/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
```diff
@@ -421,30 +436,51 @@ int main(int argc, char **argv)
...
     for (int ievt = 0; ievt < nevt; ++ievt) // Loop over all events in this iteration
     {
-      if (verbose)
-      {
-        // Display momenta
-        const int ipagM = ievt/neppM; // #eventpage in this iteration
-        const int ieppM = ievt%neppM; // #event in the current eventpage in this iteration
-        std::cout << "Momenta:" << std::endl;
+      auto getMomentum = [&](std::size_t evtNo, int particle, int component) {
+        assert(component < mgOnGpu::np4);
```
valassi, line 440: I am not used to asserts like this, but you are most probably right :-)
Is this the recommended way for internal consistency checks? (that you can disable with NDEBUG)?

We should ask @oliviermattelaer but I guess he does not mind?

epoch1/cuda/ee_mumu/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
```diff
@@ -143,6 +152,12 @@ int main(int argc, char **argv)
     return usage(argv[0]);
   }
 
+  std::ofstream dumpFile;
+  if ( !dumpFileName.empty() )
+  {
+    dumpFile.open(dumpFileName, std::ios::trunc);
```
valassi, line 158: I would rather open the file further down an dremove the need for " if (dumpFile.is_open())"... or maybe this is not possible because the latter is within a loop while the "open" call is only once?

epoch1/cuda/ee_mumu/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
```diff
@@ -421,30 +436,51 @@ int main(int argc, char **argv)
...
         std::cout << " Matrix element = "
                   << hstMEs[ievt] << " GeV^" << meGeVexponent << std::endl; // FIXME: assume process.nprocesses == 1
         std::cout << std::string(80, '-') << std::endl;
       }
+
+      if (dumpFile.is_open())
+      {
+        dumpFile << "Event " << std::setw(8) << ievt << "  "
+                 << "Batch " << std::setw(4) << iiter << "\n";
+        dumpParticles(dumpFile, ievt, 15);
```
valassi, line 480: why 6 to screen and 15 to file? is the actual precision 15 while 6 is a truncation?

valassi (changes requested):
Thanks @hageboeck ! A few requested changes (I especially like to have a consistent naming convention), plus a few questions/suggestions.

oliviermattelaer:
So those changes were not included in my epoch2 port. I'm confused with the changed requested by Andrea.
Pushing such changes in MG5aMC will not be an issue, but I need a diff file such that I can include that in MG5aMC directly.

hageboeck:
Since the functionality of dumping a new reference file is now also in the test executable, this isn't needed, any more.

PR #59, opened 2020-11-20
[eemumu epoch1] Make code compile without CUDA

hageboeck:
This disentangles testing and compilation without CUDA that was previously in #50 

In order to have automated tests, implement the following:
- Allow for setting compiler options from outside the Makefile using `MGONGPU_CONFIG=<option>`
- Make CPU version compile without cuda (requires `make MGONGPU_CONFIG=-DMGONGPU_COMMONRAND_ONHOST`,  `export MGONGPU_CONFIG=-DMGONGPU_COMMONRAND_ONHOST` or similar)
- When no cuda toolkit is found, the Makefile only compiles the CPU version. (also requires `-DMGONGPU_COMMONRAND_ONHOST` to not interfere with the default settings.)

### DIFF
--- a/epoch1/cuda/ee_mumu/SubProcesses/Makefile
+++ b/epoch1/cuda/ee_mumu/SubProcesses/Makefile
@@ -3,11 +3,11 @@ CUARCHNUM=70
 LIBDIR=../../lib
 INCDIRS=-I. -I../../src -I../../../../../tools/
 MODELLIB=model_sm
-CXXFLAGS= -O3 $(INCDIRS) -DUSE_NVTX -Wall -Wshadow
+CXXFLAGS= -O3 $(INCDIRS) -DUSE_NVTX -Wall -Wshadow ${MGONGPU_CONFIG}
 ###CXXFLAGS= -O2 $(INCDIRS) -DUSE_NVTX -Wall -Wshadow
 CUARCHFLAGS= -arch=compute_$(CUARCHNUM)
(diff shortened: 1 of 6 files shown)


lfield:
Builds and runs for me. I only tested the host build.

valassi:
Hi @hageboeck sorry for the delay. (I thought you were still waiting for the review, but I see this has been merged already). Looks good to me. Thanks! Andrea

epoch1/cuda/ee_mumu/SubProcesses/Makefile
```diff
@@ -1,23 +1,27 @@
...
-CULIBFLAGS= -L/usr/local/cuda/targets/x86_64-linux/lib
-NVCC=nvcc
 CXX=g++
+NVCC?=$(shell which nvcc)
+ifeq (, $(NVCC))
+	NVCC=@echo No nvcc to run nvcc
+else
+	USE_NVTX?=-DUSE_NVTX
+	CUARCHNUM=70
+	#CUARCHNUM=61 #(For Pascal Architecture Cards)
```
valassi, line 13: Actually, I do have a minor issue here.
My emacs gives me a warnings about indented comments.
As I got used to avoiding indeneted comments, I thought that make would fail, actually it does not.
But I just understood that there is a reason emacs complains, indented comments are parsed and executed: https://stackoverflow.com/a/18363477

Anyway, I suggest that you fix this in your next PR #63. I will repeat the comment there.

PR #61, opened 2020-11-24
add missing json info, add possibility in perf.py to plot a compariso…

roiser:
adds missing info in json and a change in perf.py plotting for adding a cpu value in the 2d plots


oliviermattelaer:
included in epoch2 PR

valassi (approved):
Hi Stefan,
by just looking superficially at the code, I see nothing controversial. Then I have not tested the json output or the json plotting. Maybe one thing that could eventually be improved is to keep in sync the json and stdout in a stringer way, using a single outputting code, but i may be a complex overkill. Just go ahead and merge.
Thanks Andrea

valassi:
Ah sorry, should have checked
```
check.cc: In function ‘int main(int, char**)’:
check.cc:505:10: warning: unused variable ‘stdwtim’ [-Wunused-variable]
   double stdwtim = std::sqrt( sqswtim / niter - meanwtim * meanwtim );
          ^~~~~~~
```
Maybe just comment out the line if not needed?
Thanks

PR #62, opened 2020-11-24
WIP - extract gpu properties

roiser:
little tool to extract gpu card properties

### DIFF
--- a/tools/gpuprops.cu
+++ b/tools/gpuprops.cu
@@ -140,8 +140,6 @@ public:
     for (auto i : allprops.front())
       maxwidth = std::max(i.first.length(), maxwidth);
 
-    std::cout << maxwidth << std::endl;
-
     for (auto p : allprops ) {
       std::cout << "DeviceID: " << did << std::endl;
       for (auto i : p) {
--- a/tools/gpuprops.cu
+++ b/tools/gpuprops.cu
@@ -151,6 +151,48 @@ public:
 
 };
 
+/* calculate cores per streaming multiprocessor 
+https://stackoverflow.com/questions/32530604/how-can-i-get-number-of-cores-in-cuda-device
+#include "cuda_runtime_api.h"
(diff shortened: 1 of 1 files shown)


lfield:
Have you tried the clinfo command?

roiser:
No I didn't know the tool. I just compiled it from source ( https://github.com/Oblomov/clinfo ) and it delivers the same info (even a bit more) than the script. 

Do you know if there is an rpm for e.g. Centos8? I couldn't find one. Also we would introduce a dependency on OpenCL, I'm a bit wary if we want to do this.

Just to add. In view of upcoming testing I thought it may be useful to extract necessary info about the underlying card to allow us to configure the performance tests accordingly.

lfield:
There isn't a one to one mapping for the properties but it does give a similar overview. I had the same idea as you when I was looking at the OpenCL API but then found this instead. It does't add many dependencies. All GPU/FPGA cards should support a version of OpenCL and the OpenCL lib is readily available. The package is not in EPEL for CentOS8 but I can ask the packager if they can release it. There are also two SYCL info versions but I haven't tied them yet. Might do that today.

lfield:
There is also glxinfo for the OpenGL API. Rather than gpuprops I would call this CUDA info but I would be surprised if this tool did not already exist.

lfield:
btw WIP pull requests should be opened as draft pull requests.

roiser:
> btw WIP pull requests should be opened as draft pull requests.

thanks, converted now, the reason why it was a "normal" PR is that I"m using by default the URL that git provides to me after pushing which seems to make it a standard PR

lfield:
Have you tried this in the samples directory?

/usr/local/cuda/samples/1_Utilities/deviceQuery

lfield:
In fact building and running deviceQuery is a [recommend action](https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html#running-binaries) for verifying the integrity of the installation.

PR #52, opened 2020-11-18
Test adding simple github actions CI.

hageboeck:
Test github CI. Note that this has two commits from #63, so that one is blocking the merge of this one here.

Fixes #30

PR #63, opened 2020-11-25
[ep1 cuda eemm] Make Makefiles more installation independent.

### LINKED ISSUE #55: src/makefile is not installation independent
we have issue with various variable
  * CUINC=/usr/local/cuda/targets/x86_64-linux/include
  * CXX=g++

hageboeck:
Reorganise Makefiles to prepare it for addition of tests.
(This version of the makefile runs already in #52, but took here a separate PR for easier reviewing).

- Remove hard-coded CUDA paths. Instead, react to CUDA_HOME if exported.
- Replace "cd xxx && make" by the more standard "make -C".
- If no nvcc is found and no MGONGPU_CONFIG is given, automatically switch to C++ random numbers.
This allows for compiling the CPU-only code.
- Separate optimisation flags from other C++ flags. This facilitates
switching between debug and optimised builds, because only one variable needs to be overwritten.
- Protect all cuda-related variables with a conditional, so the
corresponding targets don't get built.

Fixes #55.

### DIFF
--- a/epoch1/cuda/ee_mumu/SubProcesses/Makefile
+++ b/epoch1/cuda/ee_mumu/SubProcesses/Makefile
@@ -1,33 +1,53 @@
-LIBDIR =../../lib
-INCDIRS=-I. -I../../src -I../../../../../tools/
-MODELLIB=model_sm
-CXXFLAGS= -O3 -std=c++11 $(INCDIRS) $(USE_NVTX) -Wall -Wshadow ${MGONGPU_CONFIG}
-LIBFLAGS= -L$(LIBDIR) -l$(MODELLIB)
-CXX=g++
-NVCC?=$(shell which nvcc)
-ifeq (, $(NVCC))
-	NVCC=@echo No nvcc to run nvcc
-else
-	USE_NVTX?=-DUSE_NVTX
-	CUARCHNUM=70
-	#CUARCHNUM=61 #(For Pascal Architecture Cards)
-	CUINC=-I/usr/local/cuda/targets/x86_64-linux/include
-	CULIBFLAGS= -L/usr/local/cuda/targets/x86_64-linux/lib -lcurand
-	CUARCHFLAGS= -arch=compute_$(CUARCHNUM)
-	# CUARCHFLAGS= -gencode arch=compute_$(CUARCHNUM),code=sm_$(CUARCHNUM)
-	CUFLAGS= -O3 $(INCDIRS) $(USE_NVTX) $(CUARCHFLAGS) -use_fast_math -lineinfo ${MGONGPU_CONFIG}
-	# Without -maxrregcount: baseline throughput: 6.5E8 (16384 32 12) up to 7.3E8 (65536 128 12)
-	###CUFLAGS+= --maxrregcount 160 # improves throughput: 6.9E8 (16384 32 12) up to 7.7E8 (65536 128 12)
-	###CUFLAGS+= --maxrregcount 128 # improves throughput: 7.3E8 (16384 32 12) up to 7.6E8 (65536 128 12)
-	###CUFLAGS+= --maxrregcount 96 # degrades throughput: 4.1E8 (16384 32 12) up to 4.5E8 (65536 128 12)
-	###CUFLAGS+= --maxrregcount 64 # degrades throughput: 1.7E8 (16384 32 12) flat at 1.7E8 (65536 128 12)
+LIBDIR   = ../../lib
+TOOLSDIR = ../../../../../tools/
+INCFLAGS = -I. -I../../src -I$(TOOLSDIR)
+MODELLIB = model_sm
+OPTFLAGS = -O3
+CXXFLAGS = $(OPTFLAGS) -std=c++11 $(INCFLAGS) $(USE_NVTX) -Wall -Wshadow $(MGONGPU_CONFIG)
+LIBFLAGS = -L$(LIBDIR) -l$(MODELLIB)
+CXX     ?= g++
+
+#Export CUDA_HOME to select a cuda installation
+ifndef CUDA_HOME
+  NVCC ?= $(shell which nvcc 2>/dev/null)
+  ifneq ($(NVCC),)
+    # NVCC is in the PATH or set explicitly
+    CUDA_HOME  = $(patsubst %bin/nvcc,%,$(NVCC))
+    CUDA_HOME := $(warning No CUDA_HOME exported. Using "$(CUDA_HOME)") $(CUDA_HOME)
+  endif
 endif
+
+ifdef CUDA_HOME
+  NVCC = $(CUDA_HOME)/bin/nvcc
+  CUARCHNUM=70
+  #CUARCHNUM=61 #(For Pascal Architecture Cards)
+  USE_NVTX?=-DUSE_NVTX
+  CUARCHFLAGS= -arch=compute_$(CUARCHNUM)
+  # CUARCHFLAGS= -gencode arch=compute_$(CUARCHNUM),code=sm_$(CUARCHNUM)
+  CUINC       = -I$(CUDA_HOME)/include/
+  CULIBFLAGS  = -L$(CUDA_HOME)/lib64/ -lcuda -lcurand
+  CUFLAGS= $(OPTFLAGS) -std=c++14 $(INCFLAGS) $(CUINC) $(USE_NVTX) $(CUARCHFLAGS) -use_fast_math -lineinfo $(MGONGPU_CONFIG)
+  # Without -maxrregcount: baseline throughput: 6.5E8 (16384 32 12) up to 7.3E8 (65536 128 12)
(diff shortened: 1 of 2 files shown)


epoch1/cuda/ee_mumu/SubProcesses/Makefile
```diff
@@ -1,33 +1,50 @@
...
+#Export CUDA_HOME to point to a cuda installation
+ifdef CUDA_HOME
+	NVCC       ?=   $(CUDA_HOME)/bin/nvcc
+	CUINC       = -I$(CUDA_HOME)/include/
+	CULIBFLAGS  = -L$(CUDA_HOME)/lib64/
+endif
+
+ifneq ($(NVCC),)
 	CUARCHNUM=70
 	#CUARCHNUM=61 #(For Pascal Architecture Cards)
```
valassi, line 21: I leave this comment here about changes which actually came from PR #59.

My emacs gives me a warnings about indented comments.
As I got used to avoiding indeneted comments, I thought that make would fail, actually it does not.
But I just understood that there is a reason emacs complains, indented comments are parsed and executed: https://stackoverflow.com/a/18363477

Would you agree to avoiding indented comments?
Otherwise I can also tweak my emacs (though I'd rather not)

By the way, note that those definitions like CUARCHNUM=70 do NOT need to be indented. The easiest is to actually bring everything back to the first column. Indentation is really only needed for excuting build sequences.

valassi (changes requested):
Hi @hageboeck thanks for the PR.

I did not look at the code in detail yet, simply because I tried and it does not build for me, so I suggest you investigate this first:
```
[avalassi@itscrd03 bash] ~/GPU2020/madgraph4gpu/epoch1/cuda/ee_mumu/SubProcesses/P1_Sigma_sm_epem_mupmum> make
/usr/local/cuda/bin/nvcc  -O3 -std=c++14 -I. -I../../src -I../../../../../tools/  -DUSE_NVTX -arch=compute_70 -use_fast_math -lineinfo  -c gcheck.cu -o gcheck.o
make -C ../../src 
make[1]: Entering directory '/afs/cern.ch/user/a/avalassi/GPU2020/madgraph4gpu/epoch1/cuda/ee_mumu/src'
/cvmfs/sft.cern.ch/lcg/releases/gcc/8.3.0-cebb0/x86_64-centos7/bin/g++  -O3 -std=c++11 -I.  -Wall -Wshadow   -c Parameters_sm.cc -o Parameters_sm.o
/cvmfs/sft.cern.ch/lcg/releases/gcc/8.3.0-cebb0/x86_64-centos7/bin/g++  -O3 -std=c++11 -I.  -Wall -Wshadow   -c read_slha.cc -o read_slha.o
/cvmfs/sft.cern.ch/lcg/releases/gcc/8.3.0-cebb0/x86_64-centos7/bin/g++  -O3 -std=c++11 -I.  -Wall -Wshadow   -c rambo.cc -o rambo.o
In file included from rambo.cc:8:
rambo.h:12:10: fatal error: curand.h: No such file or directory
 #include "curand.h"
          ^~~~~~~~~~
compilation terminated.
make[1]: *** [Makefile:33: rambo.o] Error 1
make[1]: Leaving directory '/afs/cern.ch/user/a/avalassi/GPU2020/madgraph4gpu/epoch1/cuda/ee_mumu/src'
make: *** [Makefile:65: ../../lib/libmodel_sm.a] Error 2
```

The only specific comment I left is about indentation in the Makefile. This is actually about previous changes in PR #59. Can you fixthis too please?

Thanks Andrea

epoch1/cuda/ee_mumu/SubProcesses/Makefile
hageboeck: I didn't program `make` in a looong time. I also see vim screwing up the syntax colouring when you indent, so I might as well unindent all bodies of the conditionals. I will have a look tomorrow.

hageboeck:
> /cvmfs/sft.cern.ch/lcg/releases/gcc/8.3.0-cebb0/x86_64-centos7/bin/g++  -O3 -std=c++11 -I.  -Wall -Wshadow   -c rambo.cc -o rambo.o
> In file included from rambo.cc:8:
> rambo.h:12:10: fatal error: curand.h: No such file or directory
>  #include "curand.h"
>           ^~~~~~~~~~

Interesting! Where is your cuda include directory? Is this centos7?
It's pretty obvious that it's not in a path that's searched by `g++`, so we need something like a `CUDA_HOME` or similar to find it.

valassi:
Hi, actually I am using itscrd03: you may also try that, but I am assuming you have a similar setup? I wonder if one of us two has something extra in his environment variables?

Ah actually I do: this, plus a few things realted to gcc, cmake, git etc from the LCG stack
```
cuda_home=/usr/local/cuda/bin
if [ -d $cuda_home ]; then
  echo "Using Cuda from $cuda_home"
  cuda_version=`\ls -l /usr/local/ | grep "cuda " | awk -F "-" '{ print $NF }'`
  echo "Using Cuda version ${cuda_version}"
  export PATH=${cuda_home}:${PATH} # THIS IS NEEDED
  echo "Using $(which nvcc)"
  nvcc --version | tail -1
else
  echo "WARNING! $cuda_home does not exist!"
fi
```

Should I remove that? Note that this is the way I was building before, I am not sure the old makefiles would work without.

Anyway, I am quite sure that what you are doing will be more consistent ;-)
Just let me know

Thanks Andrea

epoch1/cuda/ee_mumu/SubProcesses/Makefile
hageboeck: Ok, problem understood:
In *recipes*, indented comments are indeed parsed as shell commands.

Here, we are not in a recipe, so indentation is possible but not required. What's "wrong", though, is that the blocks should be indented with spaces and not with tabs.

hageboeck: We would have to agree on an indentation width, though, if we chose to indent the body of conditionals.

hageboeck:
> Should I remove that? Note that this is the way I was building before, I am not sure the old makefiles would work without.

No, I think you are really close to the perfect setup.
With the last two commits, the Makefiles *should* modulo bugs do as follows:
- `CUDA_HOME` is exported. The Makefile uses that to find cuda includes and cuda libs. (Note that in your snippet, it's lower case and not being exported).
- `CUDA_HOME` is not exported, but `which nvcc` returns a valid `nvcc`. The Makefile strips `bin/nvcc` from the path and uses that to find cuda includes/libs.
- Neither of the above is true: Only the C++ part without curand will be compiled.

On itscrd04, I can successfully compile all three cases by changing my environment accordingly.

oliviermattelaer:
FYI, this is already include in the epoch2 PR.

Do not hesitate to double check and apply for a PR for that branch if you spot a change that I did not include

epoch1/cuda/ee_mumu/SubProcesses/Makefile
hageboeck: @valassi Can you check if emacs also complains if you indent a comment with spaces? That should be "legal" in make.

hageboeck:
> FYI, this is already include in the epoch2 PR.
> 
> Do not hesitate to double check and apply for a PR for that branch if you spot a change that I did not include

Hi @oliviermattelaer,
I made some more changes so the code even compiles without setting CUDA_HOME. But just go ahead and get epoch2 ready. Once epoch2 is merged and this one here is also done, I will port the changes that we might still introduce here.

epoch1/cuda/ee_mumu/SubProcesses/Makefile
roiser: Hi @valassi , following the discussion below I presume this issue can be marked as resolved?

valassi:
Hi @hageboeck thanks now the build works even with my original setup, thanks to `CUDA_HOME  = $(patsubst %bin/nvcc,%,$(NVCC))`. 

I have also changed my bash setup so that it gives something better
```
cuda_home=/usr/local/cuda
if [ -d $cuda_home ]; then
  export CUDA_HOME=$(cd $cuda_home; pwd -P)
  echo "Using CUDA_HOME=$CUDA_HOME"
  export PATH=${CUDA_HOME}/bin:${PATH}
  nvcc --version | tail -1
else
  echo "WARNING! $cuda_home does not exist!"
fi
```

About Makefiles, I see no emacs warnings/complaints any longer.

All looks good, you can merge, thanks.

Andrea

valassi (approved):

roiser (approved):
Hi, the changes look good to me. Just one minor suggestion, can we add an 

info:
   $(NVCC) --version
   $(CXX) --version

target to have a quick info about the setup we are using, we can add more stuff there in the future. 

I did this in my setup.sh script before just to be sure I'm using the right cuda and gcc versions.

hageboeck:
> Hi, the changes look good to me. Just one minor suggestion, can we add an
> 
> info:
> $(NVCC) --version
> $(CXX) --version

Done.

PR #64, opened 2020-11-25
Initial commit after running dpct

lfield:
This purpose of this epoch1_dpct branch is to show the differences made to the cuda code after running dpct. This tool attempts to automatically convert CUDA code to SYCL code. While not perfect, it will highlight the places in the code that require attention and remove some of the laborious work involved with porting. The following command was used. 
```
dpct  --extra-arg="-std=c++11" --extra-arg="-D __CUDACC__" --extra-arg "-I ../../../tools/" --comments  --usm-level=none --cuda-include-path /usr/local/cuda-11.0/include --in-root . SubProcesses/P1_Sigma_sm_epem_mupmum/*.cu src/*.cu
````


tjburch:
This is a good idea to have - I'm curious about the `--usm-level=none` flag though on the conversion command. Is this just personal preference for allocators over USM or is there a problem when rinning with USM?

lfield:
Personal objective rather than preference. I would like the resulting port to be standards-based SYCL 1.2 rather than a specific DPC++ port with vendor extensions. Even though USM will be in the 2020 spec, I would like to wait until the spec is out and implemented before using it.

lfield:
I am closing this as there is no value in merging. The diff can always be seen from the branch in my repository.

PR #65, opened 2020-11-25
Fix minor syntax errors

tjburch:
This patches a couple of small typos/syntax errors I noticed trying to compile


tjburch:
In addition to the few I found in `rambo.cc`, there's also a typo `constdouble` in `HelAmps_sm.cu` that was resolved.

oliviermattelaer:
Thanks sorry for that, I did not had a cuda machine available when i implement those cuts. Thanks for fixing this.

PR #69, opened 2020-11-26
first version of ee_mumu coming from madgraph --some PR still need to…

oliviermattelaer:
Hi,

This is the current output generated by MG5aMC which is relatively close to the current output of epoch1.
I likely miss some recent (less than two weeks old) PR. If you can put those in the comment below, this will help me to check them one by one and to include the associate change.

The mG5aMC code is not yet running for gg>tt~gg but this will likely be working soon.


roiser:
Hi @oliviermattelaer there is #61 which needs to be still approved (I'll try to ping reviewers) but includes also some changes for the json output in epoch 1 ee-mumu

oliviermattelaer:
Thanks Stefan, this should be in now

oliviermattelaer:
@valassi @roiser @hageboeck 
The Cpp code is now compiling (but requires curand obviously so it needs cuda)

@hageboeck should we put the test for epoch2? would be nice to have it setup before merging this branch no?

hageboeck:
Yes, let me check out the branch and see if I can get the test to run. You can take the commits once done.

hageboeck:
@oliviermattelaer see #78. I managed to port the tests, and they fail. :-D

hageboeck:
@oliviermattelaer I tried to push the commits to activate the CI on top, but I failed. I will reset. :-)

Issue #71, opened 2020-11-29
kernel launchers and SIMD vectorization


valassi:
I finally found some time to pursue some earlier tests on an idea I had from the beginning, namely, trying to implement SIMD vectorization in the C++ code as the same time as SIMT/SPMD parallelisation on the GPU in cuda,

The idea is always the same: event-level parallelism, with execution in lockstep (all events gop through exactly the same sequence of computations).

I pushed a few initial tests in https://github.com/valassi/madgraph4gpu/tree/klas, I will create a WIP PR about that. @roiser , @oliviermattelaer , @hageboeck , I would especially be interested to have some feedback from you :-)

Implementing SIMD in the C++ is closely linked to the API of kernel launchers (and of the methods the kernels internally call) on the GPU. In my previous eemumu_AV implementation, the signature of some c++ methods was modified by adding nevt (a number of events), or instead ievt (an index over events) with respect to the cuda signature, but some lines of code (eg loops on ievt=1..nevt) were commented out as they were just reminders of possible future changes.

The main idea behind all the changes I did is simple: bring the event loop more and more towards the inside. *Eventually, the event loop must be the innermost loop*. This is because you eventually want to perform every single floating point addition or multiplication in parallel over several events. In practice, one concrete consequence of this is that I had to invert the order of the helicity loop: so far, there was an outer event loop, with an inner loop over helicities within each event, while now there is an outer helicity loop, with an inner loop over events for each helicity.

One limitation of the present code (possible in a simple eemumu calculation) is that there is no loop over nprocesses, because nprocesses=1. This was already assumed, but now I made it much more explicit, removing all dead code and adding FIXME warnings.

So far, I got to this point 
- I introduced fptype_v and cxtype_v vector types, with width neppV=neppM (the same parameter used for AOSOA memory layouts in the GPU, though the choice of the parameter on the CPU and on the GPU may be different: on the former, it is the SIMD width that counts, on the latter it is the cacheline that matters).
- I have used cxtype_v arguments in the ixxx/oxxx function signatures, which also accept an iepgV parameter (index over event pages). Then I do the loop over the neppV events in each page within ixxx/oxxx.
- For the moment I am just doing simple tests with autovectorization. I have added -march=native to the build options, and am observing the results of objdump to see if there are some vectorized loops, eg "ymm". I am using Sebastien Ponce's excellent manual as a guide.
[...]

PR #72, opened 2020-11-29
klas - kernel launchers and SIMD vectorization

valassi:
See https://github.com/madgraph5/madgraph4gpu/issues/71
This is work in progress... comments welcome (preferably on the PR, or also directly on the code)


lfield:
Doesn't build for me. Same problem that @hageboeck fixed in PR #106. After fixing manually it terminates with `Illegal instruction (core dumped)`. I think this is because the CPU on my old laptop does not support avx2. It is a four core `Intel(R) Core(TM) i5-2520M CPU @ 2.50GHz`. 

``` 
CPPProcess.cc: In function ‘void Proc::calculate_wavefunctions(int, const fptype_sv*, mgOnGpu::fptype*, int)’:
CPPProcess.cc:574:5: error: ‘npagV’ not specified in enclosing ‘parallel’
  574 |     for ( int ipagV = 0; ipagV < npagV; ++ipagV )
      |     ^~~
CPPProcess.cc:573:9: error: enclosing ‘parallel’
  573 | #pragma omp parallel for default(none) shared(ihel,allmomenta,allMEs,cHel,cIPC,cIPD,denom,cf) private (w_v,amp_v,jamp_v)
      |         ^~~
make: *** [Makefile:82: CPPProcess.o] Error 1
```
Illegal instruction (core dumped)

valassi:
Hi @hageboeck I merged this to master and I am quite sure the main executable gives the right results (I get the same average ME exactly on cuda and c++), however the test fails. I have no idea what is wrong, maybe the reference needs to be rebuilt. Can you please have a quick look? Thanks.

I anticipate that there will be quite a few more structural changes ahead, I am tempted to simply remove this test otherwise from the CI. I do test functionality anyway. What do you think? Thanks

hageboeck:
Hi @valassi,

It seems to be rambo that gives different results. Does this use fast math?

valassi:
Hi @hageboeck interesting... yes it uses fast math but this should only make a difference for NaNs really. The biggest change is the vectorization, so maybe I did not make the right change in the runTest.cc... but again th efinal results look correct, so I am not sure. Thanks fo rlooking!

hageboeck:
Hi @valassi,

an alternative is to make the test less aggressive. After all, it's comparing the momenta to a relative precision of 1.E-12. With fast math, I can imagine that this doesn't hold, because `-funsafe-math-optimizations` is part of fast math.
If people are OK to have momenta agree to 1.E-9 or so, the test failure would probably vanish.

valassi:
Hi @hageboeck thanks. I am surprised you say so, my impression is that there is a much more fundamental problem, the results really ARE different: https://github.com/madgraph5/madgraph4gpu/pull/72/checks?check_run_id=2142392162
```
particle 2	component 1
	 madGraph:   5.849331413473452e+02
	 reference:  6.246401493489108e+02
	 rel delta:  6.356781267255054e-02 exceeds tolerance of 5.000000000000000e-12
particle 2	component 2
	 madGraph:  -3.138365726669761e+02
	 reference: -2.563623175652571e+01
	 rel delta:  1.124191510076706e+01 exceeds tolerance of 5.000000000000000e-12
particle 2	component 3
	 madGraph:  -3.490842674916366e+02
	 reference:  4.143277295121539e+02
	 rel delta:  1.842531751139762e+00 exceeds tolerance of 5.000000000000000e-12
Google Test trace:
src/MadgraphTest.cc:131: In comparing event 0 from iteration 0
   0  7.500000000000000e+02  0.000000000000000e+00  0.000000000000000e+00  7.500000000000000e+02
ref0  7.500000000000000e+02  0.000000000000000e+00  0.000000000000000e+00  7.500000000000000e+02

   1  7.500000000000000e+02  0.000000000000000e+00  0.000000000000000e+00 -7.500000000000000e+02
ref1  7.500000000000000e+02  0.000000000000000e+00  0.000000000000000e+00 -7.500000000000000e+02

   2  7.500000000000000e+02  5.849331413473452e+02 -3.138365726669761e+02 -3.490842674916366e+02
ref2  7.500000000000002e+02  6.246401493489108e+02 -2.563623175652571e+01  4.143277295121539e+02

   3  7.500000000000001e+02 -5.849331413473452e+02  3.138365726669761e+02  3.490842674916364e+02
ref3  7.500000000000005e+02 -6.246401493489108e+02  2.563623175652521e+01 -4.143277295121539e+02

  ME  0.000000000000000e+00
r.ME  2.070519044360081e-02
```
I think there is a bug somehwre (probably in the way I fixed a merge conflict with the vectorization part). Note that the last master/usptream, which does use fast math, passes the tests.

I will try to have a look later on.

hageboeck:
Oh, indeed ...
I didn't look close enough and thought that the differences were small.

valassi:
Thanks! I just pushed a bug fix to the printout in teh main driver, now I do get the same in C++ and CUDA if I print momenta (-v option).

I think I still have a (harmless) bug in alloacating too much memory.

But for the test failing, I will have another look later. Thanks!

valassi:
There is certainly one algorithmic bug: in this vectorization PR, the helicity filtering algorithm has changed for C++, now you must precompute (and set) the good helicities, like in CUDA. This might explain why the ME is zero at the end, probably isgmakin thinks that no helicities are good, so the ME is 0.

This does not yet explain why the momenta are different however, Investigating,

valassi:
Hi @hageboeck I found it: in the vectorization PR I have also changed neppR=8, was previously 4. This changes the physics results because the same set of random numbers is interpreted in different ways (the striding of the AOSOA has different parameters).

The reasoning behind this is the following: before vectorization, neppM and neppR were set to the same value, and this was essentially only decided by using single or double precision (it is the number of floating point values to fill a 32 byte cache line). This was true for both c++ and Cuda.

After vectorization, I have started using neppM in c++ to determine the number of events to compute in parallel on a SIMD register, and this changes for SSE2, AVX2, AVX512 or scalar (and it changes for floats and doubles) In CUDA I have kept this to mean how many floats or doubles you fit in a 32byte cache line of a GPU. This makes it difficult to compare results across platforms, and I want strict reproducibility whenever possible. For this reason, I fixed neppR to 8: this is large enough for all practical purposes (to ensure coalseced access in GPU and maybe SIMD in CPU, not yet done), but it makes sure that all of the different options get the same physics results.

To cut a long story short: I need to recompute the comparison file, how do I do that? (I will also need to change neppR to be hardcoded to 8 in epoch2 I guess, if the tests share the reference).

Note that hardcoding neppR is probably also a good idea if you want to run your CI test as-is on both floats and doubles, otherwise I expect you would get different results (not 100% sure - have you tried?).

Anyway, can you tell me how to rebuild the referrence please?
Thanks
Andrea

valassi:
PS Just as a further explanation: neppM is now two things. It is the striding of the AOSOA momenta, and the dimension of the fptype_v vector (using compiler vector extensions) used for SIMD. While it seems nice to keep these two parameters the same, I am not even convinced this is strictly necessary: when passing momenta to the number crunching machinery, I think I am presently passing the momentum SIMD-vector by value and not by reference, ie. I am copying the values from AOSOA into a new SIMD vector, which in principle may have a different size. It is only the size of the latter, the fptype_v dimension, which determines the SIMD speedup, because you are replacing four floating point operations by a single one. As for the memory access itself (the AOSOA striding), this seems less relevant (as I said for CUDA). But ok, this is still work in progress to be better understood, I did have a factor 1.5 speedup somewhere possibly from memory access, that I need to investigate better. 
https://github.com/valassi/madgraph4gpu/blob/978526b352a217a2ba81195166dbc037c34fe575/epoch1/cuda/ee_mumu/SubProcesses/P1_Sigma_sm_epem_mupmum/CPPProcess.cc#L45

valassi:
Ok I found the way to rebuild the reference and I committed that. Maybe it will break epoch2, lets see.

hageboeck:
> Anyway, can you tell me how to rebuild the referrence please?
> Thanks
> Andrea

It's here:
https://github.com/madgraph5/madgraph4gpu/blob/d819491f3cb53e219f9d9e7cb000ed4e5188c2b3/test/src/MadgraphTest.cc#L67

I don't remember the details, but it will create a file with a name that's auto generated from the test that's running. You can use this to overwrite the reference.

valassi:
Hi @hageboeck thanks! I had found it myself inthe meantime from some old emails you had sent :-)

Anyway, I think I fixed all isues in the tests
- I rebuilt the reference for neppR=8 in epoch1
- This broke epoch2, so I hardcoded neppR=2 also in epoch2
- Some tests were still failing and I had to increase tolerance from 5E-12 to 8E-11

All checks have passed now. I mark this PR as ready to be merged.

There's clearly some cleanup to be done, and some measurements too, and possibly some more speedups to gain. But I'd like to merge it asap, so at least it's in master and I can build on top of that.

@roiser @oliviermattelaer @hageboeck do you have any objections to merge?
Thanks Andrea

valassi:
For the moment, this is giving a factor ~4 speedup for AVX on the C++. In CUDA it seems a few % slower, for reasons to be investigated.

One thing I need to do as discussed in December is to add support for SSE too, should be fast.

Then I might try to vectorize a few more things here and there. 

And there was an additional factor 1,5 speedup (on top of the factor 4) that I had mentioned at some point. I need to understand where/when/what this was about.

But again, having this in master already now would help.

valassi:
PS Final comment: a factor 4 speedup over scalar C++, as the scalar C++ is 30% slower than the scalar Fortran, means a factor 3 speedup over the current production code, if we manage to interface to madevent - but that's another story

valassi:
While waiting to merge the PR, I also included a few patches that were in the further branch klas2. Without them, `make AVX="none"` was failing to build. This is still ready to merge.

valassi:
This PR # 72 is OBSOLETE. It is superseded by PR #132 (additional SIMD fixes), which is itself superseded by PR #152 (all SIMD fixes, plus epoch1/2 merging as per issue #139 and PR #152).

I copied a few comments to PR #132.

I am closing this now.

Issue #73, opened 2020-11-29
Bug in unique_ptr: arrays not initialized


valassi:
While working on #72 I found out there is a bug coming from PR #47 : the arrays in unique ptrs are not initialized.

This should fix it:
https://github.com/madgraph5/madgraph4gpu/pull/72/commits/e1c7b7c5b0a1db2447ec9af8a9b19e312cf2a9fb

PR #74, opened 2020-11-29
BUG FIX in PR #47: arrays were uninitialised in RAII constructors

valassi:
As described in #73 this is a spinoff of #72

### DIFF
--- a/epoch1/cuda/ee_mumu/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
+++ b/epoch1/cuda/ee_mumu/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc
@@ -69,7 +69,7 @@ std::unique_ptr<T[], CudaHstDeleter<T>> hstMakeUnique(std::size_t N) {
 };
 #else
 template<typename T = fptype>
-std::unique_ptr<T[]> hstMakeUnique(std::size_t N) { return std::unique_ptr<T[]>{ new T[N] }; };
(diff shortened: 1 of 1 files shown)


roiser:
The difference between the two lines should be that the new one initializes the array values all to 0. I just wonder where we depend on 0 inits in the code? Could you point me to the line? 

But of course it's fine to have proper 0 initializations, therefore approve it.

roiser (approved):

valassi:
> 
> 
> The difference between the two lines should be that the new one initializes the array values all to 0. I just wonder where we depend on 0 inits in the code? Could you point me to the line?
> 
> But of course it's fine to have proper 0 initializations, therefore approve it.

Hi Stefan, this has showed up in my latest changes in #71. Amongst other things I reimplemented the helicity filtering in c++ in the same way as in cuda. I was initially getting a worse performance, because some bad helicities were not filtered out. This is from https://github.com/valassi/madgraph4gpu/blob/acc333ac82223842162dcc6e4bf39bb802eafd19/epoch1/cuda/ee_mumu/SubProcesses/P1_Sigma_sm_epem_mupmum/CPPProcess.cc#L724, the missing initialization being in https://github.com/valassi/madgraph4gpu/blob/acc333ac82223842162dcc6e4bf39bb802eafd19/epoch1/cuda/ee_mumu/SubProcesses/P1_Sigma_sm_epem_mupmum/check.cc#L213

Anyway, looks like you approved :-)

Thanks, I will merge
Andrea

PR #78, opened 2020-12-01
Add tests on top of #69

hageboeck:
See PR #69.

Here, the tests for epoch1 were ported to epoch2.
The tests fail for now, since a matrix element changed by more than 1.E-9, which exceeds the test threshold of 1.E-14 (to discuss if that threshold makes sense, anyway).

@oliviermattelaer: Note that I had to rebase your commits on top of madgraph/master to make git understand what's going on. You should be able to take the diffs of my commits, and put them on top of your branch.

The good news:
The tests only needed
```diff
  #include "CommonRandomNumbers.h"
- #include "CPPProcess.h"
+ #include "gCPPProcess.h"
  #include "Memory.h"
  #ifdef __CUDACC__
  #include "grambo.cu"
```

The bad news (@roiser):
The downsides of the epochIsDirectory now shows:
Porting the test files and makefiles was a bit painful, because you have to trick git. You need something like
```
git show <commit hash of commit in ep1> | git am -p 4 -3
```
This
1. Gets the changes from that commit
2. (`-p 4`) Prunes the first 4 directories from the path, so `epoch1` is gone.
3. (`-3`) Tries to apply a three-way merge. That's always necessary, because git won't let the merge happen, because it thinks that it's not the same file (which is true).
4. If lucky, the commit applies or you need to fix the files manually. Half of them applied, half had to be edited.
5. Then it commits with the original message. That's nice!
6. Then I re-edited the commit message to add info about the original commit this came from.

TODO:
- [ ] Move common test infrastructure into something like `test/`
- [ ] Agree on reference for all (?) epochs
- [ ] Test thresholds?


oliviermattelaer:
Excellent, Thanks a lot,

For the physics point of view, I would indeed not worry about relative difference of the order of 1e-7.
But this might shows that we can be quite sensitive to numerical accuracy.

So for me this is good to go.

hageboeck:
@oliviermattelaer does this mean that I should raise the test thresholds for MEs to 1.E-8 or so?
Do you want to port the commits I added to madgraph or will you port these things in epoch3?

oliviermattelaer:
@hageboeck, Yes I think that a larger threshold makes sense.
 
>  Do you want to port the commits I added to madgraph or will you port these things in epoch3?
I have just included the commit into the epoch2 branch. Tell me if I have miss some of the changes.

Thanks,

Olivier

valassi:
Hi, thanks all! Just one comment about the tests: we should separate them into two sets of object files and executables, as we do now for the main executable check.exe and gcheck.exe. Mixing c++ and cuda can give issues, for instance I had some while adding multithreading (using Stephan's suggested omp), which otherwise does work nicely out of the box! See #82 and #83 .

hageboeck:
> @hageboeck, Yes I think that a larger threshold makes sense.
> 
> > Do you want to port the commits I added to madgraph or will you port these things in epoch3?
> 
> I have just included the commit into the epoch2 branch. Tell me if I have miss some of the changes.

@oliviermattelaer Threshold changed, CPU tests passing!
We should talk tomorrow how to merge this.

hageboeck:
Closing in favour of #109

Issue #83, opened 2020-12-03
Improve #52: separate test executable for c++ and cuda


valassi:
Hi, sorry for the delay, I only had time to look at #52 now.

I have one problem with these changes: there is a single runTest.exe for cuda and c++. I would very strongly suggest to have two separate executables, just like we are doing with check.exe and gcheck.exe. The two worlds, C++ and CUDA, are not meant to talk to each other. Either we run cuda, or we run c++. Mixing them together is error prone in various ways.

Some of these issues from mixing I had seen in the past (I do not even remember what they were) and they are the reason why I had gone for two separate CPPProcess.o, gCPPProcess.o, and even two namespaces for extra security.

Some mixing issues are more complex to foresee: for instance, I just gave it a try to add OpenMP as discussed in #82, and this now fails linking runTest.exe, because nvcc does not understand OpenMP (and it has no need to understand it).

More fundamentally, if we are testing the C++ implementation, I would build (ie also link) the test with gcc, not with nvcc.

What is happening now is quite confusing, the .o is built with gcc while the .exe is built with nvcc. I would build two separate .o and two separate .exe.
```
/cvmfs/sft.cern.ch/lcg/releases/gcc/8.3.0-cebb0/x86_64-centos7/bin/g++  -O3 -std=c++11 -I. -I../../src -I../../../../../tools/ -I../../../../../tools//googletest/googletest/include/ -DUSE_NVTX -Wall -Wshadow  -I/usr/local/cuda-11.1/include/ -c runTest.cc -o runTest.o

ln -sf runTest.cc runTest_tmp.cu

/usr/local/cuda-11.1/bin/nvcc -o runTest.exe CPPProcess.o runTest.o gCPPProcess.o runTest_tmp.cu  -O3 -std=c++14 -I. -I../../src -I../../../../../tools/ -I../../../../../tools//googletest/googletest/include/ -I/usr/local/cuda-11.1/include/ -DUSE_NVTX -arch=compute_70 -use_fast_math -lineinfo  -ldl -L../../lib -lmodel_sm -L../../../../../tools//googletest/build/lib// -lgtest -lgtest_main -L/usr/local/cuda-11.1/lib64/ -lcuda -lcurand -lcuda

unlink runTest_tmp.cu
```

In summary, I would go for 
- test.o and test.exe built for c++ with gcc
- gtest.o and gtest.exe built for cuda with nvcc

@hageboeck can you implement that? Thanks
Andrea

[...]

Issue #82, opened 2020-12-03
Implement parallel loops in C++ version of CUDA code


hageboeck:
Since events have no data dependencies, multi-threading across events should be possible. Investigate parallel for loops, e.g. `#pragma omp parallel for`.

lfield:
@hageboeck I initially looked at doing exactly that and it would be one approach. For simple events this makes sense. This is more of an objective question. Where is the concurrency and what are we targeting.

hageboeck:
It's to get away from this:
A: We compared the GPU with a single thread on a CPU. It's x times faster.
B: But that's not fair. You should compare to multi-threading on a modern CPU.

The objective is to grab the low-hanging fruit by just doing that where appropriate:
```diff
  #ifndef __CUDACC__
    //** START LOOP ON IEVT **
+  #pragma omp parallel for
    for (int ievt = 0; ievt < nevt; ++ ievt)
  #endif
    {
  #ifdef __CUDACC__
      const int idim = blockDim.x * blockIdx.x + threadIdx.x;  // event# == threadid (previously was: tid)
      const int ievt = idim; 
      // printf( "sigmakin: ievt %d\n", ievt );
  #endif
```

valassi:
Hi Stephan, thanks, very good! 

As I was proposing last Monday, CPU multi-threading with each thread running a different sets of events is what we should aim for. In principle there should be no thread safety issue (or if there is, we should fix it). We should try to fill up a CPU when we compare GPU to CPU, and a single MT application seems better than different instances of the application.

What I was initially thinking of was simple pthread threading (e.g. c++11 threads as you suggested). 

But your later suggestion for #pragma omp parallel for seems even better to me! First it would be less intrusive on the code, second the level of multithreading can be set by using OMP_NUM_THREADS (otherwise we would need to pass yet another command line argument).

My point in raising the issue on Monday was also that this would allow us to compare to the abstraction layers like Alpaka et al, which on the CPU probably do exactly this. Doing the MT directly in our c++ would be faster and probably easier, especially if we only need a single line of OpenMP.

The line you added was exactky where I was going to suggest it ;-)
https://github.com/madgraph5/madgraph4gpu/blob/9ebd056d2ff792bd027088526f5f1a870d505826/epoch1/cuda/ee_mumu/SubProcesses/P1_Sigma_sm_epem_mupmum/CPPProcess.cc#L706

Very good, make a PR and let's see... :-)
Andrea

[...]

PR #84, opened 2020-12-03
Implement multi-threading (#82). Uses OpenMP as suggested by @hageboeck.

valassi:


valassi:
This PR includes two things:
- a simple implementation of multithreading proposed in #82 (one line in cpp and one line in Makefile!), achieving immediately a x4 speedup on a 4-core machine, using the openmp suggested by @hageboeck with the build instructions of @lfield 
- a workaround for #83, commenting out the build of runTest.exe

It is not surprising that the CI complains for a failed check: it tries to execute runTest.exe but runTest.exe was not built.

It would be best to fix #83 first, separating two sets of .o and .exe for c++ and cuda.

valassi:
I have now also added a printout. 

And I disabled OMP if OMP_NUM_THREADS is not set (for backward compatibility, and to make the behavious more explicit).

lfield:
>     * a simple implementation of multithreading proposed in #82 (one line in cpp and one line in Makefile!), achieving immediately a x4 speedup on a 4-core machine, using the openmp suggested by @hageboeck with the build instructions of @lfield

@valassi to use your own words against you, the x4 speedup is not correct. You have to compare with 4 instances of the single threaded code :) I believe that is what you said to me when we were discussing my performance figures :)

valassi:
> > ```
> > * a simple implementation of multithreading proposed in #82 (one line in cpp and one line in Makefile!), achieving immediately a x4 speedup on a 4-core machine, using the openmp suggested by @hageboeck with the build instructions of @lfield
> > ```
> 
> @valassi to use your own words against you, the x4 speedup is not correct. You have to compare with 4 instances of the single threaded code :) I believe that is what you said to me when we were discussing my performance figures :)

Ah! Then I understand why we were not understanding each other! ;-)
No I do not think I ever meant to say this to you, but we certainly misunderstood each other then.

OK: a factor x4 speedup with 4 OMP threads, with respect to a single thread, all in one copy ;-)
Then I assume we would get the same factor x4, maybe a bit less, by running 4 instance 1xMT.

Anyway, this is precisely the type of studies we do in the benchmarking WG, so we will certainly make both options. The idea is to plot throughput vs "level of parallelism", whichever way you achieve it, be it with 4 1xMT copies, with 1 4xMT copy, with 2 2xMT copies (then typically the three dots more or less overlap on throughput vs parallelism plots.

lfield:
As far as I understand the Open MP version will run slightly slower due to the overhead of setting up the threads. In any case, as you pointed out, this is just the multi-core scenario which should be well understood.

valassi:
I reenabled the tests. It was as simple as adding -lgomp. This is something that nvcc digests, while it does not understand -fopenmp.

This is now ready to be merged. Any objections? Thanks

lfield:
I have tested this code and it seems to work. I don't get a huge speedup. It is ~8.5s for 4M events with 1 OMP Thread and 9.5s with 4 OMP Thread. The code currently does not let you run with less than 4 threads so the so maybe it is using more than 1 thread even if OMP Thread is set to 1. Happy to merge.

lfield (approved):

lfield:
It looks like OneAPI now [supports OpenMP](https://software.intel.com/content/www/us/en/develop/documentation/get-started-with-cpp-fortran-compiler-openmp/top.html) and offloading to supported devices

epoch1/cuda/ee_mumu/SubProcesses/Makefile
```diff
@@ -70,19 +70,19 @@ $(LIBDIR)/lib$(MODELLIB).a:
 	$(MAKE) -C ../../src $(MAKEDEBUG)
 
 gcheck.o: gcheck.cu *.h ../../src/*.h ../../src/*.cu
-	$(NVCC) $(CPPFLAGS) $(CUFLAGS) -c $< -o $@
```
hageboeck, line 73: This violates the [Makefile conventions](https://www.gnu.org/software/make/manual/html_node/Command-Variables.html), which would allow users to add a C++ flag without breaking the makefile.

hageboeck (changes requested):
Hi Andrea,
looks almost good to me.

The comment-out-comment-in commits in the Makefile should be squashed.
Maybe you can also remove all those merge commits, so the feature applies with only three commits or so, but that's mostly cosmetics.

epoch1/cuda/ee_mumu/SubProcesses/Makefile
valassi: Hi Stephan, thanks but sorry I do not get it. What are you suggesting exactly? Are you suggesting that CPPFLAGS should be added back? Thanks Andrea

valassi: Ah yes I imagine that this is waht you meant. OK will do. Thanks!
Andrea

valassi: Ok done, thanks

valassi:
I am finally merging this. Thanks for all the feeback. I also fixed a few bugs at the end.

Reminder: if OMP_NUM_THREADS is not set, this is reset to 1, for backward compatibility. If you want to set it to the maximum, just check "nproc --all".

PR #86, opened 2020-12-04
Initial SYCL version

lfield:
This is the initial commit of the SYCL port of ee_mumu for epoch1. The differences with the CUDA version can be seen by [comparing with the development branch](https://github.com/madgraph5/madgraph4gpu/compare/master...lfield:epoch1_sycl). 

To build  with OneAPI, run the following commands from the ee_mumu directory. 
```source /opt/intel/oneapi/setvars.sh
cmake -B build
cmake --build 
```


tjburch:
Hi @lfield - I was able to get the code running yesterday with reasonable output. Will take a look explicitly at the code itself asap.

tjburch (approved):
Hi @lfield 

LGTM, I can confirm I can run it (using dpcpp compiler). I didn't see anything too controversial. Wasn't sure where to add comments because the merge shows the cuda, but we're interested in the sycl you linked. 

The only question I had was in `check.cc` there were some conversions to unsigned ints that weren't in the cuda. I don't think that's sycl specific so maybe we should propagate that downward to the cuda.

lfield:
@tjburch if it is something that needs changing before this PR can be approved, the comments should be put here. The link to the epoch1_sycl branch, which is done on top of the cuda directory, has only been provided to show the signal from the noise in this PR. I would suggest an issue is created to change the request changes to the CUDA version.

tjburch:
Ok - sure, the SYCL code all looks good to me and ready to merge (happy to push the button if you prefer me to). 

The `unsigned int`s in `check.cc` seem relatively uncritical but you're right, best to make a separate PR for those in the CUDA version.

lfield:
Yes, please push the button.

Issue #85, opened 2020-12-04
Heterogeneous MadGraph: parallel CPU+GPU executions


valassi:
Triggered by the OMP work on c++ (#82) and by the build issues this cause (#83), I realised that eventually we WILL actualy need to run on CPU and GPU in parallel in a "heterogeneous" Madgraph.

This is actually one of the most interesting (and challenging) issues for resource benchmarking in the HEPIX WG.

There are various issue to address here:
- build issues (what to build with which compiler)
- code structure issues (how to best structure the relevant functions and classes - the current code needs a good cleanup anyway)
- configuration issues (how to pass the approriate command line arguments)
- optimization issues (how to share the load, knowing that the GPU is faster)
- random number issues (how to choose different sequences on CPU and GPU)
- combination issues (eg how to compute a single cross section, or output a single file of unweighted events)

I created a simple prototype in https://github.com/valassi/madgraph4gpu/tree/het. The point here was mainly a proof of concept, also trying to sort out the build (which may be useful for addressing #83). 

The current prototype runs exactly the same number of events with exactly the same random numbers in parallel on CPU (with OMP threads) and on GPU. Both computations give the same sets of events, which are not yet combined. As the GPU is much faster, essentiually the net effect is a computation that lasts as long as the CPU version, but does double events (because the same events are also on the GPU), so the throghput doubles. 

This clearly needs a lot more work (especially the optimization is tricky), but it's a useful prrof of concept.

In https://github.com/valassi/madgraph4gpu/commit/7ceb8d581b13635b0d6beadc7bc4f802d482b4f6
```
./hcheck.exe -p 16384 32 1
***********************************************************************
NumBlocksPerGrid           = 16384
NumThreadsPerBlock         = 32
NumIterations              = 1
-----------------------------------------------------------------------
FP precision               = DOUBLE
Complex type               = THRUST::COMPLEX
RanNumb memory layout      = AOSOA[4]
Momenta memory layout      = AOSOA[4]
Random number generation   = CURAND DEVICE (CUDA code)
Wavefunction GPU memory    = LOCAL
-----------------------------------------------------------------------
NumIterations              = 1
TotalTime[Rnd+Rmb+ME] (123)= ( 7.312938e-03                 )  sec
TotalTime[Rambo+ME]    (23)= ( 6.714818e-03                 )  sec
TotalTime[RndNumGen]    (1)= ( 5.981200e-04                 )  sec
[...]

PR #87, opened 2020-12-04
Heterogeneous MadGraph: parallel CPU+GPU executions

valassi:
See a more complete description in #85

I created a simple prototype. The point here was mainly a proof of concept, also trying to sort out the build (which may be useful for addressing #83).

The current prototype runs exactly the same number of events with exactly the same random numbers in parallel on CPU (with OMP threads) and on GPU. Both computations give the same sets of events, which are not yet combined. As the GPU is much faster, essentiually the net effect is a computation that lasts as long as the CPU version, but does double events (because the same events are also on the GPU), so the throghput doubles.

This clearly needs a lot more work (especially the optimization is tricky), but it's a useful prrof of concept.

PS I forgot to mention: this includes and supersedes #82. As in that one, the build of runTest (#83) is disabled here. I would suggest however to fix it after including these changes, which give a possible direction for how to combine cuda and c++ modules. One option for #83 is to keep a single runTest.exe as it is now, but make it much clearer that real modules are either c++ with gcc or cuda with nvcc, and then only thin layers integrating both are compiled with nvcc. Essentially, I only had to add -lgomp to build the combined module, probably the same is ok for runTest.


valassi:
Hm I have actually pushed more changes to the branch, but they do not show up here, I am not sure why...

valassi:
PS Consider approving/merging #84 first (OMP multi threading), as that is included here too.

valassi:
Hm I deleted and recreated the branch, I thought this would be picked up.
I am reopening this after git push -f origin 9026441:het

valassi:
This PR is now obsolete and I will close it.

It is replaced by PR #159.

I copied a few relevant comments to the general issue #85.

PR #89, opened 2020-12-06
Minor fixes (add -Wextra, fix main return etc)

valassi:


valassi:
These should be non controversial changes. I merge them.

The -Wextra flag gives a warning in runTest, which to me seems like a real issue to be fixed.

PR #93, opened 2020-12-08
macro to add newly created issues automatically to the Lounge project

roiser:
taken from https://github.com/qmacro/action-add-issue-to-project-column

thanks @hageboeck

### DIFF
--- a/.github/workflows/newissues2lounge.yml
+++ b/.github/workflows/newissues2lounge.yml
@@ -1,0 +1,16 @@
+on:
+  issues:
+    types: opened
+
+jobs:
+  assign_issue_to_project:
+    runs-on: ubuntu-latest
+    name: Assign new issue to project
+    steps:
+    - name: Create new project card with issue
+      id: list
+      uses: qmacro/action-add-issue-to-project-column@v1
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
+        project: 'Issue Lounge'
+        column: 'Atrium'


hageboeck (commented):
Looking good in principle, but the hard-coded names are maybe not a good idea. 
IIRC, you can give project numbers (2 in this case), because they are stable against renaming.

You get the number from opening the project and checking the URL.

hageboeck:
Can still merge it, though, and test. It won't do harm.

roiser:
makes sense to be safe against renaming (I did already once). So the value of the variable is "project2" ?

hageboeck:
I'm reading up on it. It might just be 'project: 2', but I haven't found it.

hageboeck:
Maybe false alarm. The [javascript code](https://github.com/qmacro/action-add-issue-to-project-column/blob/master/index.js) explicitly retrieves by name. No idea if `2` will count as a valid alias for the actual name.

.github/workflows/newissues2lounge.yml
```diff
@@ -0,0 +1,16 @@
...
+  assign_issue_to_project:
+    runs-on: ubuntu-latest
+    name: Assign new issue to project
+    steps:
+    - name: Create new project card with issue
+      id: list
+      uses: qmacro/action-add-issue-to-project-column@v1
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
+        project: 'Issue Lounge'
```
hageboeck, line 15:
```suggestion
        project: 'Issue_Lounge'
```

hageboeck: ?

hageboeck (approved):

.github/workflows/newissues2lounge.yml
roiser: I went the other way round and renamed the project itself

PR #99, opened 2020-12-09
Extend reference file for CUDA tests

### LINKED ISSUE #90: Warning in runTest when -Wextra is added
Hi @hageboeck ,  I have added -Wextra (mainly to debug vectorization, but not only).

There is now a warning in runTest builds
```
runTest.cc: In lambda function:
runTest.cc:325:66: warning: unused parameter ‘evtNo’ [-Wunused-parameter]
       auto dumpParticles = [&](std::ostream& stream, std::size_t evtNo, unsigned precision, bool dumpReference)
                                                      ~~~~~~~~~~~~^~~~~
```

Looking at the code, indeed I do not understand well the use of both ievt and evtNo in the two lambdas. In any case evtNo is indeed not used.

Can you please fix the code accordingly?
Thanks
Andrea

### LINKED ISSUE #91: build error in runTest with large values of neppR
Hi Stephan, just a question: I am doing some tests for vectorization, changing the values of neppR and neppM, sometimes beyond 32. I have seen that runTest has an hardcoded gputhreads of 32, which then fails if neppR is 64 for instance:

```
runTest.cc: In member function ‘virtual void CPUTest_eemumu_Test::TestBody()’:
runTest.cc:282:35: error: static assertion failed: ERROR! #threads/block should be a multiple of neppR
   static_assert( gputhreads%neppR == 0, "ERROR! #threads/block should be a multiple of neppR" );
                  ~~~~~~~~~~~~~~~~~^~~~
runTest.cc:285:35: error: static assertion failed: ERROR! #threads/block should be a multiple of neppM
   static_assert( gputhreads%neppM == 0, "ERROR! #threads/block should be a multiple of neppM" );
                  ~~~~~~~~~~~~~~~~~^~~~
runTest.cc: In lambda function:
```

Can the value be changed to 64 or above, or would this imply recalculating some reference values?

This is a minor issue, I will stay within 32 for the moment (and I most likely do not need to go beyond), but just to know... Thanks!
Andrea

hageboeck:
To help Andrea test various memory layouts for vectorisation, extend block size to 128 in tests.

Fix #90 
Fix #91


valassi:
Thanks! I should say I am not routinely running this test, but will at some point. When I do run it, sometimes I get failures, even when my visual test .check.exe and gcheck.exe gives consistently the same result. I will file an issue at some point if I have a specific branch where this needs a look.

PR #109, opened 2020-12-16
Refactor tests to be independent of epoch and abstraction library used.

hageboeck:
Refactor testing code, make it independent of CUDA and Epoch1.
To run their own test, each epoch/abstraction implements a test driver
interface, and instantiates a test suite defined in test/src.

In more detail:
- Move reference file for process 1 in eemumu to test/.
- Split framework-specific parts and generic test code in epoch1.
- Create base test class as general interface for madgraph runs.
- Move googletest and common testing code into test/.
- Instantiate epoch1 cuda test for eemumu.
- Port tests to epoch2

@roiser Check the Makefile the process subdirectory. As you said, they compile their own version or `MadgraphTest.o`, although this file doesn't belong into the process subdirectory.


test/include/MadgraphTest.h
```diff
@@ -0,0 +1,110 @@
...
+ * }
+ */
+template<typename FpType>
+class TestDriverBase {
+ public:
+  using fptype = FpType;
+  unsigned int nparticle = 4;
+  static constexpr unsigned int np4 = 4;
+  static constexpr unsigned int niter = 2;
+  static constexpr unsigned int gpublocks = 2;
```
roiser, line 42: For those variables it could be nice to get them out of the source and into some "config file", maybe also make a distinction between CPU and GPU execution? It's fine for me to create an issue for that and merge as such though.

test/src/MadgraphTest.cc
```diff
@@ -0,0 +1,159 @@
...
+  }
+  return referenceData;
+}
+
+
+TEST_P(MadgraphTestDouble, eemumu)
+{
+  // Set to dump events:
+  constexpr bool dumpEvents = false;
+  constexpr fptype toleranceMomenta = std::is_same<fptype, double>::value ? 5.E-12 : 1.E-5;
```
roiser, line 68: Same here, could the precision be taken out of the source and made available in a config file? (issue for now is fine)

roiser (commented):
As can be seen in two comments I would suggest to take some of the varibles out of the source code and make them available in some config file (c++ macros should be fine). But it's really a minor thing, we can create an issue and merge as such to keep going.

Issue #79, opened 2020-12-01
Overflow error in gg_ttgg example


tjburch:
In my conversion to SYCL, I noticed I was getting some NAN values for components (V2[5] specifically) in the `VVV1P0_1` function in CPPProcess. I tried to mirror the behavior in CUDA, and found what appears to be an overflow error, tracing back to the `vxxxxx` function.

Looking at 293 here - https://github.com/madgraph5/madgraph4gpu/blob/master/epoch1/cuda/gg_ttgg/SubProcesses/P1_Sigma_sm_gg_ttxgg/CPPProcess.cu#L293

If I print the components of this square root function:

```
========= sqrt Components:  p[1]=-958189386
========= sqrt Components:  p[1]*p[1]=-1679619596
========= sqrt Components:  p[2]=-1982430220
========= sqrt Components:  p[2]*p[2]=-2100755297
========= sqrt Components:  p[3]=-2045069892
========= sqrt Components:  p[3]*p[3]=-1131260949
========= sqrt Components:  vmass=0
========= sqrt Components:  vmass*vmass=0
========= sqrt Final Argument:  arg=-1689285092
```

There's already overflows in the individual components, and then that propagates to the square root function, which gets an unexpected negative value.

(@oliviermattelaer and @roiser, labeled you as assignees just to see if there were ideas here)

lfield:
@tjburch This sounds like the issue that I discovered when I did the first SYCL port in July. The [ranmar function](https://github.com/madgraph5/madgraph4gpu/blob/master/epoch1/cuda/gg_ttgg/src/rambo.cc#L9) in rambo.cc is not thread safe. The variables jranmr and iranmr are shared by the threads and the equal to zero check for those variables can fail due to a race condition where the value becomes less than zero. The simple fix is to change the equals to less than and using the common random numbers approach will also work until we find a general solution for random number generation.

oliviermattelaer:
@lfield good catch.

tjburch:
Thanks @lfield, good to know.

>using the common random numbers approach

I'm not familiar, what do I need to implement this?

hageboeck:
For common random numbers, you can include this
https://github.com/madgraph5/madgraph4gpu/blob/master/tools/CommonRandomNumbers.h,
and call one of the functions.

The functions should be thread safe in the sense that you can call them as many times as you want, but you have to say in advance how many numbers you need per batch, and you have to set up the seed manually if you don't generate all batches in advance.

[...]

PR #112, opened 2021-01-08
sycl port of gg_ttgg for epoch 1 

tjburch:
This is the closest true-to-original port of epoch 1 that I've worked on. This will compile but will not run, due to the persistence of ranmar which isn't thread safe (see issue #79).

I figured that the best option here was to get the conversion to be true to the original cuda, even if not running, and let the fixes for ranmar and others be addressed in the epoch2 code.

Another stylistic note is that I prefer the functor approach to writing and submitting kernels, so this has been used here, so sigmakin is a class rather than just a function.


tjburch:
Actually, I just realized I used USM here, and I know we were trying to disentangle USM and non-USM so maybe I ought to move this over to the oneAPI directory.

tjburch:
Replacing with #113

PR #113, opened 2021-01-08
oneapi port of gg_ttgg for epoch 1

tjburch:
Replacement of PR #112, moving to the oneapi directory since this uses the unified shared memory paradigm.

Text from that PR:

> This is the closest true-to-original port of epoch 1 that I've worked on. This will compile but will not run, due to the persistence of ranmar which isn't thread safe (see issue #79).
> 
> I figured that the best option here was to get the conversion to be true to the original cuda, even if not running, and let the fixes for ranmar and others be addressed in the epoch2 code.
> 
> Another stylistic note is that I prefer the functor approach to writing and submitting kernels, so this has been used here, so sigmakin is a class rather than just a function.


lfield:
When compiling I get the following error. 
```
/opt/intel/oneapi/compiler/2021.1.2/linux/bin/../include/sycl/CL/sycl/handler.hpp:792:5: error: no matching function for call to object of type 'const sigmaKin'
    KernelFunc(
```

tjburch:
Ah right - that was something that broke in newer oneAPI versions (after October or so?). The recent merge should sort that out, compiles with the latest.

lfield:
Now builds but crashes.
```
sm model couplings independent of event kinematics:
Segmentation fault (core dumped)
```

tjburch:
Right - that crash is a direct result of #79. As of now, I was planning on leaving the port true-to-original, since the original does have the thread-unsafe ranmar in it, and resolving in epoch2 (where ranmar is no longer in the code)

lfield:
ok. I would suggest to add a _known issues_ file in the gg_ttgg directory stating this and then I will merge.

tjburch:
Done.

lfield:
Thanks.

PR #122, opened 2021-03-11
MadEvent eemumu gridpack and some analysis (text/plots)

valassi:
These are just two independent directories with one MadEvent eemumu gridpack and some analysis (text/plots).
There should be no conflicts, at most some repetition with what is in the fortran directoriws.


valassi:
I am self-merging. I'd like the dat ato show up especially. Then we can always rediscuss whether to move/rename this somewhere else rather than as 'gridpack'. The point is that this is small changes to the existing production madevent, rather than the 'from scratch' development we are doing elsewhere based on stanadalone applications.

valassi:
@roiser @oliviermattelaer I am self merging this, there are no conflicts, at most we van rename this elsewhere.

PR #124, opened 2021-03-13
A few more madevent changes

valassi:


valassi:
Self merging, does not affect anyone else

PR #127, opened 2021-03-15
Dump the gcc and nvcc version

valassi:
A minor patch to dump the gcc and nvcc version of the ME calculation module.
Note that there is a factor 2 in the C++ between gcc8 and gcc9.


valassi:
All ok (and non controversial I think), self merging

Issue #117, opened 2021-02-17
Evaluate performance speedup from gcc8 to gcc9 for C++


valassi:
This is a followup to https://github.com/madgraph5/madgraph4gpu/issues/116#issuecomment-780071843

On a very old version of the code, while trying to understand another issue (BEFORE I understood it is actually a hardware problem on a VM, tsc clock not used, hence large overhead of system calls), I tried to evaluate if gcc9 could fix the issue. On the buggy hardware node this had no effect, but on a good node instead gcc9 was a factor 1.5 faster than gcc8 (5.5E5 throughput instead of 3.5E5).

Again, this was on an old version of the code. But one should understand if this speedup also exists for more recent versions - or in any case it would be nice to understand what is going on.

valassi:
A flamegraph is again useful.

This is pmpe04 with gcc8
```
time ./check.exe 393216
***********************************
NumberOfEntries       = 393216
TotalTimeInWaveFuncs  = 1.023022e+00 sec
MeanTimeInWaveFuncs   = 2.601678e-06 sec
StdDevTimeInWaveFuncs = 5.867377e-07 sec
MinTimeInWaveFuncs    = 2.242000e-06 sec
MaxTimeInWaveFuncs    = 9.060000e-05 sec
-----------------------------------
NumMatrixElements     = 393216
MatrixElementsPerSec  = 3.843672e+05 sec^-1
***********************************
NumMatrixElements     = 393216
MeanMatrixElemValue   = 1.372012e-02 GeV^0
StdErrMatrixElemValue = 1.307144e-05 GeV^0
StdDevMatrixElemValue = 8.196700e-03 GeV^0
MinMatrixElemValue    = 6.071582e-03 GeV^0
MaxMatrixElemValue    = 3.374887e-02 GeV^0

real    0m1.667s
user    0m1.617s
sys     0m0.046s
```
![perf-pmpe04-gcc8](https://user-images.githubusercontent.com/3473550/108188881-6900af00-7110-11eb-9553-9c4f17d2b6bb.png)

This is pmpe04 with gcc9
```
time ./check.exe 393216
***********************************
NumberOfEntries       = 393216
TotalTimeInWaveFuncs  = 7.300220e-01 sec
MeanTimeInWaveFuncs   = 1.856542e-06 sec
StdDevTimeInWaveFuncs = 4.367662e-07 sec
MinTimeInWaveFuncs    = 1.561000e-06 sec
[...]

PR #128, opened 2021-03-17
Fast math in fortran and c++

valassi:
Move the default for both Fortran and C++ to fast math. This is already used in CUDA.

Note also some discussions in
https://github.com/madgraph5/madgraph4gpu/issues/117
https://bugs.launchpad.net/mg5amcnlo/+bug/1919477


valassi:
Self merging, should be non controversial

PR #130, opened 2021-03-18
Add a few comments about build options

valassi:
Add a few comments about build options
(and also a test of restrict, with no impact)


valassi:
Self merging

PR #131, opened 2021-03-18
Ep2 gg ttgg

lfield:
First attempt at SYCL port. Runs on OneAPI CPUs and GPUs


lfield:
Runs on the V100 too.

PR #133, opened 2021-03-20
Enable C++-only builds on a system which does have nvcc

valassi:
Enable C++-only builds on a systema which does have nvcc

It is enough to "export CUDA_HOME=invalid" and only the C++ will be built
This enables emulating the CI tests


valassi:
I will self merge. This is part also of klas PR #72 but I want to test gcc10 before merging vectorization, I have an issue in gcc10 to debug.

PR #132, opened 2021-03-19
Klas2 - further extensions for SIMD and related work 

valassi:
Klas2 - further extensions for SIMD and related work 

This includes minimal changes to allow clang builds (initially used to test SIMD, which however is disabled because compiler vector extensions do not work in that case). This is WIP.

This whole PR is WIP.

PR #137, opened 2021-03-24
Makeinfo

valassi:
Add 'make info' to the github CI. Add host information to 'make info' in epoch1. 

This is useful to debug issues in vectorization. I had avx512 builds by default, but I was getting these errors
https://github.com/madgraph5/madgraph4gpu/pull/132/checks?check_run_id=2173809398
```
Run make check
Makefile:64: CUDA_HOME is not set or is invalid. Export CUDA_HOME to compile with cuda
build.avx512/runTest.exe
make: *** [Makefile:136: check] Illegal instruction (core dumped)
Error: Process completed with exit code 2.
```

I fixed this by using avx2 by default if the Makefile finds that avx512 is not supported, but I want to debug further related issues.

Also, its useful to print th envcc and gcc versionof the CI.

I am extracting these commits out of the klas barnch as a standalone PR

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -14,6 +14,8 @@ jobs:
         working-directory: epoch1/cuda/ee_mumu/SubProcesses/P1_Sigma_sm_epem_mupmum
     steps:
     - uses: actions/checkout@v2
+    - name: make info
+      run: make info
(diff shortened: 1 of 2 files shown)


valassi:
These are useful a priori for all developments. I will self merge.

Issue #139, opened 2021-03-28
Move latest eemumu developments from epoch1 to epoch2 ("merge" epoch2 into epoch1)


valassi:
Hi @roiser @oliviermattelaer @hageboeck 

All my latest developments (master, klas/vectorization, heterogeneous, unweighting etc) are all in epoch1/eemumu. It was my mistake not to base this on epoch2/eemumu. 

Luckily, the changes that have been made in epoch2/eemumu since its initial creation are relatively few.

As agree with @roiser on Friday, I will essentially merge those epoch2 changes into epoch1. I proposed to also upgrade epoch2 to the level of epoch1, but we agreed that it is better to use epoch1/eemumu as the basis, and eventually use this to backport to MG and create epoch3.

We also agreed that I will self-merge essentially all of this stuff, as there is quite a bit to be done. I will try to document here what I am doing and why.

valassi:
There are essentially two things to be ported
1. The initial differences between epoch2 and epoch1 when epoch2 was created.
2. The new changes on top of epoch2 over time.

For reference, the latter are these, quite few

[avalassi@itscrd70 bash] ~/GPU2020/madgraph4gpuTer/epoch2/cuda/ee_mumu> git diff upstream/master
[avalassi@itscrd70 bash] ~/GPU2020/madgraph4gpuTer/epoch2/cuda/ee_mumu> git log .

commit ff021bdfafa5068b841fbd1754052524270301b2
Author: Stephan Hageboeck <stephan.hageboeck@cern.ch>
Date:   Wed Dec 16 18:22:26 2020 +0100

    [ep2 cuda eemm] Port fixes in Makefile to epoch2.

commit 57497a79292cf0616e04ab9b866bba305eb93f54
Author: Stephan Hageboeck <stephan.hageboeck@cern.ch>
Date:   Wed Dec 16 17:05:24 2020 +0100

    [ep2 cuda eemm] Port CUDA tests to epoch2.

commit 05d3a6f8de663db276b4755785d0713b27b043bd
Author: Olivier Mattelaer <olivier.mattelaer@uclouvain.be>
Date:   Wed Dec 2 13:52:59 2020 +0100

    port into MG5aMC the change from https://github.com/madgraph5/madgraph4gpu/pull/78

commit a6c18e2715bfa5b39727ba6407031f6c7633ab78
Author: Olivier Mattelaer <olivier.mattelaer@uclouvain.be>
[...]

PR #140, opened 2021-03-28
Clean up and rename files in epoch2/eemumu

valassi:
Clean up and rename files in epoch2/eemumu. This is the first batch of changes for issue #139 

These changes are all in epoch2, essentially:
- cleanup: remove duplicate rambo.cc and grambo.cu (were identical)
- cleanup: remove duplicate gcheck_sa.cu and ../check.cc (they were different and only the former was used - but keep the latter only, which is more recent and closer to epoch1)
- use the same naming strategy as in epoch1: keep code in .cc files, define .cu files as symlinks (NB eventually we could get rid of symlinks and use the nvcc option to treat cc as cu, used by @hageboeck in the tests... more generally we should get rid of these gXXX filenames that I introduced)
- (temporarely?) rename check_sa.cc as check.cc, to ease comparisons with epoch1 (NB eventually I agree to go back to check_sa as a name, but now it just complicates things)


valassi:
Ok all issues fixed - self merging

PR #141, opened 2021-03-28
[epoch2] Enable C++-only builds on a system which does not have nvcc

valassi:
This is the same for epoch2 as PR #133 was for epoch1

It is needed to fix the CI test failure in the initial PR #140 version


valassi:
CI tests now ok. Self merging

Issue #143, opened 2021-03-30
test code does not build in single precision


valassi:
The test code does not build in single precision. This is true both in epoch2 and epoch1. I fixed some small related issues in epoch2 for isse #139 and found that there are issues also in epoch1.

I will fix this for both epochs.

Eventually it would also be nice to test the single precision builds in the CI - but we need a few additional changes in the Maekfiles too,

valassi:
This is done and can be closed.  I have fixed the build failures of the tests in single precision, but I essentially didabled into no-ops.

There are two things left to do:

 -   One, implemente the tests beyond MadgraphTestDouble. This should become a templated MadgraphTestFptype, but then passing the template arguments through googletest's MG_INSTANTIATE_TEST_SUITE_GPU takes some time,
 -   Two, massage the Makefiles to pass the single/double choice from outside, and add a single precision build and test in the CI.

I have opened issue #148 to follow up on these.

Issue #129, opened 2021-03-18
Check the handling of NaN/Inf (both in Fortran madevent and in our standalone C++/CUDA) 


valassi:
I open this issue as a reminder: check the handling of NaN/Inf (both in Fortran madevent and in our standalone C++/CUDA).

This is related to issue #117 about the use of fast math for complex number arithmetics. The best speed in both Fortran and C++ is obtained using the most aggressive compiler options, including -O3 and fast math. This disables IEEE 754 compliance for NaN and Inf amongst other things. It is not clear what would happen if one event out of many has a NaN or Inf. In the standalone C++ driver, this was correctly handled (the event is just disacraded) before fast math, but one should check if fast math changes this. In madevent instead it's not clear how this would be handled.

This is also related to issue #5 about single precision. In the standalone C++/CUDA we observed that single precision does lead to the appearance of one NaN every million events or so. One could check for instance what happens when both single precision and fast math are used.

valassi:
Good that I opened this issue.

Indeed, there is an issue due to fast math. In the context of issue #143 I tried to build and test in single precision.

I get one run that fails with fast math, but no longer fails without fast math:

```
With fast math
    time ./check.exe -p 2048 256 12
    ***********************************************************************
    NumBlocksPerGrid            = 2048
    NumThreadsPerBlock          = 256
    NumIterations               = 12
    -----------------------------------------------------------------------
    FP precision                = FLOAT (nan=0)
    Complex type                = STD::COMPLEX
    RanNumb memory layout       = AOSOA[8]
    Momenta memory layout       = AOSOA[8]
    Random number generation    = COMMON RANDOM (C++ code)
    OMP threads / `nproc --all` = 1 / 4
    MatrixElements compiler     = gcc (GCC) 9.2.0
    -----------------------------------------------------------------------
    NumberOfEntries             = 12
    TotalTime[Rnd+Rmb+ME] (123) = ( 7.872868e+00                 )  sec
    TotalTime[Rambo+ME]    (23) = ( 7.788832e+00                 )  sec
    TotalTime[RndNumGen]    (1) = ( 8.403606e-02                 )  sec
    TotalTime[Rambo]        (2) = ( 1.559249e+00                 )  sec
    TotalTime[MatrixElems]  (3) = ( 6.229583e+00                 )  sec
    MeanTimeInMatrixElems       = ( 5.191319e-01                 )  sec
[...]

PR #144, opened 2021-03-31
[epoch1] Single precision: fix build failures, improve NaN determination

valassi:
This patch is another spinoff from isue #139 about merging epoch2 and epoch1.

At some point I started looking at single precision. This is because in epoch2 there was one hardcoded "double" which I transformed into "fptype". For completeness I tried to build in single precision and this opened up another pandora's box also in epoch1, about single precision. There are two issues, both addressed (partially) by this PR:
- The tests were not building (issue #143): I have made minor fixes, and added a workaround for disabling the test in single precision. This needs much more work (mainly to use a templated test in google test).
- This was the first time I built single precision after turning on fast math (issue #117) and I found out, as might be expected, that there were problems with NaN handling (issue #129). This is another issue that should be addressed much more deeply, by avoiding any NaNs upfront (else the numerical precision of the MC calculation may be unreliable). For the moment, I just made sure that the determination of problematic events with "abnormal" numbers (not exactly NaNs) is better, so that these events are discarded and the final ME average makes sense and is not a nan.


valassi:
Ok I am ready in branch ep2toep1 to include this PR. So I self-merge.

PR #149, opened 2021-03-31
Merge epoch2 and epoch1 - second part (still without CPPProcess)

valassi:
This is the PR to complete issue #139.

I keep it as WIP for now, it's 80% done but still needs a few (quite important as performance-relevant) tweaks.


valassi:
I have decided to split this further into two PR. I have done everything except CPPProcess, but this is the most complex part (and I actually even see a minor performance differences). I will split that out in a third PR.

Recap about issue #139 
- first part, merged PR #140 : clean up and rename files in epoch2/eemumu
- second part, this PR #149 that I will soon merge: ensure that all files in epoch1 and epoch2 are identical in master, except for CPPProcess (and the related HelAmps_sm)
- third part, an upcoming PR: address also the differences between epoch2 and epoch1 in CPPProcess

More in detail about this PR #149 below.

---

In src:

1) Parameters_sm.h 
Remove "using namespace std;" in epoch2. Otherwise almost identical.
Copy epoch2 to epoch1.

2) Parameters_sm.cc
Add explicit std:: in epoch2. Otherwise almost identical.
Copy epoch2 to epoch1.

3) read_slha.h
Identical but for indentation: fix them manually and make them equal.
(clang-format would bring too many changes)

4) read_slha.cc
Identical but for a default parameter value in implementation in epoch2.
Fix by copying epoch1 to epoch2.

5) rambo.h/cc
Identical in epoch2 and epoch1, nothing to do

6) mgOnGpuConfig.h
Identical, except for a comment (did the percent sign disturn the metacode?).
Fix by copying epoch1 to epoch2.

7) mgOnGpuTypes.h
Identical.

8) Makefile
Almost identical, but ep1 has OMP, fastmath, Wextra.
Fix by copying epoch1 to epoch2.

9) HelAmps.h/cc
MISSING IN EPOCH1! Do this later...

---

In SubProcesses and below:

1) timer.h
Identical

2) Makefile
Almost identical but epoch1 has much more, cosmetics and copy ep1 to ep2
Now added also to ep2, as in epoch1: OMP, fastmath, Wextra, clang patch, host info

Note: at this stage, epoch1 is slightly faster than epoch2 in c++, but the inverse in CUDA.

3) Memory.h, nvtx.h, perf.py
Identical (but a symlink is missing, to be added in epoch1)

4) timermap.h
Copy epoch1 to epoch2 to add missing gcc pragmas for nvtx warnings

5) perf/data
Only in epoch1 - one json file, keep it there

6) profile.sh
Only in epoch1 - should bring it forward eventually
(anyway the basis will be epoch1)

7) runTest.cc
Initially identical, but tests had different name (e.g. EP1_CUDA_GPU vs EP2_CUDA_GPU). This is fixed by adding epoch_process_id.h where a different macro is defined per epoch, then runTest.cc is now identical. 

8) check.cc

First batch of changes

Minimal changes in epoch1:
- remove unused headers in epoch1
- remove two empty lines in the code doing the performance dump

Port to epoch2 many changes from epoch1:
- add omp.h in epoch2
- use the ep1 printout about '-d' also in epoch2
- use the ep1 printout about OMP_NUM_THREADS also in epoch2
- export OMP_NUM_THREADS=1 if not set also in epoch2
- initialize T() in hstMakeUnique also in epoch2
- comment out unused stdwtim also in epoch2
- add one space per line in the performance dump also in epoch2
- add OMP info in the performance dump also in epoch2
- [commented out] add gcc compiler info in the performance dump also in epoch2
- return 0 at the end of main also in epoch2

7bis) runTest.cc
8bis) check.cc

A large batch of additional changes (mainly in PR #144) came from fixing epoch2 check.cc to use fptype for random numbers as in epoch1. This triggered many additional checks about single precision, included in PR #144, which also includes a better treatment of NaNs.

---

This is all at the time of this PR (after some previous ones). Then the rest will be about CPPProcess.

valassi:
Self-merging.

Issue #138, opened 2021-03-24
Can we upgrade the github CI node to gcc10 (was gcc9)?


valassi:
Hi @hageboeck I added a 'make info' to the github CI actions (PR #137), mainly to debug vectorization issues on the CI nodes, but not only.

This made me realise that the CI GPU node uses gcc83:
https://github.com/madgraph5/madgraph4gpu/runs/2187324415
```
Run make info
Makefile:17: No CUDA_HOME exported. Using "/usr/local/cuda/"
/usr/local/cuda//bin/nvcc --version
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2020 NVIDIA Corporation
Built on Mon_Oct_12_20:09:46_PDT_2020
Cuda compilation tools, release 11.1, V11.1.105
Build cuda_11.1.TC455_06.29190527_0

g++ --version
g++ (GCC) 8.3.1 20191121 (Red Hat 8.3.1-5)
Copyright (C) 2018 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

Would it be possible to upgrade this to gcc92 at least (even through cvmfs if needed)?

Apart from a better performance, I think that some of my code does not build on gcc8 (I no longer use it), but this is only ion epoch1 so far so we do not see this.

Thanks
Andrea

hageboeck:
Hello Andrea,

it’s the standard nvidia cuda-enabled centos container. If they already updated, I can just rerun the container build script. I’ll have to check tomorrow.
If it’s not up to date, we need to see if that specific centos allows for never compilers.

Cheers,
Stephan

Am 24.03.2021 um 20:17 schrieb Andrea Valassi ***@***.******@***.***>>:



Hi @hageboeck<https://github.com/hageboeck> I added a 'make info' to the github CI actions (PR #137<https://github.com/madgraph5/madgraph4gpu/pull/137>), mainly to debug vectorization issues on the CI nodes, but not only.

This made me realise that the CI GPU node uses gcc83:
https://github.com/madgraph5/madgraph4gpu/runs/2187324415

Run make info
Makefile:17: No CUDA_HOME exported. Using "/usr/local/cuda/"
/usr/local/cuda//bin/nvcc --version
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2020 NVIDIA Corporation
[...]

PR #150, opened 2021-04-01
[tests] Add epoch1_eemumu_GPU and use gcc9 in the CI

valassi:
This is related to issue #139: once I merge epoch2 into epoch1, the latter will be my main reference. However only epoch2 is tested on the GPU so far. Here I am adding epoch1 to GPU tests too.

This probably depends on issue #138 being fixed. I think we need gcc9 to build (not sure if immediately or only when vectorization is there). I will open this as a WIP PR so that the CI can get tested.

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -49,11 +49,11 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: make info
-      run: make info
+      run: source /opt/rh/gcc-toolset-9/enable; make info
     - name: make
-      run: make
+      run: source /opt/rh/gcc-toolset-9/enable; make
     - name: make check
-      run: make check
+      run: source /opt/rh/gcc-toolset-9/enable; make check
   epoch2_eemumu_GPU:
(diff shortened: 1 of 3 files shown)


valassi:
I have tried various options suggested by Stephan (thanks!) for using gcc9 in issue #138
-  Running source /opt/rh/gcc-toolset-9/enable as first command in the CI job or Starting the CI job with scl enable gcc-toolset-9 bash before doing anything else. These do not work if I do it in separate "run" within the same "steps".
-  Running all commands as scl enable gcc-toolset-9 <command>. This failed (but I am not 100% sure if it is because I was on the container without gcc9?...).

In the end I implemented a mixture of the above:
- Running source /opt/rh/gcc-toolset-9/enable before each command in each "run". Ugly but it works.

Eventually it may be best to do it differently, e Stephan's last suggestion:
-  Finding out how we can convince the container to always use that environment.
I can see two ways to do this: either configuring this in the machine itself, or maybe creating a custom github action on which all steps depend. But I have not found out yet how to do it.

I will self-merge this already, the rest can come at a later time.

PR #153, opened 2021-04-02
het + epoch1/epoch2

valassi:
This merges together het #87 and epoch12 #151.

It will replace #87. Open this as WIP.

PR #152, opened 2021-04-02
klas2 (SIMD CPU) + epoch1/epoch2

valassi:
This merges together klas2 #132 (replacing klas #72) and epoch12 #151.

It will replace #72 and #132. Open this as WIP.

PR #151, opened 2021-04-01
Merge epoch2 and epoch1 - third part (CPPProcess and HelAmps)

valassi:
This is the third (final?) PR for issue #139: the idea is to make sure that epoch1 has the same code as, or at least includes relevant code from, epoch2. This is because all my next PRs (vectorization, heterogeneous) are based on epoch1.

This is the followup to PR #140 and #149.

I keep it as WIP for the moment.

PR #159, opened 2021-04-09
het + klas2 + epoch1/epoch2 (Heterogeneous GPU + SIMD CPU)

valassi:
I create this new DRAFT PR about Heterogeneous GPU+CPU, issue #85.

This is essentially about the (very few!) changes initially added in PR #87 (het), which were later merged with the epoch12 work (PR #151) into a new PR #153 (hetep12).

This new PR replaces both #87 and #153, which will be closed. It includes not ony the het and hetep12 (het + ep12) changes, but also the vectorization work (presently in klasep12 PR #152). It is based on a new branch hetklas, which merges hetep12 into klasep12.

I will eventually move here a few comments from the old het/hetep12 PRs.


valassi:
This is a very old MR about hetereogeneous processing for the standalone application, with some MEs on the CPU and some MEs on the GPU.

It is so old that it would need significant reshuffling to solve conflicts, probably.

Now that we are focusing on madevent, and that we know the CPU bottleneck is the madevent non-me fortran component, heterogeneous strategies are better focused on targeting CPU MT for that fortran part.

Cloising this as unmerged

PR #160, opened 2021-04-11
testxxx: bug fixes and tests for the ixx/oxx functions

valassi:
Add tests for the ixxxx/oxxxxx functions in both epoch1 and epoch2

This iwll be useful for testing that the vectorized versions give the same results


valassi:
Thisshould be non controversial, I will self merge

valassi:
@oliviermattelaer @roiser I have completed this and am about to self-merge it.
This is the last merge where epoch1 and epoch2 for eemumu are strictly identical.
This was the last preliminary work before merging also vectorization (where epoch1 will diverge and epoch2 will remain the same).

The point of this testxxx merge was initially to make sure I have tests for the xxx functions in no-simd mode, so that I could test the simd version. This is because I will provide the vectorized version also of all xxx that are NOT used in eemumu (eg sxxxx). However, on the way this was also useful to identify several issues and bugs.

A few details
- I created a few events with all relevant combinations (mass=0 or >0, pz = or > or < 0, pt = or > 0). I ran the test once, created some reference and then the test compares against it. It is similar to what @hageboeck had done, but instead of creating a text file to read in, I create a c++ snippet file that is compiled in the test
- The test only runs for c++ (but the no-simd version of xxx functions uses exactly the same code in cuda)
-  In addition, I explicitly test approximate functions like opzxxx against the general oxxxxx function, and this was one place where I found some bugs for instance
- I often had doubts, which I clarified looking at the Fortran version of xxx functions upstream; in the end, I systematically cross checked ixxxxx, oxxxxx, sxxxxx, vxxxxx against the fortran
- Bug: sometimes "abs" was used instead of "std::abs", with the side effect that it was giving a truncated integere result. I have removed all "using std::abs, std::min, std::max" as a precaution
- In any case I defined fpmin/fpmax functions to replace min/max, as I needed to overload them later for vector types in simd (I also have fpternary and cxternary replacements for "a?b:c" with vectors, but that's the next PR...)
- Bug(?) in ixxxxx: "p0" was calculated as the momentum, rather than as the energy in fortran (not ok if mass is not 0), I used the Fortran algo
- Bug in ixzxxx: fi[0] and fi[1] were using wrong expressions, now fixed
- Bug in omzxxx: fo[5] was wrong in one branch
- All these bugs are actually in code that is not used in eemumu, but I fixed them because I wanted to provide solid xxx functions in the vectorized code
- I did a bit more cleanup, both in code formatting, choice of variable names (to be closer to Fortran or to use something more C++ like and more vectorizable), or also in the algos sometimes (e.g. why calculate the abs value of a sqrt? it is positive alway, right?)
- Most of the changes were irrelevant to performance. Some gained maybe 1-2% (eg using hardcoded zeros for initialisation instead of assigning fi[2]=fi[1] where the latter is zero).

Voila that's all I think. The vectorization PR should come tomorrow at the latest now.

Let me know if you have comments... thanks for the patience
Andrea

PR #165, opened 2021-04-19
[ep12float] port the latest (identical) epoch1/2 from double to float

valassi:


valassi:
This is logically another patch for issue #139 : after resyncing epoch1/2 I had not yet tested for single precision

This patch re-does that port with a few minimal changes. Also testxxx is ported.

This (again!) should be the last PR before including the vectorization patch.

I will self merge.

Issue #148, opened 2021-03-31
Fix single precision tests in the code


valassi:
This is a followup to issue #143, which is done with PR #144 and which I will close.
I have fixed the build failures of the tests in single precision, but I essentially didabled into no-ops.

There are two things left to do:
- One, implemente the tests beyond MadgraphTestDouble. This should become a templated MadgraphTestFptype, but then passing the template arguments through googletest's MG_INSTANTIATE_TEST_SUITE_GPU takes some time,
- Two, massage the Makefiles to pass the single/double choice from outside, and add a single precision build and test in the CI.

Very low priority. Especially, I would do this after converging epoch2/epoch1, which I am slowly doing now.

@hageboeck maybe you are interested? Thanks

valassi:
I am working on this.

I essentially completed the infrastructure for building and running the single precision test. The problem is that I get different results in single and double precision, while the reference output is the same. I imagine that this is simply because we run common random numbers, once in single and ince in double precision. I will hack around that part.

hageboeck:
Maybe one can produce the random numbers in double and convert them to float? You might still see differences to the double implementation, but that would tell us the loss of precision that madgraph would suffer from.

valassi:
Thanks, yes that's exactly what I was trying :-)
I wasted some time in bookkeeping, I also need to produce ref files with nepp=R (in a later branch I hardcode this for all modes anyway). This determines how the random array is interpreted. Almost there I hope

valassi:
Renamed this as "fix single precision tests in the code". I have stripped off the integration in the CI to a separate issue #167.

I am about to merge a PR that fixes this first specific part "fix single precision tests in the code"

valassi:
I just merged #168 for this specific issue. Closing as fixed.

PR #168, opened 2021-04-21
Fix single precision tests in the code

valassi:
This is a fix for issue #148 (or what was originallly just the first part of it, namely fixing the code for tests in singl eprecision... but this PR does not yet integrate them in the CI).


valassi:
All checks ok. I am self merging.

PR #169, opened 2021-04-21
A few more patches for clang11

valassi:
A few more patches for clang11, before merging vectorization (which still has some issues for clang)


valassi:
All ok in the tests - self merging again

PR #170, opened 2021-04-22
[testxxx] bug fix for the sxxxxx test (it uses wavefunction [3] and not [6])

valassi:


valassi:
Discovered while debugging issues in the testxxx tests for vectorization in clang - it does not solve the latter issue, but is probably a necessary fix anyway

Will self merge

Issue #178, opened 2021-04-25
Vectorize color algebra


valassi:
This is one of the main pending items left to-do from the vectorisation issue #71 and the big PR #171.

High priority (possibly to be done before epoch3).

So far, in the vectorisation I have vectorised the xxx functions, and the FFV functions which take as input the xxx outputs. I have also vectorized what comes immediately afterwards, ie the calculation of the jamp partial amplitudes that is done with the amp outputs of the FFV functions.

There is one thing left to be done:
- vectorize the multiplication of the jamp vector by the color matrix, and then by the conjugate jamp vector (i.e., the calculation of the color algebra)

The main reason why I have not done this is that in eemumu this is trivial. However, in QCD processes this may be one of the bottlenecks, taking up more than 50% of the ME calculation! See issue #155. (In addition, I had actually not yet precisely understood what this was all about...).

Should be relatively straightforward. The challenge here is to do it in eemumu (where it is almost a no-op) in a way that is clearly portable upstream to code-generating metacode for QCD processes.

valassi:
As discussed with @roiser @oliviermattelaer @jtchilders there is also one bit before the color algebra that is still missing: there is now one call of FFV for CUDA and one call for C++. The interfaces are the same, it should be possible to easily remove the #ifdef and have a single code implementation valid for both cases there

valassi:
This is now completed with PR #185, which I just merged.

As exepected, this gained a very tiny bit of extra performance (less than 5% for 512y mode)

Before the change: https://github.com/madgraph5/madgraph4gpu/commit/6f4916ec5552f57d8b58520f924c28e1e495673c
```
-------------------------------------------------------------------------
Process                     = EPOCH1_EEMUMU_CPP [gcc (GCC) 9.2.0]
FP precision                = DOUBLE (NaN/abnormal=0, zero=0)
Internal loops fptype_sv    = VECTOR[4] ('512y': AVX512, 256bit) [cxtype_ref=YES]
OMP threads / `nproc --all` = 1 / 4
EvtsPerSec[MatrixElems] (3) = ( 4.753192e+06                 )  sec^-1
MeanMatrixElemValue         = ( 1.371706e-02 +- 3.270315e-06 )  GeV^0
TOTAL       :     3.764206 sec
real    0m3.774s
=Symbols in CPPProcess.o= (~sse4:    0) (avx2: 2604) (512y:   97) (512z:    0)
-------------------------------------------------------------------------
```

After the change: 
```
[...]

PR #185, opened 2021-04-27
Vectorize color algebra + Fixes for cucomplex

valassi:
This implement the last bits of vectorization for color agebra and related things discussed in issue #178 

Since it involved a few changes to operators for complex and vector types, I also tested again cucomplex and added a few new fixes that became necessary with the changes in the code in the last few months


valassi:
Self merging

Issue #182, opened 2021-04-26
Port to clang 11 and clang12


valassi:
A reminder: thank to the SFT team, clang11 and clang12 with OMP are now installed.
https://sft.its.cern.ch/jira/browse/SPI-1875

I should just try them out (and remove the hacks for missing OMP in the previous installation)

valassi:
Possibly fix issue #172 first

valassi:
This is now fully done.

I have tested that the latest installs of clang 11.1 and clang12.0 on cvmfs work fine, and have openmp. I removed the hack to disable openmp in clang 11.0 (only applies to that cvmfs installation).

En passant I tested the vectorisation performance of clang 11.1 and clang 12.0:
- clang 11.1 is similar to clang 10.0: it is a bit faster than gcc9 with vectorization! peak of 5.1E6 against 4.9E6, more or less
- clang 12.0 is (significantly slower) than the other two
- the difference between AVX2 and 512y is very tiny in all cases (avx2 faster in clang10/11, 512y faster in clang12, but slower than the other clangs anyway): for the moment I keep avx2 as the default on clang, unlike gcc where I keep 512y

I have also tested whether clang12.0 allow the use of cxtype_ref (issue #172), i.e. if now operator[] on vector extensions returns a non-const float, but it does not... I keep the other implementation on clang without cxtype_ref

A few minor patches, comments and logs are in PR #187 

This is fully clarified and can be closed

Issue #172, opened 2021-04-23
Validate clang-style "no cxtype ref" vectorization and use it as default


valassi:
This is a spinoff of vectorisation issue #71 and a followup to the big PR #171.

There are presently two slightly different vectorisation implementations
- the original one developed on gcc
- a recent one where I had to tweak a few things for clang

In both implementations (take double with AVX2, ie 4 doubles per vector, as an example
- floats are 4-vectors FFFF
- complex numbers are implemented as two float vectors RRRRIIII (and not as RIRIRIRI)
- compiler vector extensions (those of gcc, or those of clang) are used for vectors

A small difference between the two implementations is the following
- the original gcc version introduces a "cxtype_ref" class that is only a wrapper to two float non-const references, so that cxtype_v[0] returns a reference to R...I... in the RRRRIIII: this is the operator[], https://github.com/madgraph5/madgraph4gpu/blob/dfcc0f96a5d06cad32a3161fc0aa94569359e3b9/epoch1/cuda/ee_mumu/src/mgOnGpuVectors.h#L70
- in clang this is not possible because "non-const reference cannot bind to vector element"
- initially I tried to use in clang an operator[] returning a pair of values, rather than a pair of references, however this led to wrong results (only in the testxxx tests! not in eemumu ME averages??)... the issue is that in some places it was still used as one would use a non-const reference (and I was surprised that the code built at all)
- anyway, in the end, I realised that this operator[] was really only needed in a minimal number of places, so I built a clang implementation where there is no such operator[]
- now all the tests pass, and actually this implementation can only work on gcc, and it is much simpler, so I'd like to use it as default (removing cxtype_ref)

However:
- I observed that the average ME that is printed out is different whether this implementation is used or not
- What is really puzzling is that this is true also for the 'none' implementation (where there are no vectors at all, no compiler vector extensions, no need for any such thing)

So, this issue is just about understanding if there is a bug and where. Maybe I just read the results in the wrong way and there is no issue.

valassi:
Compare the commit logs of these two commits

With cxtype_ref, gcc/double
https://github.com/madgraph5/madgraph4gpu/commit/8edae31eae6b10b9bb5269b567a3ebcf5e91d8e4
```
-------------------------------------------------------------------------
Process                     = EPOCH1_EEMUMU_CPP [gcc (GCC) 9.2.0]
FP precision                = DOUBLE (NaN/abnormal=0, zero=0 )
[...]

PR #187, opened 2021-04-27
Clarify issues with cxtype_ref on gcc/clang + Test clang11.1 and clang12.0

valassi:
Clarify issues with cxtype_ref on gcc/clang + Test clang11.1 and clang12.0

This addresses a mix of two related issues
- issue #172 about cxtype_ref: understood as not a bug, it depends on common rancom numbers (needed for clang11, 12 as cuda11 does not support them)
- issue #182 about testing the latest clang11.1 and clang12.0 with openmp support (removed the hack for clang 11.0 with no openmp in the cvmfs install I use)


valassi:
Almost no changes. Just many comments and logs. Self merging.

Issue #180, opened 2021-04-25
Hard reboot itscrd70 and re-validate performances


valassi:
This is a related to vectorisation issue #71 and to the big PR #171. In particular, it is closely related to issue #173 about AVX512.

In order to measure the processor speed for AVX512, I need a hard reboot of my reference VM to enable CPU pass through. This would more generally allow using perf in a more complete way with hardware counters. In principle, this should not change performances, but one never knows.

An alternative is to rebuild itscrd03 on centos7, with CPU passthrough. This is a downgrade of this VM from centos8, hoping thatthis is enough to address the still-not-completely-understood issue #116.

Just a reminder and a placeholder for (hopefully unnecessary) performance comparisons.

valassi:
Ok I have been using for a few days the new itscrd03 with CPU passthrough (rebuilt on centos7, ok no tsc issue), and it works nicely, with perf stat. So I hard rebooted itscrd70, performance is good, and perf stat is available. Can be closed.

PR #188, opened 2021-04-27
Add perf stat to the throughput script + Minor patches to streamline printouts

valassi:
This is mainly about adding perf stat to allow a better analysis of AVX512 and other vectorization scenarios (issue #173)

This assumes that perf stat works - which I am able to do on itscrd03, and for my older itscrd70 will come after solving issue #180 (a hard reboot of the VM)

I also added a few minor patches to sytreamline printouts


valassi:
Will self merge

Issue #175, opened 2021-04-25
"xxx" function interface: further separation of data access and calculations


valassi:
This is vaguely related to vectorisation issue #71 and the big PR #171.

One question I am considering is whethe we should push even further the separation of data access and calculations.

Presently, the xxx functions 
- take as input an fptype pointer allmomenta for all particles in all events, plus some parameters identifying which particle in which event is needed, and other parameters for the calculation such as mass, helicity, ingoing/outgoing
- give as output the wavefunction for the given particle and even under the given calculation scenario

In practice, all these functions do two things
- they decode the allmomenta AOSOA data structure using the pIparIp4Ievt and pIparIp4Ipag functions, extracting only the 4-momenta for the relevant particle and event
- they then compute the wavefunction for that particle/event under the given scenario (mass, helicity, ingoing/outgoing)

What I am proposing/questioning is that it may be better to have the xxx functions do only the calculation, while the AOSOA decoding should be done externally.

In practice
- the xxx interface would take as input the momenta for a given particle, and a given event (or range of events, ie "event page" for vectorization): as arrays of references are forbidden, however, this may require passing four separate fptype_sv (one for energy, px, py, pz)
- in other words, the xxx function swould know nothing of AOSOA structures
- it would be the caller of the xxx functions, presently calculate_wavefunction, that encapsulates the knowledge of the AOSOA structure, possibly even through an extrenal call

One drawback of the present class structure is that there a lot of hidden couplings through hidden assumptions, eg rambo knows about the AOSOA structure too

This new suggested structure may eventually help interfacing c++/cuda to fortran, or maybe not... to be reconsidered later

In any case, one should understand if the performance remains the same, degrades, or even improves through this refactoring. One should also study if the various input fptype_sv should be passed as references or by value. Hint: in this line I had the impression that passing by value was slightly faster, which I found a bit counterintuitive, https://github.com/madgraph5/madgraph4gpu/blob/6f4916ec5552f57d8b58520f924c28e1e495673c/epoch1/cuda/ee_mumu/src/HelAmps_sm.cc#L40

Low priority. Just filing so I do not forget this...

valassi:
I have been continuing to experiment with a prototype for this in PR #189. There I have some code which "looks nice" and is fully functional, but for CUDA is slower than it was. I am using eemumu (in epoch1 because this is where I was). For instance 
[...]

PR #189, opened 2021-04-28
(DO-NOTHING) reverted tests for further separating data access from calculations in xxx

valassi:
Hi @oliviermattelaer @roiser (copying also @hageboeck ) this is a first draft of what we discussed on Monday evening. 

It is a first draft implementation of a patch for issue #175.

I created this simple
```
  struct p4type_sv
  {
    fptype_sv p0;
    fptype_sv p1;
    fptype_sv p2;
    fptype_sv p3;
  };
```
which essentially can be used to be passed around (by value...) various methods.

The idea is that now
- calculate_wavefunction passes allmomenta to the xxx functions
- the xxx functions take allmomenta as input and create wavefunctions
- internally the xxx functions know about the AOSOA structure and call pIpar (once per momentum component - actually some only call the pz part) to do the decoding
Instead eventually
- the xxx functions can be split in two: xxx proper, and a different iPar (the one in this draft)
- p4Ipar takes allmomenta as input and returns a p4type_sv (by value), i.e. all 4 components, in one go
- the new xxx functions do not know anything of allmomenta AOSOA, they just take the p4type_sv (which actually is itself an SOA, at least in SIMD code, but much simpler - in cuda it is a simple 4-vector)
- calculate_wavefunctions calls p4Ipar first, then passes its output to the new xxx

The main advantage is better encapsulation, decoupling the calculation in HelAmps xxx/FFV from data access in CPPProcess. Eventually when we link to Fortran this may well be needed.

Anyway, so far the WIP is just a proof of concept
- I wanted to check first that doing some opies by value (rather than passing references) does not give a hit, and it seems ok
- And I wanted to get some first design there

Then actually this is quit ea bit of (easy but boring/delicate) work. Also the runTests and testxxx need to be changed, and all performannce reevaluated.

Not sure I will work much on this these days, but you have some comments, please let me know :-)
Thanks Andrea


valassi:
PS Just to make it clear, this is a draft: in
```
#ifdef __CUDACC__
      const int ievt = blockDim.x * blockIdx.x + threadIdx.x; // index of event (thread) in grid
      //printf( "ixzxxx: ievt=%d threadId=%d\n", ievt, threadIdx.x );
      const p4type_sv p4vec = p4IparIevt( allmomenta, ipar, ievt );
#else
      //printf( "ixzxxx: ipagV=%d\n", ipagV );
      const p4type_sv p4vec = p4IparIpag( allmomenta, ipar, ipagV );
#endif
      const fptype_sv& pvec0 = p4vec.p0;
      const fptype_sv& pvec1 = p4vec.p1;
      const fptype_sv& pvec2 = p4vec.p2;
      const fptype_sv& pvec3 = p4vec.p3;
```
this is now all inside the xxx function. BUt the idea is that the first lines would move to calculate_wavefunction, and the xxx would get the "p4vec" as an input in the function interface.

valassi:
I have decided to revert all changes I attempted in this WIP PR.

As discuss in #175, we probably need a rather different approach, where CUDA and C++ diverge more
- in cuda we do need large arrays as input to many (all?!) functions, also because we may eventually splity them as separate kernels
- in c++ all xxx/ffv functions should deal with single events, or SIMD-vectirs of events: essentially 1 to 16 events, there is no need to pass arrays with hundreds of thousands of events there

Note also that
- eventually maybe in c++ it could be good to do everything in check.cc and the random and rambo with short arrays of events and a large loop (this may also be useful to reduce memory)
- the multithreading will probably be done elsewhere anyway

I will self-merge this. It is a DO-NOTHING PR, with many commits attempted and reverted, which in any case are only in epoch1/eemumu

PR #197, opened 2021-05-04
A few basic plots on SIMD and multithreading

valassi:


valassi:
self merging, this is a different directory

Issue #198, opened 2021-05-12
epoch2 ggttgg: different physics results for check.exe and gcheck.exe


valassi:
In epoch2 ggttgg: different physics results for check.exe and gcheck.exe

Not a random number issue - momenta are the same, but ME are different

This is visible in the average MEs and also in the individualevents, eg here the last of only 8 events

```
[avalassi@itscrd70 gcc9.2/cvmfs] ~/GPU2020/madgraph4gpuTer/epoch2/cuda/gg_ttgg/SubProcesses/P1_Sigma_sm_gg_ttxgg> ./check.exe -v 1 8 1 | tail -10 
Momenta:
   1  7.500000e+02  0.000000e+00  0.000000e+00  7.500000e+02
   2  7.500000e+02  0.000000e+00  0.000000e+00 -7.500000e+02
   3  3.118787e+02  1.091564e+02 -2.764329e+02  9.454122e+01
   4  2.317597e+02 -6.474121e+01 -1.333573e+02 -1.781487e+02
   5  3.853875e+02  1.954334e+00 -7.753632e+01  3.775021e+02
   6  5.709741e+02 -4.636953e+01  4.873266e+02 -2.938947e+02
--------------------------------------------------------------------------------
 Matrix element = 5.88378e-07 GeV^-4
--------------------------------------------------------------------------------
[avalassi@itscrd70 gcc9.2/cvmfs] ~/GPU2020/madgraph4gpuTer/epoch2/cuda/gg_ttgg/SubProcesses/P1_Sigma_sm_gg_ttxgg> ./gcheck.exe -v 1 8 1 | tail -10 
Momenta:
   1  7.500000e+02  0.000000e+00  0.000000e+00  7.500000e+02
   2  7.500000e+02  0.000000e+00  0.000000e+00 -7.500000e+02
   3  3.118787e+02  1.091564e+02 -2.764329e+02  9.454122e+01
   4  2.317597e+02 -6.474121e+01 -1.333573e+02 -1.781487e+02
   5  3.853875e+02  1.954334e+00 -7.753632e+01  3.775021e+02
   6  5.709741e+02 -4.636953e+01  4.873266e+02 -2.938947e+02
--------------------------------------------------------------------------------
 Matrix element = 2.08873e-06 GeV^-4
--------------------------------------------------------------------------------
```

valassi:
Fixed by PR #198

PR #199, opened 2021-05-12
[ggttgg] two functional bug fixes in the ME calculation and one in the Makefile

valassi:
Two bug fixes in the ME calculation
- first (fix issue #198, cuda and C++ were giving different results, now fixed): avoid using "abs" for floats, this truncates to integers
- second (changes both CUDA and C++ in the same way): the handling of energy and mass in ixxxxx was different from that in the Fortran, I now am using the Fortran version as in epoch1 eemumu... @oliviermattelaer can you please cross check and comment? otherwise I would have to fix it also in epoch1 eemumu (an dthen for upstream posrts)

In addition, a third bug fix in the Makefile, the code must be recuit if HelAmps changes


valassi:
Copying also @roiser and @ingvildh for info - related to the discussion we had yesterday on physics validation for epoch2 ggttgg

Self-mergin now

Issue #25, opened 2020-08-18
Check that we have no issues with branch divergence


valassi:
Just a note as a reminder, followin g up on 'SIMD/SIMT' issues. After investigating SOA/AOS data access and showing we have no uncoalesced amemory access (issue #16) for momenta, I was wondering how to best check in the profiler if we have issues with divergent branches, i.e. threads in our warps which go out of 'lockstep'.

The only reference I found in the profiler doc is here
https://docs.nvidia.com/nsight-compute/ProfilingGuide/index.html#statistical-sampler

![image](https://user-images.githubusercontent.com/3473550/90513556-1e51aa80-e160-11ea-9228-1fba5f2d64e4.png)

If I understand correctly, this means that we should see "Stalled Barrier" in the Warp statistics. This seems to be always at zero.
![image](https://user-images.githubusercontent.com/3473550/90513687-50630c80-e160-11ea-86a2-a6b4c5d2bfc4.png)

I would say that we have no issues with branch divergence. Not surprising really, as all threads are doing exactly the same operations...

valassi:
Note that this is also relevant to vectorisation, #71 and #72.

The fact that we get almost the full factor 4 from AVX2 is a sign that we have no divergence on the CPU.

We should keep this open to reevaluate when we add a selection cut.

valassi:
After a few months I have come back on this issue with two improvements
- one, I think I understand better which are the relevant metrics
- two, i made a small test that introduces artifically some divergence, just to see what this gives in the profiles
I also make some comments on my previous posts

(1) NEW TESTS AND METRICS

The code is in PR #202 and #203 

The main metric is sm__sass_average_branch_targets_threads_uniform.pct
https://github.com/madgraph5/madgraph4gpu/blob/2510b365c1470f56ea784db2726367fd52f045b1/epoch1/cuda/ee_mumu/SubProcesses/P1_Sigma_sm_epem_mupmum/throughput12.sh#L157
This metric should be 100% for uniform execution (i.e. no divergence) and less than 100% for divergence.

Unfortunately, it is difficult to translate a percentage of non-uniformity into a throughput degradation. In the example below, I get a 96% uniformity, but the throughput degradation is around 20-30%, not just 4%!

The test is this
[...]

PR #202, opened 2021-05-12
[warps] measure cuda thread divergence

valassi:
Two changes
- improve the profiling script to also measure cuda thread divergence
- add the option to artificially introduce some thread divergence in eemumu, just to check what it does

This essentially resolves issue #25, using command line profiling rather than GUI profiling


valassi:
Self merging

PR #203, opened 2021-05-13
[warps] add the divergence metrics also to the profile script for GUI analysis

valassi:
… analysis

Also add a samll fix in that script (add an explicit make)


valassi:
minimal addition to the previous PR #202 
self merging

PR #204, opened 2021-05-13
[ggttgg/madeventAV] add a ggttgg fortran gridpack + enable fast math for epoch2 ggttgg

valassi:
This addresses yesterday's discussion with @jtchilders @roiser @oliviermattelaer 

Note that the Fortran throughout is a factor 2 higher than C++, to be understood

Without fast math, fortran:
    TOTAL MATRIX1 :    2.0968s for     6511 calls => throughput is 3.11E+03 calls/s

With fast math, fortran
    TOTAL MATRIX1 :    3.0558s for    12090 calls => throughput is 3.96E+03 calls/s

Without fast math, epoch2 c++
    EvtsPerSec[MatrixElems] (3)= ( 1.345116e+03                 )  sec^-1

With fast math, epoch2 c++
    EvtsPerSec[MatrixElems] (3)= ( 1.906572e+03                 )  sec^-1

These numbers are **VERY PRELIMINARY**

Clearly some things to be understood in the c++. Note that in eemumu the vectorization work had improved throughput even without switching on SIMD. The above numbers are without SIMD.

If one imagines a factor almost 4 from SIMD in C++, this would still be almost a factor 2 faster than fortran... but probably we can do better.


valassi:
self merging - NB THIS ENABLES FAST MATH IN EPOCH2

Issue #22, opened 2020-08-13
Define improved performance metrics


valassi:
As our prooof of concept prototype advances (see issue #19), we should make sure we define better metrics for performance, to compare GPU implementations to one another and also to CPU implementations in C++ and Fortran.

This is a typical printout for our baseline now (see https://github.com/roiser/madgraph4gpu/commit/812f7438462c01e1052c4be48cafcdaef104195c):
```
time ./gcheck.exe -p 16384 32 12
***************************************
NumIterations             = 12
NumThreadsPerBlock        = 32
NumBlocksPerGrid          = 16384
---------------------------------------
FP precision              = DOUBLE (nan=0)
Complex type              = THRUST::COMPLEX
Momenta memory layout     = AOSOA[32]
Wavefunction GPU memory   = LOCAL
Curand generation         = DEVICE (CUDA code)
---------------------------------------
NumberOfEntries           = 12
TotalTimeInWaveFuncs      = 9.991298e-03 sec
MeanTimeInWaveFuncs       = 8.326082e-04 sec
StdDevTimeInWaveFuncs     = 1.905112e-05 sec
MinTimeInWaveFuncs        = 8.232730e-04 sec
MaxTimeInWaveFuncs        = 8.263180e-04 sec
---------------------------------------
NumMatrixElementsComputed = 6291456
MatrixElementsPerSec      = 6.296935e+08 sec^-1
***************************************
NumMatrixElements(notNan) = 6291456
MeanMatrixElemValue       = 1.394735e-02 GeV^0
StdErrMatrixElemValue     = 3.034488e-06 GeV^0
StdDevMatrixElemValue     = 7.611337e-03 GeV^0
MinMatrixElemValue        = 6.071582e-03 GeV^0
MaxMatrixElemValue        = 3.374925e-02 GeV^0
***************************************
00 CudaFree : 0.142011 sec
0a ProcInit : 0.002600 sec
0b MemAlloc : 0.075228 sec
0c GenCreat : 0.014557 sec
1a GenSeed  : 0.000014 sec
1b GenRnGen : 0.007914 sec
2a RamboIni : 0.000115 sec
2b RamboFin : 0.000058 sec
2c CpDTHwgt : 0.006667 sec
2d CpDTHmom : 0.071620 sec
3a SigmaKin : 0.000090 sec
3b CpDTHmes : 0.009901 sec
4a DumpLoop : 0.022485 sec
9a DumpAll  : 0.046493 sec
[...]

PR #206, opened 2021-05-28
Add MECalc-only throughput (no DtoH copy)

valassi:
Add MECalc-only throughput (no DtoH copy) as discussed in issue #22 

Another minor change, increase ntpbMAX to 1024 (was 256, but 512 works without issues on a V100)


valassi:
self merging

Issue #176, opened 2021-04-25
Decouple neppM and neppV in vectorised C++ code?


valassi:
This is a spinoff of vectorisation issue #71 and a followup to the big PR #171.

It is also somewhat related to issue #175 about the xxx function interface.

Presently, there are are at least three types of (AO)SOA structures in the code, determined by three parameters

- neppR determines the number of events per page in the AOSOA random number array. This is hardcoded for all (cuda, cpp, double, float, all SIMD, no SIMD...) configurations, so that the same physics results are obtained. It is essentially an internal parameter of the phase space sampler (rambo), while it does not even need to be known to the random engine itself. The random engine creates a flat array of random numbers, then rambo assignes these to different events according to neppR.

- neppM is the number of events per page in the AOSOA momenta data structure. It is known to the xxx function interface (as allmomenta, see issue #175) and it is known to rambo (the getMomentaInitial and getMomentaFinal functions). This is about how the encoding/decoding of momenta in memory is done, in both c++ and cuda. For cuda, this is tuned to a reasonable value to have coalesced memory access, fitting in 32-byte cache lines (see the studies in issue #16). For c++, this is presently coupled to neppV for vectorisation, see next bullet.

- neppV is the size of the vectors used in SIMD vectorized operations. This is dictated by which type of vectorization one wants to have (none:1, SSE42 ie 128bit: 2 doubles, AVX2 ie 256 bit: 4 doubles, 512y ie AVX512 with 256 bit: 4 doubles, 512z ie AVX512 with 512 bit: 8 doubles).

Presently, the implementation of vectorization couples the AOSOA momenta structure and the wavefunction vectorization by assuming neppM == neppV. This is worse than hardcoded, it is just an assumption throughout the code.

However, there is in principle no strict need to have neppM == neppV:
- one could have an AOSOA where each momenta event page fits 2 or 4 or more SIMD vectors, i.e. neppM=2neppV or neppM=4neppV and so on
- and one could even test the performance of the code without AOSOA and with an AOS instead, i.e. with neppM=1

About the latter point, the critical issue is that there are two separate contributions to the speedup from the vectorized implementation, one from data access and one from calculations
- the SIMD calculation is most likely that is most important: when doing a sum for 4 events, an AVX2 implementation does a single processor instruction instead of 4
[...]

PR #208, opened 2021-05-31
Remove neppV and fptype_sv from API and decouple neppV from neppM (epoch1)

valassi:
This is my WIP on decoupling neppV and neppM

I understood various things so far
- in eemumu there can be a serious performance penalty from memory copies if vectorisation is not used (even if data are contiguous!!!)
- to recover performance I used a reinterpret_cast of double* arrays as double_v arrays: however, this segfaults if the original double* array is not aligned
- therefore I have now implemented alignment and reinterpret_cast, performance is there again
- but I have not yet decoupled neppV and neppR

This is preliminary work for the integration of Fortran and C++, as discussed a few days ago with @roiser and @oliviermattelaer


valassi:
This is the preparatory work done long ago (and some more recently) in epoch1 for PR #300 in epochX.

I am self merging this and will then self merge PR #300 too.

valassi:
This completed the expolartory work for epoch1 in issue #176

PR #209, opened 2021-06-11
Improve throughput script + Results of further AOSOA tests

valassi:
A few additional tests while preparing the vchep paper


valassi:
Will self merge

Issue #16, opened 2020-08-12
AOS/SOA for input particle 4-momenta (and random numbers)


roiser:
Done, including AOSOA (AV)

eemumu_AV/master

Steered by typedefs. We need to go to the end with this idea. 

Should go back and time AOS… actually AOSOA~5.0E8, AOS~4.8E8

valassi:
Actually I have timed this yesterday, and AOSOA does seem to pay off, even if not by much.
See https://docs.google.com/document/d/1g2xwJ2FsSlxHvSUdPZjCyFW7zhsblMQ4g8UHlrkWyVw/edit#

Description | CUDA tput
-- | --
BASELINE | 5.00E8/s
SOA | 4.65E8/s speed -7%
AOS | 4.85E8/s speed -3%

I would say that we should keep some AOSOA-like structure as baseline

To do in any case
- understand better what causes the performance improvement (memory access via coalescing? SIMD vectorized computations??)
- massive clean up of the code, should use the same array dimensions everywehere: goal should be AOSOA[ngpublocks][nparticles][np4=4(E,px,py,pz)][ngputhreadsinblock]... also the random number arrays should use ngputhreads as last dimnesions (now it is hardoded as 32, the number if threads in a warp)

valassi:
Some food for thought from Vincenzo, found by chance from google "godbolt cuda soa"
https://indico.cern.ch/event/851670/contributions/3585184/

valassi:
And another interesting talk at CERN from the same google for godbolt
https://indico.cern.ch/event/932905/contributions/3920347/

valassi:
I would like to do some changes to the AOSOA, but first I's like to understand a bit better if/why the layour makes a difference. I can think of two things, memory coalescing, and instruction vectorization. So I am doing some research to understand which metrics are relevant, eg in the profiler.

About memory coalescing, google "nvidia nsight coalesce" brought me here: https://developer.nvidia.com/blog/using-nsight-compute-to-inspect-your-kernels. This is avery interesting article pointing to two metrics, and showing how to focus on them in the tools. I added them here https://github.com/madgraph5/madgraph4gpu/commit/0429183b55bd56a5a85659258d8498e6396b475e

I have then profiled the BASELINE ASA against AOS and SOA. At face value I confiorm that SOA is similar, just ~2% slower than ASA, while AOS is quite a bit slower, around 7-10%. The profiles are very interesting.

[...]

PR #210, opened 2021-06-11
Log a few tests of sector/requests in single precision

valassi:
A few more tests for issue #16


valassi:
There is no code change in this PR, only a few logs, eventually moved back to defaults

Will self merge

Issue #167, opened 2021-04-21
Allow float/double choice from 'make' (with a view on later adding float/double tests in the CI)


valassi:
This is a spinoff of issue #148 which was originally about
- fix the single precision tests in the code
- fix the single precision tests in the CI

I am about to merge a PR for the first part of the issue.

The second part remains open. It involves
- modifying headers and makefiles so that single precision builds are triggered from a -D option to make
- building an dtesting the code BOTH for float and double in the CI

PS (June 2021) This was initially "Add single precision builds and tests in the CI", it is now renamed to only include the "first part of the second part" above. The "second part of the second part" is now in #214.

valassi:
Hi @hageboeck (and @roiser @oliviermattelaer ) I am having a look at this.

If I understand correctly, your idea was to have `make MGONGPU_CONFIG=-DMGONGPU_COMMONRAND_ONHOST` or `export MGONGPU_CONFIG=-DMGONGPU_COMMONRAND_ONHOST; make`? This is what you mentioned in the logs and in PRs #38, #59, #63.

I would like to modify a bit this logic, dropping MGONGPU_CONFIG and using individual variables. In the example above, I would suggest to modify Makefiles to allow 'make MGONGPU_COMMONRAND_ONHOST=1` or `export MGONGPU_COMMONRAND_ONHOST=1; make`.

The reason for this is that I'd like to have the float/double choice available as a separate variable, like `make MGONGPU_FPTYPE_FLOAT=1` or `export MGONGPU_FPTYPE_FLOAT=1; make`. This is what I am using for variable `AVX`.

The first main idea is that I'd like to have separate SIMULTANEOUS builds in different build directories. To achieve this (like I am doing for AVX), I need to know the choice within the Makefile, as it is the MAkefile that creates the build directories. Of course I could decrypt MGONGPU_CONFIG, but it becomes cumbersome...

The second related idea is that, in the option where no simultaneous build directoes ar eallowed, and only one is allowed, then I want to log what float/double was used, and fail the build if you start mixing different options. This is what I already do for AVX (eg make AVX=512z will fail if previously you did make AVX=avx2, without a clean in between). Same thing, I need to know the choice and it's better to get that straight than decrypt it from MGONGPU_CONFIG.

Third reason, I'd prefer avoiding the injection of arbitrary stuff into the compiler options. Presently in MGONGPUCONFIG you can inject anything and this goes straight to g++.

[...]

PR #211, opened 2021-06-14
[ggttgg] fix float support in ggttgg (+ fix sqrt/sqrtf/std::sqrt in cuda/c++ also in eemumu)

valassi:
Hi @roiser @oliviermattelaer @hageboeck @cvuosalo this is a comprehensive PR to fix float support in ggttgg, using the same code and techniques as in eemumu (which may be debatable, but at least are consistent).

Do you have any comments? Thanks
Andrea

PS Strangely, I do not get the same exact results in cuda and c++, unlike eemumu. Maybe this is a real numerical instability? To be checked, but the basics look ok.


epoch2/cuda/gg_ttgg/src/HelAmps_sm.cu
```diff
@@ -822,24 +1614,24 @@ __device__ void FFV1_1(const cxtype F2[], const cxtype V3[], const cxtype COUP,
       P1[3]) - M1 * (M1 - cI * W1));
   F1[2] = denom * cI * (F2[2] * (P1[0] * (-V3[2] + V3[5]) + (P1[1] * (V3[3] -
       cI * (V3[4])) + (P1[2] * (+cI * (V3[3]) + V3[4]) + P1[3] * (-V3[2] +
-      V3[5])))) + (F2[3] * (P1[0] * (V3[3] + cI * (V3[4])) + (P1[1] * (-1.) *
-      (V3[2] + V3[5]) + (P1[2] * (-1.) * (+cI * (V3[2] + V3[5])) + P1[3] *
+      V3[5])))) + (F2[3] * (P1[0] * (V3[3] + cI * (V3[4])) + (P1[1] * (-one) *
```
hageboeck, line 1617: This could also be done as `((fptype)-1.)` if desired.

hageboeck (approved):
It's too much too read in detail, but I have an idea for testing it:
We could add a build to the CI that compiles with `-DMGONGPU_FPTYPE_FLOAT`. This should prevent future compile problems. I suspect that tests would fail, but at least we could add compilation without testing in `float` mode as a first step.

epoch2/cuda/gg_ttgg/src/HelAmps_sm.cu
valassi: > This could also be done as `((fptype)-1.)` if desired.
Yes it is also one I had considered. I think that in the end I had so many 1., 2. and 1/2 all over the place that this seemed a more readable form

valassi:
> It's too much too read in detail, but I have an idea for testing it:
> We could add a build to the CI that compiles with `-DMGONGPU_FPTYPE_FLOAT`. This should prevent future compile problems. I suspect that tests would fail, but at least we could add compilation without testing in `float` mode as a first step.

Adding CI tests for floats would certainly be useful at some point. I would start with eemumu however, because there were many issues in the googletest infrasttructure for floats, which I fixed in the meantime (for manual tests but not the CI).

The oonly issue is that we should clean up those #defines and just allow a definition from outside. But then some guards must be added all over the place to make sure one of the options is added, and not more. I have it in my mental to do for later... since som etime.

Otherwise, about the tests of ggttgg, one would need to include the changes of eemumu, similarly tto what  have done. I think that most of them are here
https://github.com/madgraph5/madgraph4gpu/pull/168
plus later additions in upstream/master
I actually have the impression that ggttgg tests are disabled (runTest is not built), for similar issues to what I had in eemumu. Probably now it should be easier to port that, using the changes from eemumu

valassi:
Ah, actually I see thta https://github.com/madgraph5/madgraph4gpu/pull/148 was exactly abut these two issues for eemumu
- one, fix the google tests (done for eemumu)
- two, massage makefiules, allow float/double from outside, and pass  to the CI... this second part has been stripped off to #167

valassi:
Ok I have done a couple changes following some hintgs from @hageboeck (thanks!)
- replace sqrt and sqrtf by std::sqrt in c++
- conversely, add the distinction between sqrt and sqrtf in CUDA: previously sqrt was used also for floats... th echange does not seem to make any change in results or performance, in any case

I wil merge

PR #213, opened 2021-06-16
[fpsingle] allow float/double choice in the Makefiles + allow simultaneous float/double builds

valassi:
Hi @hageboeck @roiser @oliviermattelaer this is the PR for the first part of #167 

Essentially:
```
make FPTYPE=f
```
will build with floats

Also
```
make USEBUILDDIR=1 FPTYPE=f
make USEBUILDDIR=1 FPTYPE=d
```
will allow simultaneous builds (as done for AVX now)

About common random:
```
make RNDGEN=common
```

The choice of env variables and their possible values is quite arbitrary... maybe ugly and not precise enough but short.

Let me know if you have any comments...

Thanks
Andrea


hageboeck:
LGTM. Did I see correctly that `bash throughput12.sh -h` will print help on arguments?

valassi:
Thanks! 

About throughput12.sh. Well that scripts is a bit my own thingy, so the documentation is MINIMAL... it just says whic options exist. Hopefully they are self explanatory. But you are more than wecome to try it out and give feedback :-)

Just one or two words of caution on that
- I created it for epoch1/epoch2 eemumu comparisons, but now this is obsolete... I should drop the epoch2 eemumu comparisons and focus on eemumu ep1 (which is what it mainly does)
- There is some minimal ggttgg testing, but this whould probably go to its own script in ggtgg
- In other words: if you only play with the other options, only for eemumu ep1, you should be ok

C++ example: `/throughput12.sh -avxall -flt` is probably the most comprehensive you want for c++
You can throw it -omp if you want the throughputs with OMP multithreading
You can throw in -detailed if you want more info from perf

CUDA example: `./throughput12.sh -div -req -nocpp` gives you extra ncu profiling for divergence and for sectors/requests (and then you can diable c++ with -nocpp)

Let me know if you have questions./.. a

valassi:
Ok I merge this then, thanks again Stephan.
Andrea

PR #216, opened 2021-06-25
FIXES for clang: dump gcc toolchain and ensure nvcc actually uses it

valassi:
This started off as a followup of issues reported by @hageboeck on itscrd04
- After an upgrade of clang from 10 to 11 by puppet, things no longer worked: he fixed it by upgrading cuda from 11.1 to 11.3 on his machine: he fixed i
- On my side however, using cvmfs clang, clang11.1 did not work because it was setup with gcc 10.3: I fixed it by downgrading the setup to gcc 10.2 (thanks to Dima, https://sft.its.cern.ch/jira/browse/SPI-1924)

As a followup:
- I understood there is a toolchain in clang and found a way to dump it (actually I am dumping the gcc used to build libgcc.a)
- As I was there, I also dumped the gcc/clang version used by nvcc
- I realised that nvcc with clang was NOT using clang, because "-ccbin" was missing, so I added that
- After adding that, nvcc builds did change and in particulary one failed, I added a simple fix for bracket nesting level

Now everything looks fine. Apart from the dumps, the main functional change is that nvcc builds with clang actually do use clang as host compiler

### DIFF
--- a/epoch1/cuda/ee_mumu/SubProcesses/Makefile
+++ b/epoch1/cuda/ee_mumu/SubProcesses/Makefile
@@ -74,7 +74,7 @@ else
 endif
 
 # Set the host compiler for nvcc
-override NVCC := $(NVCC) -ccbin $(subst ccache ,,$(CXX))
+CUFLAGS += -ccbin $(subst ccache ,,$(CXX))
 
 # Set the build flags appropriate to each AVX choice (example: "make AVX=none")
 # [NB MGONGPU_PVW512 is needed because "-mprefer-vector-width=256" is not exposed in a macro]
--- a/epoch1/cuda/ee_mumu/SubProcesses/Makefile
+++ b/epoch1/cuda/ee_mumu/SubProcesses/Makefile
@@ -148,9 +148,9 @@ $(info Building in BUILDDIR=$(BUILDDIR) for tag=$(TAG))
 ifeq ($(USECCACHE)$(shell echo $(CXX) | grep ccache),1)
   override CXX:=ccache $(CXX)
 endif
-ifeq ($(USECCACHE)$(shell echo $(AR) | grep ccache),1)
-  override AR:=ccache $(AR)
-endif
+#ifeq ($(USECCACHE)$(shell echo $(AR) | grep ccache),1)
+#  override AR:=ccache $(AR)
+#endif
 ifneq ($(NVCC),)
   ifeq ($(USECCACHE)$(shell echo $(NVCC) | grep ccache),1)
     override NVCC:=ccache $(NVCC)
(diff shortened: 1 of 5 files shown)


valassi:
Hi @hageboeck can you please check this? For me it works in manual builds, but it does not within the CI.

The makefiles are getting a bit convoluted (I add ccache and remove it in one place)... but ok, seemed ok for me.

Maybe you have some special setup in the CI? 

I am off this afternoon otherwise I would check... thanks
Andrea

hageboeck:
> Hi @hageboeck can you please check this? For me it works in manual builds, but it does not within the CI.
> 
> The makefiles are getting a bit convoluted (I add ccache and remove it in one place)... but ok, seemed ok for me.
> 
> Maybe you have some special setup in the CI?
It's a more or less standard centos8 container with cuda (provided by nvidia).

In the failed CI job, it tries to invoke the executable `ccbin`,  so there's presumably an error quoting a variable or a variable is empty. I'm at least not aware of an executable that's called `ccbin`.

hageboeck:
> Maybe you have some special setup in the CI?
I just see that it's github's CI job that failed (not the GPU job that runs at CERN). This is a standard github container that runs `ubuntu-latest`. It must still be a thing where a variable happens to be empty or similar, though.

epoch1/cuda/ee_mumu/SubProcesses/Makefile
```diff
@@ -73,6 +73,9 @@ else
   override RNDGEN = common
 endif
 
+# Set the host compiler for nvcc
+override NVCC := $(NVCC) -ccbin $(subst ccache ,,$(CXX))
+
```
hageboeck, line 78: I think I understand now:
This force-overrides the `NVCC` variable. In the C++-only CI builds, it's empty, but after this line, it's not.

The subsequent parts test this if this variable is empty, however. If empty, no cuda is built. If set, it tries to build CUDA, but it expands to `"" -ccbin <expansion of CXX>`.

epoch1/cuda/ee_mumu/src/Makefile
```diff
@@ -22,6 +22,9 @@ ifneq ($(wildcard $(CUDA_HOME)/bin/nvcc),)
   CUINC = -I$(CUDA_HOME)/include/
 endif
 
+# Set the host compiler for nvcc
+override NVCC := $(NVCC) -ccbin $(subst ccache ,,$(CXX))
```
hageboeck, line 26: And the same here. I don't know which of the two triggered the error, but it should be the same solution.

valassi:
Hi Stephan, thanks a lot.
Indeed one of the main issues is that I had not tried out a C++-only build.
If I do this (export CUDA_HOME=whatever) then I get some failures like
```
ccache  -ccbin /cvmfs/sft.cern.ch/lcg/releases/clang/11.1.0-b24ba/x86_64-centos7/bin/clang++ -o ./runTest_cu.o -c -x cu runTest.cc  -DMGONGPU_FPTYPE_DOUBLE -DMGONGPU_COMMONRAND_ONHOST
Cleaned cache
Cleaned cache
ccache: invalid option -- 'b'
```
I will work on a fix.

Issue #223, opened 2021-06-29
Build and test on Marconi M100 (Power9 + A100)


valassi:
As discussed at the June 21 meeting https://indico.cern.ch/event/1028452/

The M100 system documentation is here: https://wiki.u-gov.it/confluence/display/SCAIUS/UG3.2%3A+MARCONI100+UserGuide

valassi:
First patch for Marconi M100 (Power9+A100) is PR #224
- Disable vectorization in C++, need to understand better the hardware and the compiler
- Port throughput12.sh script (different /proc/cpuinfo, different nvidia-smi, different ncu...)

valassi:
About Power vector intrinsics: http://cdn.openpowerfoundation.org/wp-content/uploads/resources/Intrinsics-Reference_final/Intrinsics-Reference-20200811.pdf

valassi:
I have committed a basic patch for vectorization on Power9/M100 in PR #228 
- essentially I just added -D__SSE4_2__
- I also added -mcpu=power9 -mtune=power9 which seem to gain a few 2-3%
- the option -flto can give HUGE improvements (a factor 4 without vectorization??) but needs further analysis, I will open an issue for that
- all other build options do not really seem to give much
- I also made some basic fixes of the objdump parser (but it does not work with -flto)

By the way, the CUDA build options probably needs to be reviewed too. I am building with gcc as host compiler now, not sure what "-Xcompiler" does.

The following is the current baseline (for C++) from https://github.com/madgraph5/madgraph4gpu/commit/ad4f39f5c5bf7f7968c4f36f92c797364d3d73cf
```
On login01 [CPU: PowerNV 8335-GTG, POWER9, altivec supported] [GPU: 4x Tesla V100-SXM2-16GB]:
=========================================================================
Process                     = EPOCH1_EEMUMU_CPP [gcc 8.3.1]
FP precision                = DOUBLE (NaN/abnormal=0, zero=0)
Internal loops fptype_sv    = SCALAR ('none': ~vector[1], no SIMD)
Random number generation    = COMMON RANDOM (C++ code)
OMP threads / `nproc --all` = 1 / 128
EvtsPerSec[MECalcOnly] (3a) = ( 1.419728e+06                 )  sec^-1
MeanMatrixElemValue         = ( 1.372113e-02 +- 3.270608e-06 )  GeV^0
TOTAL       :     5.771237 sec
[...]

PR #224, opened 2021-06-30
First patch for Marconi M100 (Power9+A100)

valassi:
First patch for Marconi M100 (Power9+A100) issue #223 
- Disable vectorization in C++, need to understand better the hardware and the compiler
- Port throughput12.sh script (different /proc/cpuinfo, different nvidia-smi, different ncu...)


valassi:
All successful now, self merging

valassi:
Renamed this PR: Marconi has V100s, not A100s

Issue #220, opened 2021-06-29
Build and test using the Intel compiler


valassi:
This was discussed during the June 28 meeting https://indico.cern.ch/event/1053713/

Would this help with the AVX512 issue #173 ?

valassi:
Note that icc 19.1 should be supported already with cuda 11.0
https://docs.nvidia.com/cuda/archive/11.0/cuda-installation-guide-linux/index.html

valassi:
Vector extensions give some issues on icc. This needs some debugging. The error from the compiler makes no sense, either those are both double or they are both fptype_v...?
```
ccache icpc  -O3 -std=c++17 -I.  -Wall -Wshadow -Wextra -fopenmp  -ffast-math  -march=nehalem  -DMGONGPU_FPTYPE_DOUBLE -DMGONGPU_COMMONRAND_ONHOST  -c rambo.cc -o rambo.o
In file included from rambo.h(6),
                 from rambo.cc(1):
mgOnGpuVectors.h(63): error: a value of type "mgOnGpu::fptype_v" cannot be used to initialize an entity of type "double"
      cxtype_v( const fptype_v& r, const fptype_v& i ) : m_real{r}, m_imag{i} {}
                                                                ^
```
Note that in principle simple things should work
https://stackoverflow.com/a/43801280

valassi:
Brief status:
- with nvcc, things look very messy (even using -ccbin, it seems that gcc is used?)
- setting an invalid CUDA_HOME one can build only the C++: the vector builds give the issues above, the scalar builds look ok

```
On itscrd70.cern.ch [CPU: Intel(R) Xeon(R) Silver 4216 CPU] [GPU: NVIDIA Tesla V100S-PCIE-32GB]:
=========================================================================
Process                     = EPOCH1_EEMUMU_CPP [icc 1910]
FP precision                = DOUBLE (NaN/abnormal=0, zero=0)
Internal loops fptype_sv    = SCALAR ('none': ~vector[1], no SIMD)
Random number generation    = COMMON RANDOM (C++ code)
OMP threads / `nproc --all` = 1 / 4
EvtsPerSec[MECalcOnly] (3a) = ( 1.209906e+06                 )  sec^-1
MeanMatrixElemValue         = ( 1.372113e-02 +- 3.270608e-06 )  GeV^0
TOTAL       :     6.724400 sec
    19,820,008,713      cycles                    #    2.669 GHz                    
[...]

PR #225, opened 2021-06-30
Intel compiler icc (and icx!) patches

valassi:
WIP - Intel compiler icc patches

This is a running patch for issue #220


valassi:
Note that up to here the vector builds are still not building - this is why this is WIP

valassi:
This is now a good patch for issue #220 because I moved from icc to icx

valassi:
all ok, self merging

PR #228, opened 2021-07-02
Basic vectorization for Power9 on Marconi M100

valassi:


valassi:
This is the second patch for issue #223

I will self merge

Issue #229, opened 2021-07-02
Investigate link time optimizations (-flto)


valassi:
This is a spinoff of the Power9 issue #223.

I realised that adding -flto there (gcc link time optimizations) gains almost a factor 4 for scalar C++ code (and more than a factor 2 for simd code). I am using gcc8 there. Compare:
- no lto: https://github.com/madgraph5/madgraph4gpu/commit/ad4f39f5c5bf7f7968c4f36f92c797364d3d73cf
- lto: https://github.com/madgraph5/madgraph4gpu/commit/86b7c15227444e78bd177e895742e685f22ddc8d

Amongst the things to be understood:
- what does lto really involve in our case
- does it also provide a benefit for x86 (or is it only on power9)
- disadvantages? keep in mind the LHC experiments have their own linking ways, eg fpic etc
- try out a newer gcc compiler (eg is the lto-dump tool only available with gcc10?)
- how to interpret objdump when -flto is switcehed on

valassi:
In PR #230 I checked -flto also on Intel processors: again I get large throughput increases

Compare
- no lto https://github.com/madgraph5/madgraph4gpu/commit/370ce41325dc13d164a99d6d71bf6c76edb50f12
- lto https://github.com/madgraph5/madgraph4gpu/commit/9293ea88685c28622ca2d81a2387259d228f3231

Double:
```
No LTO
FP precision                = DOUBLE (NaN/abnormal=0, zero=0)
Internal loops fptype_sv    = SCALAR ('none': ~vector[1], no SIMD)
  EvtsPerSec[MECalcOnly] (3a) = ( 1.315891e+06                 )  sec^-1
Internal loops fptype_sv    = VECTOR[4] ('512y': AVX512, 256bit) [cxtype_ref=YES]
  EvtsPerSec[MECalcOnly] (3a) = ( 4.960773e+06                 )  sec^-1

LTO
FP precision                = DOUBLE (NaN/abnormal=0, zero=0)
Internal loops fptype_sv    = SCALAR ('none': ~vector[1], no SIMD)
  EvtsPerSec[MECalcOnly] (3a) = ( 4.377819e+06                 )  sec^-1
Internal loops fptype_sv    = VECTOR[4] ('512y': AVX512, 256bit) [cxtype_ref=YES]
  EvtsPerSec[MECalcOnly] (3a) = ( 1.070241e+07                 )  sec^-1
```

Float:
```
No LTO 
[...]

PR #230, opened 2021-07-04
Test -flto also on Intel (spinoff of M100 tests on Power9)

valassi:


valassi:
Just add the -flto option also on Intel (commented out) as per issue #229

Observe big throughput increases (factor 2-4) also on Intel processors

valassi:
self merging

PR #231, opened 2021-07-05
Various LTO tests (emulate LTO by inline, use -flto only on one file, etc)

valassi:
This is another patch for issue #229. The idea is to recover some of the LTO benefits with code changes such as inlining, rather than with the -flto build flag.


valassi:
Self merging - the baseline is not changed, I only added a few options to explore (disabled by default)

PR #232, opened 2021-07-09
Upgrade googletest from 1.10.0 to 1.11.0

valassi:
Upgrade googletest from 1.10.0 to 1.11.0

This is needed to port to gcc11.1 and avoid [-Werror=maybe-uninitialized]

A small fix in a test .cc is also required (fixed using the advice by googletest)


valassi:
Copying @hageboeck for info

Self-merging (not controversial I think)

PR #233, opened 2021-07-09
Adding always_inline recovers all advantages of LTO (within 3-5%)

valassi:
Next patch for issue #229 

Adding always_inline recovers all advantages of LTO (within 3-5%)

Keep this disabled by default for the moment


valassi:
self-merging

Issue #244, opened 2021-09-16
epochX - code generation, diffs, backports of latest cuda/c++


valassi:
I put down a few ideas following this afternoon's code generation workshop https://indico.cern.ch/event/1061524/ about how I suggested I would structure the repo (and also specific steps I plan to implement for cuda/c++)

What I would create is a new epochX directory, next to epoch1 and epoch2.

The plan is to backport my vectorization changes (now the latest are in epoch1/cuda/ee_mumu) to epochX/cuda.

This epochX would be an evolving directory (i.e. not a semi-fixed one like epoch1/2), containing both manually developed code and code-generating code infrastructure.

In a "steady state", when all new manual developments have been backported to the code generating code, one should be able to run the code generating code and obtain EXACTLY the same code as that which has been manually developed and committed to the repo. This requires a functioning code generating infrastructure (eg with only a few plugin files in github, but relying on an "external" installation of Mg5aMC somewhere configurable), and some relocation and diff scripts: essentially yo generate new code in a tmp area, then you compare it to what is committed in the repo. It may also require some beautifier to bypass trivial formatting differences (to be seen).

Beyond the steady state, one develops manual code and commits that to the github repo. At this stage, you have in github a code generating plugin that is "older" than the manual developments, and that is perfectly ok. At some point, you do the effort of backporting the new manual changes into the code generating plugin, and ensure that you reach the steday state again. Then you can iterate and do new manual developments.

Specifically for the "cuda" directory (native cuda, plus vectorized c++, with ifdefs for the moment), I would imagine working on the following steps in sequence:

1. Create the basic infrastructure, with "manual" code and code generating code. Ensure that you get the external MG5aMC and diff scripts working. In practice, start with Olivier's current bazaar code generatoing code, add that as a plugin to github. Generate processes (eemumu and ggttgg? probably both) with that code generating code, commit that to github as the first "manual" version. This would be clearly different from the latest epoch1/eemumu (missing all vectorization), but also from epoch2/ggttgg (I checked today, there are small differences). This is a steady state. Make a PR for this.

2. Replace epochX/ggttgg by the manual code that is currently in epoch2/ggttgg. This is beyond steady state, Start doing actual work on the code generating plugin, ensure that eventually the plugin is able to generate exactly the same ggttgg code. Make sure that the new plugin also works fine on eemumu: generate a new eemumu version and just check it looks ok, commit that to github. This is again a steady state, for both ggttgg and eemumu. Make a second PR.

[...]

PR #245, opened 2021-09-20
WIP: EpochX step 1 - basic infrastructure

valassi:
This is step1 in the plan outlined in issue #244.

It just includes a script to automate the code generation and the comparison to a previous version. I used Olivier's tutorial plugin as-is (with some renaming) and I generated eemumu.

The problem is that _code generation is not stable_. Repeating the same identical code generation twice produces code which is functionally identical, but is NOT line-by-line identical. The order in which some functions are generated (or even the order in which some lines within the same functions are generated) is not strictly reproducible and may vary from one run to another. @oliviermattelaer do you have any feedback on this? I will investigate anyway.

Example,
```
...
Files ee_mumu.NEW/src/mgOnGpuConfig.h and ee_mumu/src/mgOnGpuConfig.h are identical
Files ee_mumu.NEW/src/mgOnGpuTypes.h and ee_mumu/src/mgOnGpuTypes.h are identical
diff -rs ee_mumu.NEW/src/Parameters_sm.cc ee_mumu/src/Parameters_sm.cc
46d45
<   mdl_CKM3x3 = 1.; 
48a48
>   mdl_CKM3x3 = 1.; 
132,133d131
<   cout << setw(20) <<  "mdl_CKM3x3 " <<  "= " << setiosflags(ios::scientific)
<       << setw(10) << mdl_CKM3x3 << endl;
137a136,137
>   cout << setw(20) <<  "mdl_CKM3x3 " <<  "= " << setiosflags(ios::scientific)
>       << setw(10) << mdl_CKM3x3 << endl;
diff -rs ee_mumu.NEW/src/Parameters_sm.h ee_mumu/src/Parameters_sm.h
26,31c26,32
<         mdl_Gf, aEWM1, mdl_MH, mdl_MZ, mdl_MTA, mdl_MT, mdl_MB, mdl_CKM3x3,
<         mdl_conjg__CKM1x1, mdl_conjg__CKM3x3, mdl_MZ__exp__2, mdl_MZ__exp__4,
<         mdl_sqrt__2, mdl_MH__exp__2, mdl_aEW, mdl_MW, mdl_sqrt__aEW, mdl_ee,
<         mdl_MW__exp__2, mdl_sw2, mdl_cw, mdl_sqrt__sw2, mdl_sw, mdl_g1, mdl_gw,
<         mdl_vev, mdl_vev__exp__2, mdl_lam, mdl_yb, mdl_yt, mdl_ytau, mdl_muH,
[...]

PR #247, opened 2021-09-27
WIP: EpochX step 2 - backport epoch2 ggttgg

valassi:
This is step2 in the plan outlined in issue #244.

It includes wrk in progress (not yet completed) for backporting the code generation of the current epoch2 ggttgg.
- NB: epoch2 ggttgg is more advanced than the code generation in upstream MG5aMC (hence the need for this backport)
- NB: epoch2 ggttgg is less advanced than the code generation in eemumu epoch1, which includes vectorization (hence this step2 comes before the backport of eemumu epoch1 and vectorization)

I fixed quite a few issues and have now most files from templates inside the plugin. Will continue work on it tomorrow. 

One issue I am having now is that there is a code indenter which I would like to disable (another option is to use the clang-format beautifier, but if I can avoid all beautification for the moment I find it easier to work with).

Note that the overall plan in #244 has slightly changed. I am now committing BOTH the auto generated code (.auto dir) and the manual code one is currently working on. I will document that there


valassi:
Ok the indentation comes from "class CPPWriter(FileWriter)" in madgraph.iolibs.file_writers. This is called from write_aloha_routines in UFOModelConverterGPU. To disable the indentation it is enough to use FileWriter instead of CppWriter. This works better with large copied code (eg ixxx) but works porly on the FFV routines. I will try to work like this anyway, and fix the FFV a posteriori. Otherwise getting diffs of the code becomes a nightmare.

valassi:
I have completed the backport of src (HelAmps) from ggttgg epoch2.
There was a lot of work on formatting, to make sure the final code looks nice (without CppWriter).
I also introduced the constexpr 'one' for '1.' as in ggttgg (useful later for vectorization).

Note that the manual c++ is still 50% faster than the auto c++.
I need to fix the SubProcess and make sure they are all in sync,

They I also need to generate and test eemumu for this 'epoch2 ggttgg' version

valassi:
I have first merged PR #245 for step 1 of issue #244.

I am now merging this step2 PR #247.

Self-merging (I am ignoring the CI errors, one because the GPU CI node seems down and two because anyway this PR only creates an epochX directory, which is not being tested yet).

PR #248, opened 2021-09-28
Fortran cpp/cuda bridge studies

roiser:
studying how to combine fortran and cpp/cuda code.


roiser:
note to myself, transposition code wrapped into template class, probably extra step needed for having cuda and c++/template separated...

roiser:
The first steps for this work have been completed, i.e.

   - the bridge fortran/C(++) works
   - the pointer of the fortran 4-momenta is passed to the C++ side
   - the transposition of the fortran 4-momenta from row-major to column-major is done in a (templated) kernel
  
next step is to include the "golden tag" matrix element calculations. I propose to merge this work and continue with a new PR.

valassi:
@roiser I have seen you completed PR #248 about the cuda/fortran bridge (great!) and are waiting for a golden tag

If you agree I would merge PR #254 and I would then tag the result (eg as "epochX2" or "golden_epochX2")?

Then for your integration work I suggest you keep working in a "fortran-cuda" directory (in epochX or in epoch2, as you prefer...)?

PR #254, opened 2021-10-11
EpochX step2 bis - clean up gFile/File names, cc/cu suffixes and symlinks 

valassi:
This is an addition to step 2 of issue #244, previously merged in PR #247.

While the code there was functionining and "in a steday state", there were a few inconsistencies like duplicate files. Also, the strategy there was to consider .cu files as the main files, and cc files as symlinks. To prepare for step3 an dthe backport of eemumu epoch1, I revert this to using .cc files as the main files and .cu files as the symlinks, as I am doing in epoch1 eemumu.

The code is now again in a steady state. This is ready to merge. The result can essentially be considered "the golden tag" for epoch2 comparisons to alpaka, kokkos, sycl.

### DIFF
--- a/epochX/cudacpp/tput/throughputX_log_eemumu.txt
+++ b/epochX/cudacpp/tput/throughputX_log_eemumu.txt
@@ -8,22 +8,22 @@ On itscrd70.cern.ch [CPU: Intel(R) Xeon(R) Silver 4216 CPU] [GPU: 1x Tesla V100S
 =========================================================================
 runExe /afs/cern.ch/user/a/avalassi/GPU2020/madgraph4gpuX/epochX/cudacpp/ee_mumu/SubProcesses/P1_Sigma_sm_epem_mupmum/gcheck.exe -p 2048 256 12 OMP=
 FP precision               = DOUBLE (nan=0)
-EvtsPerSec[MatrixElems] (3)= ( 7.280263e+08                 )  sec^-1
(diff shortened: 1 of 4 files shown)


valassi:
@roiser I have seen you completed PR #248 about the cuda/fortran bridge (great!) and are waiting for a golden tag

If you agree I would merge PR #254 and I would then tag the result (eg as "epochX2" or "golden_epochX2")?

Then for your integration work I suggest you keep working in a "fortran-cuda" directory (in epochX or in epoch2, as you prefer...)?

roiser:
> @roiser I have seen you completed PR #248 about the cuda/fortran bridge (great!) and are waiting for a golden tag
> 
> If you agree I would merge PR #254 and I would then tag the result (eg as "epochX2" or "golden_epochX2")?
> 
> Then for your integration work I suggest you keep working in a "fortran-cuda" directory (in epochX or in epoch2, as you prefer...)?

Hi @valassi yes I would prefer for the time being to keep the work towards a full workflow separated in the fortran-cuda directory. Please note that I'm ONLY developing the code for now and don't look into code generation yet (this will come later). Another two caveats, I'm currently going into the direction of running cuda code for the ME calculations only (C++ vectorized later) and there are still the missing parts that @oliviermattelaer explained in the last meeting. I which are not implemented. I'll open an issue for this in a moment. 

Coming back to your previous point. Yes I would then use whatever you define as the golden tag and take this into the "workflow" directory.

PR #257, opened 2021-10-13
WIP: fortran-cpp bridge, incorporate gg_tt workflow from golden tag

roiser:
Branch to include the epochX/cudacpp/gg_tt.auto code of the [golden_epochX2](https://github.com/madgraph5/madgraph4gpu/tree/golden_epochX2) tag into the fortran/cpp bridge


epoch2/fortran-cuda/pp_ttx/SubProcesses/bridge.h
```diff
@@ -0,0 +1,352 @@
...
+    m_goodHelsCalculated = true;
+  }
+
+  gProc::sigmaKin<<<gpublocks, gputhreads>>>(devMomentaC2.get(), devMEs2.get());
+
+  //
+  //
+  //
+
+  checkCuda(cudaMemcpy(mes, devMEs2.get(), m_evt * sizeof(T),
```
valassi, line 137: HI Stefan, looks nice! I am having a look to understand better what we discussed this morning, namely splitting the bridge proper from the ME calculation engine. Two remarks on this line, which is a copy from devMEs to the mes.

I assume the mes is a double* array allocated in the Fortran, right?

First point: this will not work if we compile the ME calculation in cuda in float, while you have the mes as double from the Fortran. You need some explicit loop and conversion I guess.

Second point, which was the first one I noticed: for maximum performance of the copy device to host, I understand (I may be wrong) that it is better to use pinned memory, ie use cudaMallocHost for the destination.

I mention these two points because they go in the direction of what I had in mind: I would encapsulate the ME calculation in "my" part as taking input devMomenta, computing output devMEs, and also doing a copy to hstMEs (if required). Then "your" bridge would just need to fill in the devMomenta (after doing the appropriate copy and transposition) from the Fortran momenta, and fill the Fortran MEs from the CUDA hstMEs.

Normally I would allocate the devMomenta, devMEs and hstMEs in "my" class, but I would allow setting it from an external pointer if you pass it. The prototype is there and evolving in PR #302.

roiser: Hi Andrea, 

yes its a double array but the Bridge class is templated which will also ripple down to the cuda/C++ vector types. So keeping the two should be enough I presume. Or do you mean to have double on the fortran side and float in C++, does this make sense? 

Ok, I'll try the cudaMallocHost but didn't look yet into performance I must say. thanks

I briefly looked at the API class, IIUC its partly doing the same thing as the cpu/gpu sequence functions. It's fine for me to call this, but in that case it would be better to allow passing pointers indeed for everything that is host. For the 4-momenta you could also allocate them and give me the pointer, whatever you prefer. The MEs I will need to pass into the API as we want to fill them directly in the Fortran world when returning from the calculation. 

Stefan

valassi: Hi Stefan, thanks a lot, just repeating here what I sent via email.

So: I just committed the PR, the diff that you are currently interested in is probably this:
https://github.com/madgraph5/madgraph4gpu/pull/300/files#diff-d6e3597839440c2a5b48f277c56754f4f7c6ad1a9e398dde27fdac7e7b6b9e8bL259-R265
I hope this is what you expected? Just replace fptype_sv* by fptype* in the sigmakin interface.

However, I am also working now on some class prototype to encapsulate the functionality.
This should allow you to do just something like the following (making it up now...)
[I am checking out your https://github.com/roiser/madgraph4gpu/blob/edcc2ba9f77cfb70b13b9bfb1e48d5591998bfcc/epoch2/fortran-cuda/pp_ttx/SubProcesses/bridge.h#L197]
   Fortran -> gives you momenta and mes arrays
In the bridge you would  instantiate this new class, lets call it MEKL now for MEKernelLauncher, can become CPPProcess or whatever:
```
   MEKL mekl( nevt );
   hst_transpose( mekl.hstMomenta(), momenta, mekl.nevt() ); // transpose the fortran momenta into the array preallocated (aligned) by mekl
   mekl.computeHstMEs(); // compute the MEs (this is the slow part, but uses vectorization internally) into the array preallocated by mekl
   ... memcopy mekl.hstMEs() into mes, or actually loop and cast real to double if MEKL was compiled in float...
About the last point, see what I discussed in https://github.com/madgraph5/madgraph4gpu/pull/257#discussion_r761168723
```

Similarly in CUDA
```
   MEKL mekl( nblocks, nthreads, useHstMEs=true);
   ... cudaMemcpy(devMomentaF2.get(), momenta, ...); // copy the fortran host momenta to fortran dev momenta
   dev_transpose( mekl.devMomenta(), momenta, mekl.nevt() ); // transpose the fortran dev momenta into the array preallocated (aligned) by mekl
   mekl.computeDevMEs(); // compute the MEs (this is the slow part) into the array preallocated by mekl
   mekl.copyDevMEsToHstMEs(); // copy MEs dev to host into the array preallocated by mekl
   ... memcopy mekl.hstMEs() into mes, or actually loop and cast real to double if MEKL was compiled in float...
```

A couple of additions that I still plan
- of course the class also needs to take care itself of the get/set good helicities, you can remove this from your code
- one option, eg in the CPU case where maybe you do not need BOTH the fortran mes and the hostMEs, would be to pass the fortran mes array in the ctor, so that this one is used... (but then you have to make sure it is aligned... otherwise it will work, but it will be a tiny bit slower because I cannot reinterpret cast in the vectorization)  
- similarly one can imagine passing an external momenta array, but there the penalty is maybe 10% if it is not aligned

The idea then is that this MEKL is a prebuilt packagesd class that you just need to link in your bridge. No need for complicating the makefile etc...

valassi: I think we are converging. On your three points

> yes its a double array but the Bridge class is templated which will also ripple down to the cuda/C++ vector types. So keeping the two should be enough I presume. Or do you mean to have double on the fortran side and float in C++, does this make sense?

Not sure, but it's a possibility I am keeping in mind for the moment. Most likely we will use double anyway, but it would be nice to just do the ME in float and see what it gives. Even if the bridge is templated, I think that MadEvent in any case is always double (but maybe you do use assignments towards the bridge anyway, I did not check).

> Ok, I'll try the cudaMallocHost but didn't look yet into performance I must say. thanks

In the snippets above, this can be encapsulated by the `mekl.copyDevMEsToHstMEs();`... so no need to put it in the bridge

> I briefly looked at the API class, IIUC its partly doing the same thing as the cpu/gpu sequence functions. It's fine for me to call this, but in that case it would be better to allow passing pointers indeed for everything that is host. For the 4-momenta you could also allocate them and give me the pointer, whatever you prefer. The MEs I will need to pass into the API as we want to fill them directly in the Fortran world when returning from the calculation.

Yes I think we are converging :-)
This is exactly the last point I was mentioning, I can add the option to pass external array pointers and use those internally. There may be a tiny performance penalty especially for vectorization if the arrays are not aligned, but functionally it shoud work. I tried various options, should be around 10% for ggttgg. We can test/prototype variouus options and then decide

PR #258, opened 2021-10-13
Initial port of epochX for SYCL

lfield:
This port builds but requires further testing.


lfield:
Ready for merging into master.

PR #267, opened 2021-10-20
EpochX step 4: vectorize ggttgg (and ggtt)

valassi:
This is the beginning of a WIP PR for vectorizing ggttgg (and ggtt) code. This represents "Step 4" in issue #244.

It relies on the previous epochX3 PR #253, where I completed the backport of eemumu vectorization.

The ggttgg code generates out of the box, but there are a dozen build errors to fix. It will require a few iterations.

PR #270, opened 2021-10-20
Complete EpochX step 4: vectorize ggttgg (and ggtt)

valassi:
I merged too early (should not do it in the evening...)

I had not pushed yet the final commits.

Now this completes PR #267, ie the step 4 in issue #244


valassi:
Self merging

PR #280, opened 2021-10-26
Header-only MadgraphTest

### LINKED ISSUE #268: google test and CI - ggttgg test results (and test directory restructuring)
Hi @hageboeck (and @roiser @oliviermattelaer )

I am progressing in building ggttgg in vectorized mode, should be quite easy now that epochX3 is done.

I have interfaced my performance test scripts to also run functional tests and am having a few issues

1. Unless I am mistaken, there is no ggttgg test reference result file. Am I wrong? I only see the eemumu directory here
https://github.com/madgraph5/madgraph4gpu/tree/master/test/eemumu
If you cannot generate it, can you remind me how to generate it please? Otherwise I will find out, I think I have done it already for eemumu.

2. The eemumu ref file is hardcoded in the test presently
https://github.com/madgraph5/madgraph4gpu/blob/eea41614399c7b0b5cc9c00d726cad0fb4059741/test/src/MadgraphTest.cc#L78

3. Actually, even worse, I just realised that the code itself is called eemumu
https://github.com/madgraph5/madgraph4gpu/blob/eea41614399c7b0b5cc9c00d726cad0fb4059741/test/src/MadgraphTest.cc#L64
Is this test portable (eg to 6 external particles instead of 4) or does it need rewriting?

4. To solve point 2 above we probably need to use some gymnastics using process names as decided by madgraph. In that case it probably makes sense to rename the c++ methods and the directory and file names using the names of the P1 directory. This would be "epem_mupmum" and "gg_ttxgg". Then it becomes quite easy to generate with the code generator, see for instance this file that I added to make a nice printout when I am running a test
https://github.com/madgraph5/madgraph4gpu/blob/eea41614399c7b0b5cc9c00d726cad0fb4059741/epochX/cudacpp/ee_mumu/SubProcesses/P1_Sigma_sm_epem_mupmum/epoch_process_id.h#L6
(By the way Stefan is this the directory naming you would like to use everywhere also for consistency with Andy's tools?)

5. Eventually this also need to be integrated in the CI...

For the moment I will let the tests fail in ggttgg both manual and auto in epochX4. Initially I thought I'd give it a try, but especially point 3 above scares me a bit, this may be more work than it seems at first sight?

Any feedback/suggestions?
Thanks
Andrea

hageboeck:
## Header-only `MadgraphTest`
In order to make it easier to instantiate the test suite with the floating-point type that madgraph uses, the test suite is converted to header only here.
This allows for pulling the compilation of tests down into each implementation
directory, directly using the floating-point type that madgraph uses.
This also allowed for removing some test instantiation logic.

### Adapt epoch{1,2,X}:
Since there is no more object file in the test/ directory, the
Makefiles of the cudacpp codes had to be simplified a bit:
The `runTest` target now only needs to ensure the presence of googletest and adds
an include statement for `MadgraphTest.h`. The previously dedicated compilation step
is thereby replaced entirely by including a header.

## NB:
This is already running tests in the gg_ttgg folder for epochX.

## TODO:
- [ ] Update code generator

This should fix #268


hageboeck:
Hi @valassi,

It took quite long to figure out how to make it mergeable. Commits were fractured over my and your branch, and master advanced in the mean time, so reconverging the branches proved to be tricky. I managed to get it done via a rebase of your and my branch onto master, and fixing some conflicts on the way.
Do you also see frequent merge conflicts here (or similar locations where run times vary)?
https://github.com/madgraph5/madgraph4gpu/blob/286abc999a1b60923161108d9235292e0e15272e/epochX/cudacpp/gg_ttgg.auto/CODEGEN_gg_ttgg_log.txt#L135-L138

From my side, I would say it's done now. What do you think?

NB: I don't manage to fix the GPU node. Maybe the new driver doesn't play along with the container. I'm asking for a dedicated node now.

hageboeck:
Update: The centos stream migration enabled me to fix the runner as well. Let me know if I can merge.

valassi:
> Hi @valassi,
> 
> It took quite long to figure out how to make it mergeable. Commits were fractured over my and your branch, and master advanced in the mean time, so reconverging the branches proved to be tricky. I managed to get it done via a rebase of your and my branch onto master, and fixing some conflicts on the way. Do you also see frequent merge conflicts here (or similar locations where run times vary)?

Hi @hageboeck thanks a lot!

I am sorry this was more tricky than it whould have been. I thought I had fixed all issues and that a merge would be easy, but obviously it was not, apologies about that.

Looks good to me, I will merge this.

Andrea

valassi (approved):

Issue #260, opened 2021-10-15
EpochX sycl remove hardcoded parameters


lfield:
Some constants were hardcoded to avoid passing them to the GPU. This hack needs to be reversed. 

https://github.com/madgraph5/madgraph4gpu/blob/sycl/epochX/sycl/gg_ttgg/SubProcesses/P1_Sigma_sm_gg_ttxgg/CPPProcess.cc#L69

lfield:
Fixed in PR #286

PR #286, opened 2021-10-29
Fix for hardcoded cIPC and cIPD

lfield:
Fix for Issue #260. The variables cIPC and cIPD following the same method as cHel.


nscottnichols (approved):
Compiles and runs without issue. Code looks good and consistent with `cHel` workaround

Issue #261, opened 2021-10-15
EpochX sycl << operator needed


lfield:
The print statements have been commented out  as the <<operator needs to be implemented for the complex number class

https://github.com/madgraph5/madgraph4gpu/blob/sycl/epochX/sycl/gg_ttgg/src/Parameters_sm.cc#L102

lfield:
Fixed in PR #288

PR #288, opened 2021-10-29
Adding << operator for complex numbers

lfield:
This PR fixes issue #261.


nscottnichols (approved):
Compiles and runs without issue. Tested new operator by adding `std::cout << "TEST: " << cxtype(0, 1) << std::endl;` at the beginning of main and it works like a charm.

Issue #276, opened 2021-10-25
Reassess ggttgg (and eemumu) throughput in Fortran without helicity recycling


valassi:
A brief followup from today's meeting, https://indico.cern.ch/event/1081635/

Presently I see 1.8E3 throughputs in C++ for ggttgg without SIMD. In the past I have evaluated the Fortran ME throughput as 4.0E3, which is more than a factor 2 higher. 

For comparison, Fortran eemumu was evakluated in only 20% faster than C++ no SIMD.

Olivier suggested that there is a factor ~2 speedup in the latest Madgraph from helicity recycling, and this is not yet in our C++/CUDA. Maybe I had used a recent Madgraph where helicity recycling was implemented? For an apple to apple comparison, shoulc compare to Fortran without helicity recycling.

valassi:
I am working on a WIP PR #298

I have regenerated gridpacks and computed throughputs for revisions 365 and 370 in 2.7.0_gpu, which correspond essentially to 2.8.x and 2.9.5. The former has no helicity recycling, the latter does.

This exercise was quite surprising. I realise I was using MATRIX1 throughputs for comparisons to CUDA/C++. But from 365 to 370 the "throughput" of MATRIX1 decreases (by factors 2, 3, 15 for eemumu, ggtt, ggttgg), instead of increasing!

Probably I should use instead something like the SMATRIX1 throughput? This does increase by around 1.2x for eemumu and 3x for ggtt and ggttgg.

See the following:

For 28x with helicity recycling
commit 20b4335ba12ec98144bfb8af3d6da4c497375dd1
```
    tput/log_ee_mumu_auto.txt:TOTAL MATRIX1  :    0.0089s for    24520 calls => throughput is 2.76E+06 calls/s
    tput/log_ee_mumu_auto.txt:TOTAL SMATRIX1 :    0.0273s for     6118 calls => throughput is 2.24E+05 calls/s
    tput/log_gg_tt_auto.txt:TOTAL MATRIX1  :    0.0324s for    59980 calls => throughput is 1.85E+06 calls/s
    tput/log_gg_tt_auto.txt:TOTAL SMATRIX1 :    0.0514s for     4997 calls => throughput is 9.72E+04 calls/s
    tput/log_gg_ttgg_auto.txt:TOTAL MATRIX1  :   22.8280s for  1715136 calls => throughput is 7.51E+04 calls/s
[...]

Issue #279, opened 2021-10-26
Implement helicity recycling in our CUDA/C++


valassi:
This is a followup of issue #276.

In order to compare C++/CUDA and Fortran throughputs, we should make sure that they use the same algorithm. This is presently not the case: we are comparing a faster Fortran with helicity recycling to a slower C++ without helicity recycling.

In issue #276, I will follow up a better estimate of slower Fortran without helicity recycling, that can be directly compared to C++.

But what we really need to do is implement helicity recycling in the CUDA/C++. @oliviermattelaer is this something that would be complicated (an/or maybe is already underway)? Thanks

oliviermattelaer:
I would actually doubt that doing helicity recycling for gpu is a good idea since this blows up the size of the code and the memory requirement. For vectorised cpu, that is obviously an option.

valassi:
Within PR #401, note that I had to introduce this fix while moving to v311
https://github.com/madgraph5/madgraph4gpu/pull/401/commits/04d4b8e41b7386164bdfde2277f1743d7ebb9a18
This indeed picks up a new feature of v311 (#360), but it is only relevant to helicit yrecycling (#279).

In practice
- initially I started generating code (madevent only, or madevent+cudacpp) without disabling heliicty recycling
- under this configuration, code generation of madevent+cudacpp did not work, and I had to fix it as above
- this was especially puxxling because I had to change some code generation lines (for generating cudacpp) which were not used in my normal cudacpp-only version
- after fixing it, I got code with helicity recycling, which however did not build (not clear why, but not an issue today): to fix the build I disabled helicity recycling
- in retrospective, the reason why I had to fix those lines is thus for helicity recycling code, which in any case I later decided to disable: if I had disabled helicity recycling before trying to generate madevent+cudacpp code, probably that patch would not hav ebeen necessary

PR #298, opened 2021-11-15
gridpackX (reassess Fortran throughputs with/without helicity recycling)

valassi:
This is a WIP PR for issue #276 about reassessing Fortran throughputs


valassi:
This completes the scripts and code needed to study #276. I am self merging.

Note that this does not address helicity recycling in cuda/c++, which must be done in #279

PR #300, opened 2021-11-30
Remove neppV and fptype_sv from API and decouple neppV from neppM (epochX)

valassi:
This is a followup (in epochX) to PR #208 (in epoch1) about issue #176

Essentially it is about removing neppV and fptype_sv from the public APIs that we will offer users

The details of how I arrived to some pieces of code is in klas5 PR #208. In this new PR from klas5X, I have just immediately committed to epochX the results of old developpments in epoch1. I think that I will merge in any case also those epoch1 developments to keep the history.

Note also that this epochX status needs to be cleaned up. I am supporting many options (unlaigned data, neppM smaller than neppV etc) that we should not support in production code. We should just keep the fastest, which is aligned data, with neppM=neppV. I have already added some sanity checks, but I should cross check those, and keep only the strictly relevant ones (also because some sanity checks are expensive, eg on alignment of every buffer)

### DIFF
--- a/epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/aloha/template_files/gpu/helas.cu
+++ b/epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/aloha/template_files/gpu/helas.cu
@@ -1,7 +1,7 @@
   //--------------------------------------------------------------------------
 
-#ifdef __CUDACC__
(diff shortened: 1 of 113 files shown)


valassi:
I have just self merged PR #208 which included the preparatory work for this in epoch1 eemumu.

I am now self merging this PR #300 about epochX, including code generation.

valassi:
I have now closed issue #176 as this PR has completed it.

Issue #301, opened 2021-12-02
API (5) Encapsulate the ME calculation in a single class


valassi:
This is a placeholder for one of the next big steps in the development process: an API to encapsulate the ME calculation in a single class. Let's call this a "ME engine" or "ME kernel launcher" for the time, say MEKL.

The idea is to make it easy to plug and play the ME calculation into different types of workflows, these two in particular:

1. A standalone C++/CUDA application as we have it now, specifically:

(1a) CPU only, pure C++. A sampler produces hstMomenta on the host. The MEKL produces hstMEs on the host. 

(1b) GPU, in C++/CUDA. A sampler produces devMomenta on the device. The MEKL produces devMEs on the device. Someone (probably another component, not MEKL?) may also need to copy devMomenta to hstMomenta on the host and/or devMEs to hstMEs. But neither of these two extra tasks needs to necessarily happen all the time: you may decide to only compute cross sections and not need to copy momenta to host as hstMomenta (for writing to file) at all, and you may want to aggregate devMEs into avreage cross sections on the device, without copying them to hstMEs at all.

2. An external application such as the MadEvent-based Fortran application, specifically:

(2a) CPU only. A sampler like MadEvent produces momenta on the host. These may be hstMomentaFortran in Fortran matrix convention, which may need a transposition to hstMomenta in the C++ matrix convention needed by the MEKL. The MEKL produces the hstMEs: this is the same components as in case (1a) above.

(2b) GPU. A sampler like MadEvent produces momenta on the host. These may be hstMomentaFortran in Fortran matrix convention, which may need a transposition to hstMomenta in the C++ matrix convention needed by the MEKL. Then someone copies the hstMomenta to devMomenta (or maybe it is the hstMomentaFortran that are copied to devMomentaFortran and it is then the GPU that does the transposition to devMomenta - I think this is what Stefan does now). Then the MEKL computes devMEs from the devMomenta, this is exactly the same component as the MEKL above in 1b. Then the relevant outputs (devMEs individually, or some aggregates) must be passed back to the host and back to the Fortran framework.

In all of the above, the common core component (which is what spends most of the compute time for complex processes) is the MEKL computation of hstMEs from hstMomenta (CPU) or devMEs from devMomenta (GPU). This is more or less the current sigmaKin function, with a few things to change
- before the klas5X PR #300, hstMomenta was an fptype_sv* and devMomenta was an fptype*; after that PR, both will use a simple fptype* array
[...]

PR #302, opened 2021-12-02
WIP: prototype API to encapsulate the ME calculation in a kernel launcher MEKL 

valassi:
This is a first draft prototype for issue #301.

Note that it has only very few commits on top of the klas5X branch PR #300 (which I plan to merge soon)


valassi:
I have finally chosen a different approach and implemented that in #324.

This new approach, that is also implemented in the random and rambo kernel launchers, is to create host/device memory buffers outside (MemoryBuffers.h, see #320) and then pass them to the constructors of the kernel launchers. This is much cleaner and readable.

I am closing this PR #302.

PR #303, opened 2021-12-06
Install target

roiser:
Add install target at the code generator level which will export all necessary functionality to use the ME binaries from outside. The target will 

  - create an "include" directory on the top level directory
  - copy header files into the include directory as necessary
  - copy gCPPProcess.o and CPPProcess.o into the lib directory

The "clean" target was modified to remove all installed artifacts.


valassi:
Hi Stefan, as discussed, it would be nice to have all ggtt, ggttgg, eemumu generated, and with no diffs between manual and auto.

Typically this is what I do:
```
export MG5AMC_HOME=/data/avalassi/GPU2020/MG5aMC/2.7.0_gpu
./CODEGEN/generateAndCompare.sh gg_tt [or gg_ttgg, or ee_mumu]
./CODEGEN/diffCode.sh gg_tt gg_tt.auto -r
```

Otherwise two small things we discussed
- now make install does not work with "USEBUILDDIR=1", ie it only installs the version that is locally built
- in that case, I have forgotten to copy the .build.tag file (eg  .build.none_d_inl0_curdev) to the lib directory

These two small issues can be fixed later on, for the moment you can just use the default installation. Remember that by default you will get 512y/double builds. You must pass AVX= and FPTYPE= to change those if you want, see https://github.com/madgraph5/madgraph4gpu/wiki/Building-madgraph4gpu-and-measuring-throughput

Thanmks
Andrea

Issue #304, opened 2021-12-07
support USEBUILDDIR in make install - and copy .build.tag file to lib


valassi:
Two minor issues as discussed in https://github.com/madgraph5/madgraph4gpu/pull/303#issuecomment-987798556

-    now make install does not work with "USEBUILDDIR=1", ie it only installs the version that is locally built
-    in that case, ie if USEBUILDDIR is unset, I have forgotten to copy the .build.tag file (eg .build.none_d_inl0_curdev) to the lib directory

valassi:
This is completed in PR #305 (which also does a few more things)

valassi:
This can now be closed as PR #305 has been merged

PR #305, opened 2021-12-07
'make install' improvements

valassi:
A few improvements in 'make install'

As discussed in #304:
-    now make install does not work with "USEBUILDDIR=1", ie it only installs the version that is locally built
-    in that case, ie if USEBUILDDIR is unset, I have forgotten to copy the .build.tag file (eg .build.none_d_inl0_curdev) to the lib directory

In addition:
- improve the logic so that .h and .o are only copied if they are not already there
- add an avxall-install target


valassi:
Hi @roiser let me know if you hav eobjectsions or suggestions
thanks
a

valassi:
Argh I need on etiny fix, the modelsm stuff in the makefiles..

valassi:
ok now ready for review and merge after a small fix

roiser (approved):
all good, thx

Issue #39, opened 2020-10-29
Investigate masses/couplings in constant memory instead of compile time constants


valassi:
Hi @oliviermattelaer, this is a followup of our chat this afternoon and a spinoff of issue #23.

We should check again the effect of having masses/couplings in constant memory instead of compile time constants. This commit saves 4 registers (down from 178 to 174)
https://github.com/madgraph5/madgraph4gpu/commit/6679569f005a0b1caacf12f7a226933c2d3b2cbe

Is this because some values are hardcoded to 0? Or is this irrelevant?

If it is because they are 0, then maybe we could move the others to constant memory, but still keep in mind that couplings are only purely real or purely imaginary (is this true?), and thus save the registers, while still allowing variable couplings (in constant memory) that are not hardcoded at compile time.

Thanks, Andrea

_Originally posted by @valassi in https://github.com/madgraph5/madgraph4gpu/issues/23#issuecomment-718857719_

valassi:
I have just opened PR #306. I copy the relevant text here

This is a followup to #23 and #39

Yesterday while working on unrelated things (older WIP in klas3/klas3base in eemumu epoch1) I realised there was a performance regression of around 20% in cuda for eemumu between epoch1 and epochX, because epoch1 was using hardcoded parameters while epoch1 was using parameters read from file and then set in constant memory.

I moved also epoch1 to the (slower, but more default) reading f parameters from files. However, I also added the option to use hardcoded physics parameters, bth in c++ and cuda, if an ifdef switches them on. This PR is the result of that.

The implementation is fully functional, but it includes a few hacks which should be improved. The complication comes from the fact that one should use constexpr in the calculation of derived parameters, and there are two issues:
- first, sqrt is not constexpr: I worked around this by using a nice solution based on Netwon Raphson that I found on SO, and this is code that can remain
- second, a lot of complex arithmetics is also not constexpr, and something as simple as multiplying two complex numbers is not constexpr; I worked around this by massaging manually a few formulas, and only including all constants that are relevant to eemumu, ggtt and ggttgg

Eventually, the hack of the second issue above should be removed
- the easiest would be to use a custom complex implementation, which is something that I had in mind anyway to go beyond std, thrust and all others (we really only need basic + - * /)
- once that is fixed, it is possible to derive automatically generated code, so one would have to backport in python (essentially a third set of methods, merging the two presently used for the declaration in .h and assignment in .cc)

This PR in any case is now fully functional (not WIP) and can be merged

[...]

Issue #23, opened 2020-08-17
Compile-time const (and constexpr) vs constant-memory constants (layouts; physics parameters?)


valassi:
I dump here a few observations I made while analysing AOS/SOA/ASA memory layouuts is issue #16. Th ebottom line is that there is a (small) performance penalty in using constant memory for variables that could instead be taken as compile time constants.

In that context, I want to compare a few different layouts (eg ASA with 1, 2, 4.. events per "page"). Initially my code was only allowing some parameters to be defined at compile time, so every different test needed a rebuild. I then thought of allowing runtime choices of these parameters. 

This is implemented here for instance: https://github.com/madgraph5/madgraph4gpu/commit/576ba405215afe49d4fd1288eba25ef7af2ea168. In this version both the random and momenta lyaouts (neppM and neppR) are ready to be received as command line options (not done fully yet). For neppR, this is passed around in function signatures. For neppM, this wwould have been too complex an dI pass it around in device constant memory.

Anyway, the point is that then I compared, for the same choice neppM=32, the difference between the compile time and constant memory constants. It is clear that there is a performance penalty, even if small. This is at the level of 5%, eg throughput goes down from 6.0E8 to 5.0E8 (all these numbers have the usual large spread due to VM load). I compared with ncu the profiles, and this is quite interesting. I would say that the problem is not an extra load on memory (constant memory is fast), but rather the fact that there are extra arithmetic operations which somehow can be avoided if the constants are known at compile time.

This is the overview
![image](https://user-images.githubusercontent.com/3473550/90365041-82dc0f00-e065-11ea-8699-435c80656f0d.png)

One of the most interesting plots is the instruction statistics. Here it is clear that the constant memory version needs many more IMAD (integer multiply and add) operations. I believe these are the runtime calculation of the momenta indices, which comehoe can otherwise be avoided if the constants are know at compile time
![image](https://user-images.githubusercontent.com/3473550/90365369-f251fe80-e065-11ea-88f0-945069f0adf7.png)

Also interesting to know that 2 more registers are used, but this has impact on occupancy
![image](https://user-images.githubusercontent.com/3473550/90365511-2af1d800-e066-11ea-9de7-395975c6a845.png)

Conclusion: if possible, it is best to use compile-time constants rather than constant memory. The difference is small, but it's probably worth it to avoid the penalty unless necessary.
[...]

PR #306, opened 2021-12-09
Option to hardcode physics parameters (with hacks to remove)

valassi:
This is a followup to #23 and #39

Yesterday while working on unrelated things (older WIP in klas3/klas3base in eemumu epoch1) I realised there was a performance regression of around 20% in cuda for eemumu between epoch1 and epochX, because epoch1 was using hardcoded parameters while epoch1 was using parameters read from file and then set in constant memory.

I moved also epoch1 to the (slower, but more default) reading f parameters from files. However, I also added the option to use hardcoded physics parameters, bth in c++ and cuda, if an ifdef switches them on. This PR is the result of that.

The implementation is fully functional, but it includes a few hacks which should be improved. The complication comes from the fact that one should use constexpr in the calculation of derived parameters, and there are two issues:
- first, sqrt is not constexpr: I worked around this by using a nice solution based on Netwon Raphson that I found on SO, and this is code that can remain
- second, a lot of complex arithmetics is also not constexpr, and something as simple as multiplying two complex numbers is not constexpr; I worked around this by massaging manually a few formulas, and only including all constants that are relevant to eemumu, ggtt and ggttgg

Eventually, the hack of the second issue above should be removed
- the easiest would be to use a custom complex implementation (#307), which is something that I had in mind anyway to go beyond std, thrust and all others (we really only need basic + - * /)
- once that is fixed, it is possible to derive automatically generated code, so one would have to backport in python (essentially a third set of methods, merging the two presently used for the declaration in .h and assignment in .cc)

This PR in any case is now fully functional (not WIP) and can be merged


valassi:
This is only adding an optional feature - I am self merging to more easily sync with other developments

PR #312, opened 2021-12-10
Minor tests (-O0) and fixes (find) in Makefile's

valassi:
Test if the code works with -O0 instead of -O3: functional results are ok, but C++ loses a factaor 20 or more in performance!

MInor fix in the 'find' statements to silence errors if 'find' is applied to non existing directories


valassi:
Self merging, shoud be non controversial

Issue #310, opened 2021-12-09
Split the sigmakin kernel into smaller kernels


valassi:
This is just a placeholder to discuss the idea of implementing smaller kernels.

A lot of pointers already exist related to this:
- Stefan's WIP PR #242
- issue #12 about cuda graphs
- issue #11 about cuda streams
- issue #175 about the internal APIs

One of the main points towards using smaller kernels is the need to allow each ixx/oxxx and each ffv function to handle pointers to large buffers for many events and to do the indexing themselves. This is discussed in https://github.com/madgraph5/madgraph4gpu/issues/175#issuecomment-988215980 for instance. Presently instead only the ixx/oxx functions are able to find an event in the input array, but then their output (and all inputs/outputs of the ffv functions) refer for CUDA to a single event. This is the first thing that must be changed to allow smaller kernels.

valassi:
Note a basic prototype idea in #313: it would be enough to add for instance bufferAccessWavefunction( fptype* buffer, int iw6 ) wherethe iw6 index is passed in the same way as the ip4 now.

valassi:
Note that PR #328 is the first step for this and contains several comments related to this

valassi:
The color algebra optimisation has a separate issue #155. Improving the timing measurements is in #372.

valassi:
One of the main issues in splitting kernels is the clarification of the relative roles of MEK and CPPProcess: who holds intermediate data buffers? who orchestrates the order of kernels? who is allowed to have process specific stuff? I am discussing this largely in #356, specifically in the econtxt of running alphas issue #373 and draft PR #434

valassi:
Kernel splitting is implemented in PR #1049 (merged) and PR #1066.
Full documentation in https://arxiv.org/abs/2510.05392v2 that should appear tomorrow

Issue #309, opened 2021-12-09
separate buffers for different particles (SOAOSOA?)


valassi:
This is about the second issue discussed in https://github.com/madgraph5/madgraph4gpu/issues/175#issuecomment-988656992

Eventually, it may better to change the buffer indexing so as to have completely separate buffers for different particles. This would simplify the ixxx/oxxx api (no need to pass ipar) and make the code more understandbale. However it is clearly something to be synced with the fortran bridge work. Not urgent.

One possibility could be to move from the current AOSOA to an SOAOSOA (or back to an SOA):
- now we have momenta as AOSOA[npagM][npar][np4][neppM}, with npar particles, np4=4 components of 4-momenta, and nevt=npagM*neppM events split into npagM pages of neppM events per page
- eventually we could have SOAOSOA[npar][npagM][np4][neppM], so that it is easy to split these into several separate AOSOA[npagM][np4][neppM]
- or we could use SOA[npar][np4][nevt=npagM*neppM], although some tests I had done long ago seems to say there was a penalty f rthis (tests to be repeated)

Anyway, it could be useful to investigate this

valassi:
This can be closed. As far as momenta are concerned, the new MemoryAccessMomenta class does the jobs taking an ipar argument both in CUDA and in C++. 
See https://github.com/madgraph5/madgraph4gpu/pull/322/files#diff-66d130d9cd4df7a88bc6d32cff137904ac4f1c7f55e9ba05188cda231212318b in PR #322

PR #313, opened 2021-12-11
New buffer access in xxx (and ffv) functions - remove ipagV from C++ api

valassi:
This is a WIP PR for the ideas discussed in the first bullet of https://github.com/madgraph5/madgraph4gpu/issues/175#issuecomment-988656992

The idea is to start separating CUDA and C++, so that CUDA does a data access in a large buffer inside the XX functions (and eventually in the FFV functions to split into smaller kernels), while C++ takes a single event or at most a SIMD vector of events.

In practice this means removing ipagV from many interfaces.

Note that ipar is still there for CUDA - this cannot be changed unless/until a new memory structure is introduced, as discussed in the SOAOSOA #309 

Note also that ip4 is still in the interface of the bufferAccessMomenta. I think this is unavoidable for CUDA: we want to have an AOSOA for events and 4-components, with a striding not fixed a priori. This also does not harm for c++: it could be avoided as one could just get an fptype_sv[4] and index it in a very intuitive way, but it does not harm to have it as a function.

The same idea could be used eventually for splitting the kernels (#310): it would be enough to add for instance bufferAccessWavefunction( fptype* buffer, int iw6 ) wherethe iw6 index is passed in the same way as the ip4 now.

This is WIP, a lot to do, but the basic idea is there now.


valassi:
This is now complete thus far and can be merged. Main changes
- removed ipagV and ipar from c++ internal APIs
- started differentiating cuda and c++ (cuda xxx functions dela with large arrays, c++ deals with a single event or event vector)
- renamed and moved indexing functions to MemoryAccess.h
- for the moment, I did keep the complex logic to deal with aligned, unaligned and arbitrary arrays (it can be simplified later on), I actually removed sanity checks and assumptions, and reenabled alignment checks at every call, the overhead is minimal

valassi:
Self merging

PR #315, opened 2021-12-14
WIP: move Stefan's bridge into core CUDA/C++ directory

valassi:
Hi @roiser this is a WIP PR to move your Bridge class into the core CUDA/C++ directories (following up on #257)

I like your interface, extremely simple with just one T* in and one T* out. From Fortran we do not need much more.

Concerning our previous discussions on packaging, I think that the easiest is to move this Bridge here in the main C++/CUDA. Then you just need to load the object/library from Fortran.

In a way, this is a replacement (or a wrapper) for what I discussed in the MEKernelLauncher API #301 and #302. It can actually be a wrapper, in the sense that inside this Bridge then many things can be changed. 

It actually simplifies things enormously to have this here, because then we know exactly how the Fortran will call this, and we can simplify many things (I can simplify and remove a lot of junk I have introduced "just in case", which turns out to be mostly useless, eg the non-reinterpret-cast case in C++). Also I can easily switch to SOAOSOA or other formats, as everything is encapsulated in teh same set pf files/classes.

What I would plan is to add one option (commmand line option, or ifdef option, will decide later) to build also a special version of the check/gcheck executable, "emulating" Fortran an dusing the bridge:
- on CPU, contine to use CPU random numbers and CPU rambo, but then transpose C++ momenta to Fortran Momenta, and then call the CPU bridge that starts from Fortran momenta
- on GPU, as above use CPU random numbers and rambo (unlike gcheck that now does these on the GPU), again transpose c++ momenta to Fortran molmenta, and call the GPU bridge 
This just means that in the Bridge one also needs the inverse transposition CPP to Fortran, which in any case is a good idea for tests. 

Note by the way that now that sigmakin uses fptype and not fptype_sv, one can in principle remove the transpose2 (the ME transpose), as both on Fortran and on C++ these are simple arrays. Or maybe I would still leave it as is, for two reasons, one inside the Bridge (or inside its internals) one can use cudaMallocHost where needed for faster GPU-to-CPU transfers), and two one can in principle try to delegate a Fortran double calculation to a ME float calculation.

How does this sound? Lets discuss later
Andrea

PS This is WIP: here so far I only committed <10% of what this should be eventually...


roiser:
Hi @valassi , thanks for ploughing ahead. Few things on this PR

I have another version of the bridge in my github fork which is simpler than the one here. No PR yet, but will do it today so its visible. 

Please also note that the interface may become much richer, e.g. there are many parameters that need to be passed from Fortran to C++/Cuda (# evts, # particles, stride, blocks, threads and probably more physics stuff which I didn't need to for this simplified version). 

Yes the second transposition should be removed after the new layout of the data structures with aligned members. 

The few lines of the cpu_sequence and gpu_sequence functions should be probably what you implemented in the KLAS interface and should not be duplicated (including the variables becoming class members and not part of the function). 

The most important thing for me is where these basically two files should go. If the functionality of the two is "call the C (and C++) function to go into C++/Cuda" and "transpose from row-major to column-major)" to me semantically this belongs to the fortran code. Technically and from a build infrastructure point of view it may be easier to have them close to the ME calculation as otherwise one needs to replicate all the complicated machinery for C++ building on the Fortran side ... Being pragmatic I guess it should be on the ME side... 

Related to this. Do I understand correctly that you want to create the C++ sequence before ME, transpose everything C++ -> Fortran then call the bridge transpose it back and continue? Wouldn't it be easier to connect this right away to the Fortran executable? 

Let's discuss later ...

valassi:
This has been replaced by #325, which I am about to merge. Closing this one as superseded.

PR #317, opened 2021-12-17
API (1) Random number classes - runtime choice of random numbers

valassi:
This is another PR along the lines of #302 and #315. I will probably merge it before those.

While working on the bridge test, I realised that I needed some substantial changes to random numnbers too. THis is the corresponding PR. Random number generation is now encapsulated in "RandomNumberKernel" classes a la MEKernelLauncher. This can be used with the bridge and with the MEKernelLauncher. The names of the classes may change, but I wuld syat the structure looks ok.

One important aspect of this was also to simplicy and improve the choice of random numbers. Now teh choice between curand host, device and common random can be done at runtime. There is still a build time switch because curand may not be available in the C++ build, but this is handled separately.

Still WIP but almost done. Just missing code generation, testing on ggtt/ggttgg, and also changes to the wiki.


valassi:
Ah I forgot to say, one point was also moving curand generation out of rambo, as it does not belong there at all. This is done.

valassi:
I realise the twiki was not mentioning RNDGEN as an option for curand vs common. Better this way, indeed it is some sort of hidden variable. Now that it has changed, there is nothing to change on the twiki. One thing less to do.

valassi:
This is now complete including the full backport.

I will not merge in master yet for a while, but I will include this as step 1 and as teh basis of the other PRs for bridge and the ME kernel (I will add one for the rambo kernel too)

valassi:
This was presented at the meeting today
https://indico.cern.ch/event/1103955/
I am merging this now

Issue #318, opened 2021-12-17
Clarify build strategy for heterogeneous applications (and clean all build options)


valassi:
I am opening an issue that is a bit of catch-all container in the area of build options, of c++ vs cuda, of host vs device.

This started off from the work I want to do to integrate the bridge, with a simple test emulating fortran random/sampling to connect to cuda ME.

Take a component like rambo for instance. This can do the work on the host or on the device, even if in both cases the ME is computed on the device. The point is that I need a build of the c++/host version of rambo to link to the cuda/device version of the GPU. So far, fo rthings like rambo we only had EITHER a gcc build of the c++/host version, OR a nvcc build of the cuda/device version. Now I'd like to add also a c++/host version that I can use with the ME cuda. (All these issues will become the norm for truly heterogeneous workloads as in #85).

The easiest would be essentially to build rambo c++/host with nvcc. After all, nvcc is a c++ compiler too. In a way, it would be nice for instance to test SIMD C++ vectorization in an nvcc build. The problem is that setting simply CXX=nvcc runs into various other issues. Some may be fixed using the options to forward unknow to compiler/linker, but not all issues. There are also some -ccbin and -Xcompiler to clean up. Also, is CXXFLAGS really needed on all link instructions in the Makefile? There is quite some cleanup to do.

On the code side, there are (my fault) many different namespaces for cuda and c++. I am converging on the idea of having just two, say mg5amcOnCpu and mg5amcOnGpu: the latter is for a __CUDACC__ (ie nvcc) build, the former for a gcc/clang build.

Also on the code side, setting things like rambo as both __device__ and __host__ should ensure that a single nvcc build makes it usable both on the CPU and on the GPU.

So, in principle, one could aim for
- CPU-only application: mg5amcOnCPU namespace, build all using your favorite gcc/clang/ipcx compiler
- CPU+GPU application (which for instance requires cudaMallocHost instead of malloc on the host): mg5amcOnGpu namespace, build all using nvcc, making sure that it delegates the c++ stuff correctly to your favorite glcc/clang/icpx compiler (so in principle you should get the same results)

This is not urgent, but about some of the issues it's better to think earlier rather than later..

valassi:
On second thought. It does not make sense to build the c++ simd versions with nvcc anyway also because it uses a different definition of neppV and fptype_sv every time, the SIMD types are different. One probably needs to do separate builds and link them together, as in the multi simd idea #177.

[...]

PR #319, opened 2021-12-17
WIP: a few fixes to use CXX=nvcc

valassi:
This is a WIP PR related to #318.

I managed to build the code using 'export CXX=nvcc' and get good SIMD performance from the underlying gcc102.
So there should be no problem whatsoever building the SIMD C++ kernels using nvcc.

This is WIP because I still need to backport to code generation.


valassi:
Two more issues to fix, in total
- backport
- fix -pthread in build with default gcc (was ok with nvcc, no longer gcc)
- fix the CI build

valassi:
Some of these points have been added in PR #354, but the main points of the PR are not (probably ar enot needed)

roiser:
Hi @valassi I'm only coming late to this PR. IIUC this is about building the C++/CPU code with nvcc? Or is it just to show that we could also use nvcc?

valassi:
This is old, I will close it. There are #318 and #531 that remain open.

Issue #323, opened 2021-12-24
Workplan for January


valassi:
This is a dump of the things I plan to do in January after the break.

It touches the internal APIs for kernel launchers, the bridge, the memory access, going further to MT, heterogeneous apps and split kernels. All of these things are tightly interconnected.

One of the main ideas of my recent development was the separation of calculations (kernels and kernel launchers), memory buffers (MemoryBuffer.h) and memory access (MemoryAccessMomenta.h and related templated classes).

Another related main point has been allowing __host__ functions also in CUDA (#318). This is related to the idea of a test for the Fortran bridge.

So far this is achieved
- API1 PR #317, random number kernel launchers
- API2 PR #320, memory buffer classes
- API3 PR #321, rambo kernel launchers and new templated memory access classes

This is WIP and very urgent
- API4 PR #322, use the new memory access classes in HelAmps (turn HelAmps functions into header-only templated functions)

The next urgent steps are the following. I have a few bits and pieces of code in various PRs, but some must be reviewed in terms of the above
- API5, a new apimes branch/PR for #301 replacing PR #302, matrix element kernel launcher classes (differently from 302, separate MemoryBuffers in a separate class): I assume this will still be with SIMD MEs in gcc and CUDA MEs in nvcc (no SIMD host MEs in nvcc)
- API6, bridge, completing PR #315 (reimplement that in terms of the MEKL of API5), add a --bridge option to check.exe with random/rambo on host, transpose to fortran momenta buffers, then launche the bridge on host or on device

More forward looking, a tiny bit less urgent, but toghtly related
- API7, "full SA workflow" class, encapsulating the stuff that is now in check_sa.cc (but also in runTest.cc): start from no input and come up with ME averages and full events
- API8, use the above API7 to implement host multithreading SA, multi-GPU SA, heterogeneous SA (replacing PR #159) 
- API9, add MemoryAccessWavefunction classes, add MemoryBufferWavefunction classes, modify HelAmps with a second WV_ACCESS template class parameter to implement kernel access to wavefunction buffers from xxx/ffv functions, use all of this for splitting the calculate_wavefunction kernel (PR #242) and possibly for color algebra on tensor cores (#155, #118)

A few other points that are at the level of wishes, to be reevaluated, and in any case not urgent
- Should fptype/cxtype become also template parameters, specifically in xxx/ffv functions in HelAmps? Not urgent (but it would be nice to allow a bridge test with rnd/rambo in double and the MEs in float, for instance)
[...]

PR #325, opened 2022-01-07
API(6) Integrate Stefan's Bridge in core CUDA/C++ and develop bridge tests

valassi:
This PR implements the API6 step described in #323. It comes after and includes the API5 step in #324.

@roiser, this replaces my earlier WIP PR #315. I have taken most of the changes I had developped there, but I had to adapt them to the rest of the changes in the code. It includes Stefan's Bridge from PR #257 and integrates it organically in the rest of the core code. 

As discussed in #315, I have added a test in check_sa (which was now very easy after all the other changes), where I run curand and ramb on host, transpose C momenta to F momenta, and then launch the Bridge starting from F momenta as if they had been generated from MadEvent. This provides both a functionality test and a performance test. The functionality looks good. The performance is quite bad (but there are a few loose ends I need to improve in BridgeKernel), for the moment the transpose F2C (NOT the C2F, which is avaidable - the F2C, which is unavoidable in the Fortran bridge from MadEvent) seems to give a factor 20x overhead in cuda and a factor 2.5 overhead in C++/512y.  

This is still WIP (hopefully finish later this afternoon). Amongst the things to do:
- Reimplement Bridge in terms of the MatrixElement Kernel: this may sort out some loose ends in performance, as I should be able to avoid allocations and deallocations in every call (but still, there are only 12 calls for the 12 iterations in my test...).
- Backport to code generation, run the tests for ggttgg (I expect the performance hit will be much less...)

If a performance hit is confirmed, I will create a sperate ticket about improving performance for the transpose.


valassi:
This is now COMPLETE, no longer WIP. I would even say that the whole chain API1-API6 is ready to be merged.

I will try to present the status on Monday.

This has been a very interesting exercise. I think that having a full bridge test in checks_sa was VERY useful. Some lessons learnt
- Functionality wise, the bridge as is seems ok and ready to be integrated further with madevent fortran (modulo the fact that we'll need some extra parameters as discussed for single diagram enhancement)
- Performance wise, the test shows that the 'transpose' has a huge impact. But looking better, more than the 'transpose' itsefl, it is the memcpy of momenta from host to device that has a huge impact. In CUDA this is a factor 20 (!) in eemumu: it seems huge, but after all it was a factor 2 for MEs, and there are around sixteen times as much data to copy for momenta, so a factor 20 sounds reasonable. For C++ it was a factor 2.5 for eemumu - I improved the algorithm a bit (used constexpr where possible, and also changed the order of pipelining) and it went down to a factor 1.2 i.e. 20%. The important point however is that this overhead is significantly lower for ggtt, and is negligible for ggttgg (just ~2% for cuda, and invisible for C++). We are in the "lucky" situation where the ME calculation is just so much slower than the rest, as we knew,

Some points to do eventually, and comments/oibservations
- I just fixed Bridge.h to use T* instead of some hardcoded double*. Even if fortran will use double always, I kept everything as template. Eventually it would be nice to test mixing double (eg in rambo) and float (eg in ME), but this is not there yet.
- I changed 'make install' to add a few more headers and objects. Stefan you need to tell me if this is enough. The list of headers pulled in is large: if this is an issue, one simple option is to make Bridge non templated, but if it is templated then all the stuff it needs in he implementation mus be visible.
- In the check_sa test with --bridge, I hacked the "3a MECalcOnly" timer to actually include the bridge memcpy/transpose (but not that which I did elsewhere to convert C momenta to F momenta). This muct be done better. In any case, for the moment we can remember that MECalcOnly is the relevant metric (but it means different thinbgs in normal SA mode and in bridge SA mode).

Voila this concludes a first huge chunk of PRs. More will come next week.

valassi:
This was presented at the meeting today
https://indico.cern.ch/event/1103955/
I am merging this now

Issue #251, opened 2021-09-30
Understand scalability issues (crash from 12 to 120 iterations) in BMK container


valassi:
I thought that I had opened a ticket somehwre, either here or in BMK JIRA, but I had not.

I need to understand scalability issues (crash from 12 to 120 iterations) in BMK container. This is mentioned in https://its.cern.ch/jira/browse/BMK-734?focusedCommentId=3929331&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-3929331

<< More problematic, I saw that using 120 instead of 12 iterations made the CI nodes completely unusable (and it may be worse for systems being benchmarked!). I think that this is out of memory, but I should cross check. Two patches needed: in BMK, we need some memory monitor that kills the benchmark before the benchmark kills the machine (did we not have this already?); in madgraph4gpu, we need to reduce memory and understand why it grows with more iterations (ticket to be opened). Eventually anyway we need also multithreading in madgraph4gpu, beyond omp, so the question o fmemory will also come up again. >>

valassi:
I have now created https://its.cern.ch/jira/browse/BMK-813 to discuss the implications of this in the BMK project

valassi:
I have merged PR #326. This includes a patch that may well fix this issue. There used to be a matrixelementALL array that was growing with the number of iterations, so certainly it was not possible to increase the number of iterations indefinitely. Now the new code updates the statsistics at each iteration, without the need to store all MEs in a huge array first. To be tested.

valassi:
I have finally checked in https://its.cern.ch/jira/browse/BMK-813 that this issue is now fixed.

Indeed there used to be a memory leak (or at least a growing memory) that was rapidly make benchmarking (or production) impossible. This was a very old version of the code.

I also mentioned above that apparently I identified the precise issue, so this can be certainly closed.

valassi:
(For reference, this was in the standalone part - and what I checked so far in BMK-813 is also just the standalone part: madevent will come later on)

Issue #101, opened 2020-12-10
Cross Section calculation in the standalone app (low priority)


roiser:
A placeholder issue to remind us that we need cross section calculation for a minimal fully functioning workflow execution

valassi:
I have merged PR #326. This includes a crosssection kernel, that takes two ME and weight arrays as input and produces/updates an EventStatistics struct as output. As discussed there, some limitations
- It only uses a host kernel (as the old check_sa), eventually these could be done on the device too...
- It only does mean ME and mean WG (as the old check_sa), not a proper cross section
- Should include selection cuts at some point, see #100

roiser:
with madevent integration this can be closed

valassi:
I would keep this open at lowest priority. Not part of the madevent branch of development, but a useful addition to the standalone app IMO. MR #326 was one part of this (part of "API7" in the January 2022 workplan #323)

Issue #100, opened 2020-12-09
Selection cuts


valassi:
I do not remember if we have this already open.

@oliviermattelaer had suggested (and given code) for selection cuts to avoid infrared/collinear divergences in gg to ttgg processes.

This should be integrated in the software. Note that this may be interesting in the context of studying thread divergence.

valassi:
Just one extra comment. From the software point of view, if the filtering is done on momenta, before ME calculation, this may require the definition of a new component. 

Now we have
- random generation: minimal input (only seeds), output is random numbers
- sampling/rambo: input is random numbers, output is momenta
- MEs: input is momenta, output is MEs

What we probably need is
- random generation: minimal input (only seeds), output is random numbers
- sampling/rambo: input is random numbers, output is ALL momenta
- filtering: input is ALL momenta, output is FILTERED momenta
- MEs: input is FILTERED momenta, output is MEs

The filtering step probably requires some repacking, and then some rescheduling. Either we compute MEs for ALL momenta and just mask results at the end, but that's silly/inefficient. Or, if we want  to compute MEs only for FILTERED momenta, then we need to repack them into arrays with fewer elements, and reschedule. This is needed to optimise both GPU kernel launchers and CPU vectorization.

One challenge is to have variable sized arrays, with sizes not known in advance, and optimise all of this.

Another challenge is that we must probably split the kernel launching up to filtering, and later for MEs.

valassi:
Like #183 (massive rambo) this is needed for physics correctness of ggttgg (see also issue #146). This was mentioned by @oliviermattelaer in the June 21 meeting https://indico.cern.ch/event/1028452: without cuts, all cross sections are divergent, so we cannot compare physics results as we would be comparing infinities

Olivier, I think you had sent us some code for the cuts or am I wrong? Also, can you remind us which cuts are needed? Some cut on collinear gt (invariant mass? pT? angle?)...

Thanks
Andrea

PR #326, opened 2022-01-11
API(7a) CrossSection kernels

valassi:
This includes yet another kernel of teh sort discussed in #323. It is a CrossSectionKernel, that just prepares EventStatistics from ME arrays. It is now delegated stuff that was previously in check_sa.

Limitations
- It only uses a host kernel (as the old check_sa), eventually these could be done on the device too...
- It only does mean ME and mean WG (as the old check_sa), not a proper cross section

Note that this is likely to solve #251. There was clearly an arrray that was growing with etthe number of iterations in the old check_sa (the matrixelementALL array), this is now removed because the calculation of averages is done iteratively: they are computed on each iteratin and updated using the new data.

Note that this goes along the direction of issue #101 and indeed eventually can be used for a proper cross section (just need to combine WG and ME with the right physics formulas,,.,). One more missing howeer
- Should include selection cuts at some point #100 

Indeed, the EventStatistics struct reuses stuff I had prototyped (in gitlab) in the unw branch for #101.

It should be ready to be merged when the CI passes.

PS Last point I was forgetting: as in the old check_sa, but even more precisely, the calculations of means and stddevs is done using differences to a reference value, in order to improve the precision. An latrnbative would be to use quadriple precision stats in EventStatistics, but this other approach can be a nice learning ground.


valassi:
All checks passed, should be non controversial after yesterday's discussion. Self merging.

Issue #307, opened 2021-12-09
Simple custom complex class (cxsmpl)


valassi:
This is a placeholder for an idea I had in mind since quite some time. But unless I am wrong I have not opened a ticket for this yet.

The idea is to add yet another complex typedef, in addition to std, thrust and cucomplex. We can simply create a custom complex class that works for ALL backends (c++, cuda, and possibly alpaka, kokkos, sycl).

The point is that we only really need very basic operations in complex: addition, subtraction, multiplication, division. All of these can be very simply inlined in simple code, and actually some of these ALREADY are inlined in a header for the vector complex class. So essentially this would be an extension to the scalar type (RI) of the present cxtype_v class (RRRRIIII), it should be quite easy to do.

One additional benefit of this approach would be that we can define many methods as constexpr. This would be useful for instance in the definition of hardcoded parameters (issues #23 and  #39 and PR #306).

valassi:
Note: this is IMO a better alternative to trying to use thrust complex also in c++ (#147).

valassi:
This is done in #327, which I am about to merge. 

Note that I enable cxsmpl by default for c++ but keep thrust for cuda, as these are the fastest options.

The constexpr business for #39 can start off from here but needs more work.

Closing.

PR #327, opened 2022-01-11
Simple complex class (cxsmpl)

valassi:
This is a WIP PR for #307. It includes a fully functional simple complex class, alternative to thrust or std.

In cuda however this is around 15% slower than thrurst (maybe to be improved with constexpr or align or else?). In C++ it has the same performance. The interest is that eventually it may provide a constexpr version for hardcoded parameters (#39).

It is WIP (I need to clean this up and backport it). I will probably leave it disabled, but it can ceratonly be merged as the old default code is still there (and actually this PR includes a massive cleanup of headers, which may be useful - now mgOnGpuTypes is nicely split into mgOnGpuFptypes.h and mgOnGpuCxtypes.h, and in addition the ifdefs are tidied up).


valassi:
This is ready to be merged and I will self merge it soon.

In the end I enabled the new simple complex class by default for c++, as it gains 5-10% for float on all processes and it is essentially the same on double. On cuda I keep using thrust as this new class seems 10% slower.

valassi:
The tests have completed. Self merging.

Issue #155, opened 2021-04-06
Color algebra optimisations


valassi:
Hi @roiser @oliviermattelaer 
I open a generic issue about color algebra optimisations.

(Hi @ingvildh, this is what we discussed this morning and where I suggested that maybe the A100 tensor cores could be interesting)

This can be considred one issue in the epic of gg to ttgg optimisations, issue #146. I put some snippets of the code using a commit I used there, https://github.com/madgraph5/madgraph4gpu/commit/dd8711d0aa22c85429802b23de3169a68a97f298

This only exists for QCD processes, eg gg to tt(g)(g), not eemumu. It is one of the only places where calculating the "matrix element" (of the scattering matrix) actually involves matrix multiplications.

The more particles in the final state, the larger the "color matrix" involved.

For ggttgg, IIUC this is essentially this snippet here:
https://github.com/madgraph5/madgraph4gpu/blob/dd8711d0aa22c85429802b23de3169a68a97f298/epoch2/cuda/gg_ttgg/SubProcesses/P1_Sigma_sm_gg_ttxgg/gCPPProcess.cu#L1133
```
 const int ncolor = 24; 
...
// The color matrix;
  static const double denom[ncolor] = {54, 54, 54, 54, 54, 54, 54, 54, 54, 54,
      54, 54, 54, 54, 54, 54, 54, 54, 54, 54, 54, 54, 54, 54};
  static const double cf[ncolor][ncolor] = {{512, -64, -64, 8, 8, 80, -64, 8,
      8, -1, -1, -10, 8, -1, 80, -10, 71, 62, -1, -10, -10, 62, 62, -28}, {-64,
      512, 8, 80, -64, 8, 8, -64, -1, -10, 8, -1, -1, -10, -10, 62, 62, -28, 8,
      -1, 80, -10, 71, 62}, {-64, 8, 512, -64, 80, 8, 8, -1, 80, -10, 71, 62,
      -64, 8, 8, -1, -1, -10, -10, -1, 62, -28, -10, 62}, {8, 80, -64, 512, 8,
      -64, -1, -10, -10, 62, 62, -28, 8, -64, -1, -10, 8, -1, -1, 8, 71, 62,
      80, -10}, {8, -64, 80, 8, 512, -64, -1, 8, 71, 62, 80, -10, -10, -1, 62,
      -28, -10, 62, -64, 8, 8, -1, -1, -10}, {80, 8, 8, -64, -64, 512, -10, -1,
[...]

Issue #118, opened 2021-02-28
Monitor the use of tensor cores


valassi:
[Initial title - Are we making (can we make) any good use of tensor cores?]

While looking up FP32/FP64 core documentation for the paper, I realised that (e.g. in V100) we are probably not using the tensor cores. The FLOPs they give are huge.

Since these are mainly optimized for matrix multiplication, I was wondering fro instance if we could use them for the QCD color flow algebra.

Unfortunately however my impression is that these are of little use for anything that needs FP64 precision, as they seem to be optimised for FP16 actually. So, not sure what we can do with them, but we should investigate.

Note that this figure seems to describe excatly what we are doing for aggregating MC info (whether eit is the physics results like average MEs and cross sections, or even timing measurements): the underlying precision is FP16, then some calculations are done in FP32, but I assume the final precision is eventually converted back to FP16. What we are actually doing is using FP64, and trying to keep FP64 all the way (eg by accumulating differences to a mean, rather than individual elements directly), without needing FP128... we probably need 4x the precision.
https://developer.nvidia.com/blog/programming-tensor-cores-cuda-9/
![image](https://user-images.githubusercontent.com/3473550/109427582-4ebdaf80-79f3-11eb-9d63-25f53bbab4d8.png)

valassi:
Hi @ingvildh, as discussed this morning, this is one area where studying the A100 could be interesting, maybe especially for color algebra (issue #155).

It would be useful to understand better how the diagram above for V100 is modified for A100 however. From waht I had found and written above, it seems V100 tensor cores are limited to single precision, and this is probably a no-go for us.

This is a more recent link I found for programming tensor cores, but it does not specify if double precision is achievable https://developer.nvidia.com/blog/accelerating-matrix-multiplication-with-block-sparse-format-and-nvidia-tensor-cores/

valassi:
Hi @roiser, as discussed, this is one ticket where some tensor core stuff could be documented IMO. But for color algebra specific stuff I would suggest using #155. (Note that for A100 in general, nit just tensor cores, there is also #156 if needed).

roiser:
I believe the discussion on tensor core usage had shifted towards applying them for the color matrix multiplication, which we can do since the A100, as this is the first platform to provide double precision calculations on tensor cores. If I read this ticket well the main question it asks is whether tensor cores are useful for us as at the time of opening this was at max available at float precision. I propose to close this one and leave #155 open for future discussions.

[...]

Issue #282, opened 2021-10-27
Detailed analysis of performance regression from cuda 11.0 to cuda 11.4


valassi:
This is a spinoff of issue #277.

There are two big performance regressions in cuda ggttgg:
- one (with the same identical code) in moving from cuda11.0 to cuda 11.4 (both with gcc9.2)
- another (with different effects for cuda 11.0 and 11.4) in changing the code from epochX2 to epochX4

This issue is about analysing the first of these two effects. Produce some nsight plots, follow up, etc...

For the moment I will probably still use cuda 11.4 or even cuda 11.5 however.

valassi:
Generic instructions for building an dtesting the code, now updated for epochX/ggttgg, are in https://github.com/madgraph5/madgraph4gpu/wiki/Building-madgraph4gpu-and-measuring-throughput

A modified and simplified version (for cuda only) is below
```
  # Go to your chosen data directory and clean previous build products
  cd /data/valassi
  rm -rf madgraph4gpu

  # Optionally set up ccache (add cacche to PATH, set CCACHE_DIR, instruct madgraph4gpu to use it via USECCACHE)
  export PATH=/cvmfs/sft.cern.ch/lcg/releases/ccache/4.3-ed8d3/x86_64-centos7-gcc8-opt/bin:$PATH
  export CCACHE_DIR=$(pwd)/CCACHE_DIR
  export USECCACHE=1

  # Download the code
  git clone https://github.com/madgraph5/madgraph4gpu.git
  export MADGRAPH4GPU_HOME=$(pwd)/madgraph4gpu
  cd $MADGRAPH4GPU_HOME
  git checkout master
  git reset --hard 26d40755be840a55ef2e357392492546375ee34a

  # Set up gcc10.2 (set PATH, LD_LIBRARY_PATH, CXX)
  . /cvmfs/sft.cern.ch/lcg/releases/gcc/10.2.0-c44b3/x86_64-centos7/setup.sh

  # Set up cuda11.1 (set CUDA_HOME and PATH)
  export CUDA_HOME=/usr/local/cuda-11.1
  export PATH=${CUDA_HOME}/bin:${PATH}

  # Make a clean build of the ggttgg code (do not use a separate build directory)
  unset USEBUILDDIR
  cd $MADGRAPH4GPU_HOME/epochX/cudacpp/gg_ttgg/SubProcesses/P1_Sigma_sm_gg_ttxgg
  make cleanall
  make gcheck.exe

[...]

Issue #51, opened 2020-11-18
More precise assessment of RDC cost


valassi:
On July 25 during the hackathon @oliviermattelaer removed CUDA Relocatable Device Code, ie "-rdc=true"
https://github.com/madgraph5/madgraph4gpu/commit/bc03049a84968d94f814a91905a787a9aa995af3
We found at the time that this was giving us a significant performance boost. It would be useful however yo go back there and assess more precisely how much this gains (less than 1%, some 10%, a factor 10?...).
The drawback of that choice was that we must use a single file compilation, with no loaded libraries. This is not a problem for the moment, but as we get into more complicated workflows (eg with MadEvent and PDFs) this may change, so it is important to quantify the benefit in having no rdc (or the cost in having back rdc).

valassi:
As discussed now with @oliviermattelaer , probably at some point I got rid of FFVxxx functions from the headers, because it was faster. I mean, in cuda these must be in the same source anyway, so there is no need for a header. Then in the end I did not add it back for c++ either, as also in c++ these functions are now called within the same file where they are defined and only there (CPPProcess.cc). This may need to be reviewed, especially if it turns out that we can move away from single source and use RDC again

oliviermattelaer:
In the hackathon, the assesment was that for 
e+ e- > mu+ mu- the gain was HUGE.
The slide from the hackathon was mentionning a 13x factor (but a reduce occupance by a factor of 2 due to higher number of register).
The code passed from being memory bound to compute bound 

For g g > t t~ g g
the speed-up was a factor of two (1.9x):
In this case, we face issue that we did not have enough register available and therefore we were still memory bound.

oliviermattelaer:
[Hackaton_day2.pdf](https://github.com/madgraph5/madgraph4gpu/files/5567863/Hackaton_day2.pdf)

Here is the presentation of the hackaton with the associated details

valassi:
> 
> 
> [Hackaton_day2.pdf](https://github.com/madgraph5/madgraph4gpu/files/5567863/Hackaton_day2.pdf)
> 
> Here is the presentation of the hackaton with the associated details

Wow, great, thanks a lot! I did not remember we had all this material still from the hackathon ;-)

valassi:
Note the similarities of RDC in CUDA to LTO and inlining in C++ (issue #229 ... there factors 2-4 seem to be possible speedups)

valassi:
I have done a few recent tests for this in PR #332.

[...]

PR #328, opened 2022-01-12
API (9a) Templated FFV functions (memory access for wavefunctions/amplitudes) - move to cuda 11.6 (and test icx2021/2022 and gcc112)

valassi:
This is the WIP beginning of a PR for step9 of #323, namely adding memory buffers and memory access classes for wavefunctions and amplitudes. The motivation for this is allowing to split the cuda signakin kernel into many smaller kernels for the XXX and FFV functions (also being protyped in #242), and possibly a sperate kernel for color algebra (#155 and #118). 

Note that the motivation here is exclusively cuda/device, but I am defining an interface that applies also to c++. The idea is also to clean up and simplify the MemoryAccess classes, which work well but are not documented and may be an overkill (maybe easier to inline some boilerplate without going too much into variadic template overcomplications).

This is WIP. For the moment I have just
- added a proof of concept of using fptype* instead of cxtype_sv for wavefunctions in the XXX interfaces, and then using reinterpret casts
- I have also added the cxtype_ref class always, because this may be another way to move from RI pairs to complex types... it really depends wich memory structure we want in CUDA, but if we keep RI always contiguous, then a reinterpret cast may be a much simpler solution (and I can undo/revert the changes adding the cxtype_ref)

A lot of things to be done, including in random order, first step
- adding a second template parameter to the XXX functions, for the wavefunction access class
- add the wavefunction memory access class (initially just a reinterpret cast for a single event - this will always be the case for C++, well modulo the neppV vectorization handled by cxtype_v)
- use the memorybuffer class for wavefunctions
- do the same trick of moving from cxtype_sv to fptype also for the FFV functions
- similarly add a template parameter there for wavefunction memory access in FFV functions
- add memry buffers, memory access classes for amplitudes too
- integrate the new memory buffers/access for amplitudes in the FFV functions an dthe calling sequence
- add a second template parameter to the FFV funvtions for amplitude access

The first step above (quite long) should provide the same performance, with a single event access as is now. That is to say, the interfaces would change a lot, but the algorithm would remain the same and the performance should saty the same.

The second step then is
- move to multi-event buffers (possibly AOSOA) for wavefunctions and amplitudes too, only in cuda
- this will generate a lot more traffic with global memory, and it is likely to decrease performance if taken alone? in any case, do not yet attempt to split kernels

The third step is then
- split the kernels and check if it gets any better...


valassi:
Ok I have reached some temporary conclusion here: finally move to cuda 11.5!

- I have done some tests of RDC #51 (I will rebase on a merge of those tests). Indeed there is a penalty from rdc, even if only at 20% level or so (not the factors 10 we had seen long ago), RDC is no go, especially with a separate HelAmps
- I also did various tests with inlining some bits and not others, but this is no good as a solution
- out of lack of ideas, I tried to see if cuda 11.5 builds the code faster: now, it does BOTH build the code in a reasonable time, AND the runtime performance is faster! the performance regression #282 disappears... note however that this regression disappears ONLY if I also move to templated code in this apiwf PR. So essentially I must do the two together, move to templated methods AND also move to cuda 11.5

At this stage, all looks good: 
- I am able to move to templated FFV in this PR
- I am able to move to cuda 11,5 (I must do that actually)
- and out of these two things I even get a 2%+ performance increase (5.1 instead of 5.0 max tput on ggttgg cuda)

Now this PR needs to be completed with sevaral other points (even before moving to split kernels)

valassi:
Rebased after merging rdc PR #332

valassi:
I have done some tests with cuda11.6 which was released last week (see #334).
- It does not solve the regression with the code before this new PR with templated FFV (#282)
- However it behaves just like cuda 11.5 in this PR for templated FFVs: it has goo dbuild times (unlike cuda 11.1) and it has good runtime performance (unlike cuda11.5/6 before the move to templated FFVs)

I am therefore MOVING TO CUDA11.6 (not cuda11.5) as baseline in this PR #328.

valassi:
I am going to merge this PR even if the work it contains is logically incomplete.
- There are memory buffers for wavefunctions and amplitudes, but they are not used/tested (and a hack reinterpret cast is used in CPPProcess.cc between w_sv and w_fP), this is a bit halfway an old and a new way of doing these things
- The memory access classes by themselves are essentially dummy. And even the cxtype_ref has been always added when maybe it should not be there.
- All these things will become clearer only moving twoards a startegy for splitting kernels.

The above was about the limitations. What this PR does contain however is
- The move of FFV functions to templated FFV functions, which is required for memory access in split kernels. (By itself, using the old cuda11.1, this would explode build times).
- The move to cuda11.6. This brings back build times of templated functions to a reasonable level. (And it actually needs the templated functions, as with the older code it would otherwise have a performance regression #282).
- Combining cuda11.6 with templated functions, cuda runtime is now faster (~5.2 instead of ~5.0 for ggttgg).
- A few fixes for icx2021/clang13 and icx2022/clang14. Performance tests for icx2021, icx2022 and also gcc112.

---

With respect to the stuff I mentioned at the top, some todo and next steps are below

To clean up
- Clean up the business of fptype* s cxtype_sv* for wavefunctions and amplitudes in FFVs. Maybe hide again the cxtype_ref class if not needed.
- Clean up how memory buffers are assigned for wavefunctions and amplitudes in CPPProcess. The clean way will probably be to move this OUTSIDE calculate_wavefunction, and into the kernel launcher class calling calculate_wavefunction. This demands quite some structural changes, which are in any case needed for splitting kernels. In any case those reinterpret casts between w_sv and w_fp should disappear.

A new "first" step would then be
- With the buffers oustide calculate wavefunctions, do not change the structure to global memory AOSOAs. Each buffer would still be a single event. You should get teh same performance as now.

The second step (only as a test case to study, clearly not a production version) would remain
- move to multi-event buffers (possibly AOSOA) for wavefunctions and amplitudes too, only in cuda
- this will generate a lot more traffic with global memory, and it is likely to decrease performance if taken alone? in any case, do not yet attempt to split kernels

The third step (probably to be done together with the second) would be
- split the kernels and check if it gets any better...
- I would actually start by putting jamps to global memory, rather than amps... and maybe I would start by just splitting sigmakin into two kernels, one for XXX+FFVs, the other for the color matrix algenra

---

I am also mertging this now because there are a couple of orthogonal issues which are higher priority
- Olivier's #337 which requires generating pptt rather thna ggtt
- Some possible (almost certain) bugs in SIMD, due to some initializer lists
- Maybe also investigate why icx2022 gives different physics values from icx2021 and other compilers../ another bug?

All checks have now succeded. I will merge

PR #332, opened 2022-01-17
Throughput tests for RDC and for CUDA 11.5

valassi:
This is a zero-net-change PR with some tests for RDC and CUDA 11.5. I just want some commits to be in the history for the doc.

This was started as a helper to PR #328, in the middle of it. It includes
- Tests showing that "-rdc true" (and even more having a separate HelAmps.cc file) has a cost, though only 20% or so, not the factor 10 that had been observed in #51 
- Tests showing that, BEFORE THE APIWF PR #328, the performance regression from cuda 11.5 was still there in ggttgg, ie tput was 30% slower in cuda 11.5 than in 11.1, as reported in #282
- Also a few tests of inline vs noinline, related to the LTO studis in #231


valassi:
Self merging - all files are identical to the previous master

Issue #334, opened 2022-01-18
Evaluate the move from cuda 11.5 to 11.6


valassi:
Cuda 11.6 was released yesterday.

By coincidence, I had just decided to move back from cuda 11.1 to cuda 11.5 yesterday, as the move to templated FFV functions in the code (#328) allows and requires that (see #282).

I should try out cuda 11.6, both:
- on the old code before templated FFV #328, just to see if the regression had been fixed in the compiler, still on the old code
- on the new code with templated FFVs, to see if it is still viable to use cuda 11.6 (or if there is a regression, or even a speedup! and to see if build times are ok)

valassi:
I have installed cuda116.

The first test for #282 is done in PR #335
- the same performance regression from cuda 11.1 to cuda 11.4/5 is observed with cuda 11.6 on the old (pre-template-FFV) code, ie cuda 11.6 does not solve that regression, see https://github.com/madgraph5/madgraph4gpu/pull/335/commits/94dee1f59d601cf1231bf3335b6650e9a81302a4

valassi:
I have also tested the second part in #328
- cuda 11.6 behaves just like cuda 11.5 in this PR for templated FFVs: it has good build times (unlike cuda 11.1) and it has good runtime performance (unlike cuda11.5/6 before the move to templated FFVs)

I am therefore MOVING TO CUDA11.6 (not cuda 11.5) as baseline in the context of PR #328.

This issue can be closed.

PR #335, opened 2022-01-18
Throughput test for CUDA 11.6

valassi:
Just a simple test for #282 and #334: 
- build/run present master (BEFORE apiwf #328) with cuda 11.6: I get the same 30% penalty as with cuda 11.5 (i.e. the performance regression is not fixed in cuda 11.5)


valassi:
This is a zero-net-change PR, I will self-merge (I want the log in the history as doc)

Issue #337, opened 2022-01-20
wrong assert in cudacpp plugin (and test code generation for pptt)


oliviermattelaer:
hi,

I do have the following crash
```
  File "/Users/omattelaer/Documents/workspace/convert_to_git/git-mg5amcnlo/PLUGIN/CUDACPP_SA_OUTPUT/model_handling.py"\
, line 1187, in get_matrix_element_calls
    res = self.super_get_matrix_element_calls(matrix_element, color_amplitudes)
  File "/Users/omattelaer/Documents/workspace/convert_to_git/git-mg5amcnlo/PLUGIN/CUDACPP_SA_OUTPUT/model_handling.py"\
, line 1176, in super_get_matrix_element_calls
    scoeff = PLUGIN_OneProcessExporter.coeff(*coeff) # AV
  File "/Users/omattelaer/Documents/workspace/convert_to_git/git-mg5amcnlo/PLUGIN/CUDACPP_SA_OUTPUT/model_handling.py"\
, line 992, in coeff
    assert(False)
AssertionError
```
when doing (from  master)
```
generate p p > t t~; output standalone_cudacpp mytest
```

Is there a real reason for this assert?

valassi:
Hi Olivier, interesting, it looks clearly wrong. I will have a look. I guess I need to generate a test pptt too as otherwise I will not see that? I am surprised I do not see the issues in the three processes I use routinely

valassi:
I rename this to also mention "and test code generation for pptt" - the reason why I never had the issue reported by Olivier is that I never tried pptt, I gues?

valassi:
Thanks Olivier, this was easily confirmed. I will work on PR #340 about this.

I just realise I have no idea what the pptt code should look like. If I take the default madgraph without my plugin (say 270gou which is what I am based on for now), that was working for you initially and was producing code that made sense?

Thanks Andrea

valassi:
Hi @oliviermattelaer 

this specific issue is fixed easily: I just disabled the "assert(False)". I think that I had added that to understand if/when thi smethod "coeff" was used at all. I found that it is not used in eemumu or ggtt, but it it is used in uutt (or in the uutt within pptt).
https://github.com/madgraph5/madgraph4gpu/pull/340/commits/2876844936dd0ad7f1db892b1296d77739e5cffc

[...]

Issue #272, opened 2021-10-21
Add an example of a calculation with nprocesses>1


valassi:
Just a reminder. In the epochX work #244, I kept several of the comments I had all over the code "FIXME this assumes nprocesses=1". I am not sure what the code should look like or how it should behave with nprocesses>1.

Eventually we need to add a test of one simple such example. This may even be a very simple process, as long as it has nprocesses>1.

@oliviermattelaer maybe this is for you to suggest one ;-)

Low priority anyway (Fortran integration is more important)

valassi:
Adding pptt is definitely needed - Olivier's bug #337 probably comes from this

oliviermattelaer:
So in the latest gpucpp branch.
All P directories would have exactly one matrix-element function.
This matrix-element can still be related to multiple initial/final state but all of them will have the exact same matrix-element.

(Compare to my original email, this is now done in a "decent" way meaning that identical matrix-element are correctly handle)

valassi:
Thanks Olivier! I updated to your latest upstream code in PR #619.

This is related to the discussion we had at the meeting earlier this wek https://indico.cern.ch/event/1263518/
You showed that for instance this process changes as follows
![image](https://user-images.githubusercontent.com/3473550/229080358-ae23dccc-dcca-4b16-a9c4-c7cc4430bf65.png)

I could close this as fixed by PR #619. But I will keep it open for now, maybe I will clean up the code, remove a lot of nprocesses and/or add a few cleaner comments mentioning this #272 and PR #619

valassi:
In the end I have done a lot more analysis about this issue. I did not want to simply merge MR #619 and close this one. I wanted to make sure that before the MR we had a problem with nprocesses>1, and after the MR we did not.

This took several iterations and extra issues opened on the way
- I wanted to reproduce Olivier's screesnhot above to see what has changed.
- I understood the above probably came from work done with Nathan on https://github.com/nscottnichols/madgraph4gpu/blob/f9689784390e2975a055711b3a8d79f7b7688860/epochX/sycl/CODEGEN/generateAndCompare2.sh#L58 on a SUSY example, so I tried several SUSY examples, initially gq_ttllq. On the way I opened #624 #625 #627 #628
- Then I actually realised I could probably just test a SM process like gq_ttq (but I also did gq_ttllq for SM), #626 (which also gave some of the issues above)
- Then I also tried in parallel to produce the plot with "launch", and this gave me #629 

[...]

PR #340, opened 2022-01-20
Fix assert in pptt code generation

valassi:
This is a WIP for pptt code generation (#272). I immediately hit the bug reported by @oliviermattelaer in #337.


valassi:
This is a fix for the first issue in #337 (the assert error during pptt code generation).

The second issue in #337 (the build error of pptt code due to nprocesses=2) is now moved to #343.

I am self merging this patch, which only contains minor changes.

PR #345, opened 2022-01-24
WIP - Create formatted table from log parser + A few more fixes for clang/isnan

valassi:
This is a WIP PR for creating a formatted table from log parser.

On the way, I decided to rerun all three processes both in double and float, both with and wthout inlining. In this process I came across a few more build warnings and other issues that I also fixed.

Most notably, I changed the handling of fast math checks for isnan. I realised that with clang my worakround to disable fast math within the file would probably not work. So instead I now disable fast math in the Makefile for the CrossSectionKernels.cc, shidh is the only source code containing those checks.


valassi:
All tests pass - self merging

Issue #349, opened 2022-01-25
uudd generation: cIPC[0] should be excluded?


valassi:
Hi @oliviermattelaer I see an issue in uudd generation: there is a cIPC[0] created, which obviously has no initializer. This variable (and everything associated with it) should not be there at all.

Similar to #343 about nprocesses=1, is this a problem in my plugin or in the base code? I will try a fix anyway in my plugin for the moment.

This is a followup to #344. (Olivier suggesting trying out the generation of uudd)

valassi:
Ok this will be fixed by PR #350, which I am about to merge.

(I am instead preparing a separate PR for heft, which has other issues).

Closing this one as fixed. It's a bit of a hack, but it's good enough I think: just remove all lines containing IPD if cIPD has 0 length.

PR #350, opened 2022-01-25
Fixes for uudd code generation

valassi:
This is a fix for #349 (followup to #344).

I have fixed (hacked...) code generation to remove all lines containing IPD if the size of cIPD would be 0. The generated code now builds and runs ok.

I have also added a reference test file for uudd


valassi:
Ok this is now complete. The heft generation has different issues and I am creating a separate PR.

I will self merge this one as soon as the tests are ok

Issue #351, opened 2022-01-25
Code generation for EFT gg>h : issues with model parameters


valassi:
Hi @oliviermattelaer this is the second part of the followup of #344.

You suggested to try EFT gg>h (nice idea!). Indeed I had (have) a few minor issues to fix, but now I am also facing a more fundamental issue: I have
```
      // Wavefunction(s) for diagram number 1
      vxxxxx<M_ACCESS, W_ACCESS>( momenta, 0., cHel[ihel][0], -1, w_fp[0], 0 );

      vxxxxx<M_ACCESS, W_ACCESS>( momenta, 0., cHel[ihel][1], -1, w_fp[1], 1 );

      sxxxxx<M_ACCESS, W_ACCESS>( momenta, m_pars->mdl_MH, +1, w_fp[2], 2 );


      // Amplitude(s) for diagram number 1
      VVS3_0<W_ACCESS, A_ACCESS>( w_fp[0], w_fp[1], w_fp[2], cxmake( cIPC[0], cIPC[1] ), &amp_fp[0] );
      jamp_sv[0] += +2.*amp_sv[0];
```

The problem here is that the `m_pars->mdl_MH` should possibly be in constant memory rather than in this instance data memeber. At least, for the SM Z/W masses this is what we have, they end up in the cIPD block I think.

My question: is there something fundamentally different in your base code generation for EFT, which makes it such that the Higgs mass does not end up in cIPD? I will wait for ou to answer rather than attempt a fix on this one.

(Just FYI if you have lost it in my changes: I have also been prototyping a hardcoded constexpr version of all physics parameters, PR #306. This is even faster than constant memory. It is a related, but DIFFERENT story, just teling you for info. In any case we first need to fix the issue above).

Thanks! Andrea

valassi:
Ehm... I tried to generate the SM gg>h, but I get an error, of course because there is no such diagram in the SM...
```
DEBUG:  True [misc.py at line 2192] 
INFO: Checking for minimal orders which gives processes. 
INFO: Please specify coupling orders to bypass this step. 
INFO: Trying process: g g > h WEIGHTED<=2 @1  
Command "import /data/avalassi/GPU2020/MG5aMC/2.7.0_gpu/CODEGEN_cudacpp_gg_h.mg" interrupted in sub-command:
"define vl~ = ve~ vm~ vt~" with error:
[...]

PR #352, opened 2022-01-25
Fixes for EFT gg>h code generation

valassi:
This is a WIP PR for issue #351 (follow up of #344).

It contains a few fixes and tests for EFT gg to h code generation, but there are still a few fundamental issues to understand (see #344, should the Higgs mass go to cIPD?)


valassi:
Ok all issues have been clarified
- model parameters #351 has been fixed by removing mass and helicity (but #356 was also providing another fix if necessary)
- note that rambo 2>1 gives a warning #358 but this makes sense, so I am kjust keeping the warning

All checks have passed (I have then just submitted some log changes). I will self merge

valassi:
(One minor point, I was unsure wehether we should have one testxxx in Subprocesses, or one identical version in each PSigma... I did the former, after I saw that mgOnGpuConfig follows the same pattern, and I checked in pptt that it is only generated once. I guess this only works for files which are identical for all PSigmas - mgOnGpuConfig apparently is one such file - so nothing to worry about)

valassi:
Ok all checks completed, self merging

Issue #96, opened 2020-12-08
Check for minimum gcc and cuda versions


roiser:
Currently we rely on a minimum of Cuda 10.2 and gcc 8. The gcc version should always and the Cuda version only in case of CUDA_HOME be checked. 

Fixes #92

valassi:
Note that because of the issues discussed in #277, #281, #282, I have downgraded back from cuda 11.4 to cuda 11.1 as baseline. Unlike cuda 11.0, this allows using gcc10, but I am using now gcc10.2 (as my cvmfs 10.3 has some issues). I am using c++17 in gcc and c++14 in cuda.

valassi:
I am moving cuda from c++14 to c++17 (#333). I am adding a check for a minimum cuda version 11.2 (minimum with c++17 support) in mgOnGpuConfig.h.

valassi:
I have also added a minimal clang check (need clang>=11, simply because I no longer test clang10).

And as a temporary patch I have disable the check on gcc if it is used below clang, so that the CI can use clang13 with a gcc toolchain <9.3 (#355)

valassi:
I keep this open, as we can probably keep updating this as we move to more recent compilers. 

Otherwise, two oustanding missing itemms
- A check for icx is also missing.
- And an update to gcc10, ALSO in the toolchain?

PR #354, opened 2022-01-26
Move CUDA builds from C++14 to C++17

### LINKED ISSUE #333: Move nvcc builds back from c++14 to c++17
Until today I had downgraded nvcc builds to c++14 (even if gccc builds use c++17), because I was using cuda 11.1 which does not support c++17. 

I was using cuda 11.1 becuase of the performance regression (ggttgg slower by 20%) described in #282.

With the move to templated FFVs described in #328, a move back from cuda 11.1 to cuda 11.5 is possible (no performance regression, the code is actually faster!) and necessary (as otherwise the builds take too long). 

I am therefore moving back to cuda 11.5. This supports c++17, so this move will allow reenabling c++17 builds in nvcc. I will do this AFTER merging #328.

The move from c++14 to c++17 will allow cleaning up the code in various places (eg in the MemoryAccessAmplitude variadic templates, and more).

valassi:
This is the PR to fix #333.

Still WIP for now but will complete and merge soon

### DIFF
--- a/epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/madgraph/iolibs/template_files/gpu/Bridge.h
+++ b/epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/madgraph/iolibs/template_files/gpu/Bridge.h
@@ -174,7 +174,7 @@ template <typename T>
 void Bridge<T>::gpu_sequence( const T *momenta, T *mes, const bool goodHelOnly )
 {
   constexpr int neppM = MemoryAccessMomenta::neppM;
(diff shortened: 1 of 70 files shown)


valassi:
This is now ready to be merged, it includes
- moving cuda from c++14 to c++17 (close #333) 
- add some checks on minimal cuda11.2 and gcc9.3 (#96) 
- also a few fixes related to OMP coming from the nvccsimd branch (#319), but note OMP is kept disabled

valassi:
I have added a (TEMPORARY?) patch  to allow gcc<9.3 when it is used as toolchain for clang, else the CI was failing (#355)

valassi:
All checks pass, I am self merging

Issue #353, opened 2022-01-25
Improve optional handling of hardcoided parameters


valassi:
This is a followup to PR #306. I had prototyped the hardcoding of physics parameters. at least for the three processes I normally use. 

This was okish, but to make it more general I needed a simpler complex class with constexpr arithmetics. This is in principle done (PR #327 for issue #307), though it must be completed with constexpr methods, probably.

Note that the handling of physics parameters (from m_pars, or from cIPC/cIPD constant memory, or from hardcoded constexpr) is also relevant to non-SM models, like the EFT gg>h test of #351.

valassi:
I am prototyping this in PR #359, all looks good so far - but I will pass on to other things first.

PR #359, opened 2022-01-26
Improved handling of hardcoded parameters using the cxsmpl class

valassi:
This is a WIP PR implementing #353 (improved handling of hardcoded parameters using the cxsmpl class).

The rationale is that hardcoding parameters gets a significant runtime boost for cuda and c++.

This is a followup to PR #306. There I had done a first proof of concept, but the full automatic generation of hardcoded parameters was not possible, because some complex parameters are computed in the header, and this requires constexpr complex arithmetics which is not there in std::complex.

This is now possible because
- the simpler cxsmpl class has been introduced (#307 and #327), and I have just now made constexpr its arithmetic operators
- constexpr arithmetics requires c++17, which I added this morning also to CUDA (#333 and #354)
- this in turn was possible thanks to the upgrade from cuda11.1 to 11.6 (#334), which came across during a move to templated FFV functions (#328), which provided a worakround to the cuda peformance regression observed in previous code (#282)

There is now a full proof of concept for all Parameters ardcoded in Parameters.h. This is WIP because I still need to move this to code generation (and try it out on different models, even including EFT for instance).


valassi:
All tests pass, I am self merging

PR #361, opened 2022-01-28
create shared libraries

roiser:
create shared libraries so we can link them for the full workflow.


valassi:
Hi @roiser I created #367, where I fixed the conflict. I will add more stuff there.

Issue #339, opened 2022-01-20
Probable bug in initializer list use in SIMD inline functions


valassi:
I realised there are probably one or more bugs in the way I used initializer lists for SIMD compiler vector extensions, eg
https://github.com/madgraph5/madgraph4gpu/blob/5ac285cebc04cd7eb4d6679c6a3a5e2645803d30/epochX/cudacpp/ee_mumu/src/mgOnGpuVectors.h#L220

Essentially I have coded assuming that "fptype_v{x}" would initialize all elements of the vector to x, but actually only the first element is x and the others are zero.

This needs to be debugged. It is also the reason why I added a new testmisc test infrastructure in #336

valassi:
Maybe the cause of #338, maybe not

valassi:
This is now fixed by PR #364, which I just merged. I am closing this issue.

NOTE: this patch fixes some bugs, but the bugs were actually harmless so far. The bug is essentially that only the first element of an array was initialized to a given value, with the subsequent ones initialized to 0. However, this function was essentially only used with the given value equal to 0, so eveything was correctly initialized to 0 anyway. Waht I mean is that no change in average ME was observed or is expected to be observed.

roiser:
Hi @valassi , thanks for pointing me to this closed issue, as we were discussing there are a few places in the code where eg. a SIMD type is initialised with ```{0}```, this is not a problem when the initialising value is 0 as by default all values of the SIMD array will be 0, just the code above sets only the first array value. 

With my current work I stumbled over the case where I needed to initialise all entries of a SIMD type with non 0 and then the above will not work as only the first entry will be the initialised value and all other set to 0. The proposed solution in this case is to use ```std::fill_n``` e.g. 

```std::fill_n(&mysimd[0], nelems, value);```

the disadvantage here is that it can't be constexpr anymore. So maybe one should only use it for non 0 init values and for 0 just do an empty ```{}``` but the latter is a minor issue as also ```{0}``` is doing the right thing anyway. Its maybe just a bit confusing.

**Comment by valassi** [comment]
Hi @roiser, this old issue was fixed with this simple approach in PR #364
```
inline
cxtype_v cxmake( const fptype& r, const fptype_v& i )
{
  //return cxtype_v( fptype_v{r}, i ); // THIS WAS A BUG! #339
  return cxtype_v( fptype_v{0} + r, i ); // IIII=0000+r=rrrr
}
```
Check the changes in that PR there are also many other tests and examples

[...]

PR #364, opened 2022-02-01
Bug fixes and comments in SIMD initializer lists

valassi:
This is a patch to fix bug #339. It also includes a few changes from '{}' to '()' in initializers to make them more understandable, a few comments about some initializers, and some tests. It concerns both vectors in compiler vector extensions, and other arrays (or arrays of arrays, or arrays of vectors etc).


valassi:
All checks succeed - self merging.

NOTE: this patch fixes some bugs, but the bugs were actually harmless so far. The bug is essentially that only the first element of an array was initialized to a given value, with the subsequent ones initialized to 0. However, this function was essentially only used with the given value equal to 0, so eveything was correctly initialized to 0 anyway. Waht I mean is that no change in average ME was observed or is expected to be observed.

Issue #363, opened 2022-01-28
Test bbar to ttbar


valassi:
As discussed with @oliviermattelaer : it would be useful to test bbar to ttbar to check the cIPC handling. This process should depend on the b mass, which in principle should be handled in the same way as the Higgs mass is handled. The higgs mass was problematic in EFT gg>h (seee #351)

valassi:
This is done in PR #365, which I am about to merge. Closing.

Issue #358, opened 2022-01-26
RAMBO with a single final particle? (Do we need sampling in a 2->1 process?)


valassi:
Hi @oliviermattelaer , next physics question :-)

For the EFT gg>h process of #344: after fixing other generation/build issues I am now getting this warning
```
ccache /cvmfs/sft.cern.ch/lcg/releases/gcc/10.2.0-c44b3/x86_64-centos7/bin/g++  -O3  -std=c++17 -I. -I../../src -I../../../../../tools -DUSE_NVTX -Wall -Wshadow -Wextra -ffast-math   -march=skylake-avx512 -mprefer-vector-width=256  -DMGONGPU_FPTYPE_DOUBLE -I/usr/local/cuda-11.6/include/ -c RamboSamplingKernels.cc -o RamboSamplingKernels.o
In file included from RamboSamplingKernels.cc:8:
../../src/rambo.h: In function ‘void mg5amcCpu::ramboGetMomentaFinal(mgOnGpu::fptype, const fptype*, mgOnGpu::fptype*, mgOnGpu::fptype*) [with R_ACCESS = KernelAccessRandomNumbers<false>; M_ACCESS = KernelAccessMomenta<false>; W_ACCESS = KernelAccessWeights<false>]’:
../../src/rambo.h:121:60: warning: ‘z[0]’ is used uninitialized in this function [-Wuninitialized]
  121 |     if ( nparf != 2 ) wt = (2. * nparf - 4.) * log(energy) + z[nparf-1];
      |                            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~
```

I looked at the code, and indeed z[0] is never initialized. Initially I was concerned this might be a bug in the translation from Fortran to C indexes, and maybe that I had introduced this in changing the interfaces. But I realised that
- the algorithm is the same, with z[0] uninitialised, in the launchpad cc version https://bazaar.launchpad.net/~madteam/mg5amcnlo/3.x/view/head:/madgraph/iolibs/template_files/rambo.cc#L178
- the algorithm is the same, with Z(1) uninitialised, in the launchpad Fortran version https://bazaar.launchpad.net/~madteam/mg5amcnlo/3.x/view/head:/vendor/CutTools/examples/rambo.f#L36

Eventually, I realised that: maybe we do not need RAMBO (or any other phase space sampling) at all, if we have a 2->1 process?

[...]

PR #365, opened 2022-02-01
Test bbtt + Add minor fixes for hEFT

valassi:
This is a patch where I tested that all is ok out of the box for bbar to ttbar (#363).
I just had to add a reference file.

I realised I had forgotten to add a reference file also for EFT gg to h (#351) so I did that.
I noticed that this contained mostly NaN final momenta, because of the rambo 2-to-1 issue #358.
So I added a very simple patch (and a WARNING) for rambo 2-to-1, too.


valassi:
All tests succeeded - self merging

PR #366, opened 2022-02-01
Improved fixes for rambo 2-to-1

valassi:
This is a followup to PR #365 - I realised there was a better way to get the four momentum o fth efinal particle, just add those of the initial particles, this is more generic, the previous one had buggy physics


valassi:
This represents a better fix for #358. I will self merge

Issue #362, opened 2022-01-28
Cleanup of Makefiles


roiser:
The Makefiles are getting a bit messy. In order to keep them maintainable I suggest to clean them up at some stage.

valassi:
Agreed! :-)

Already tried to do a few things (eg remove compile time flags from link commands), but it's a long process. Please add some of your desiderata if any ;-)

Some of my todo
- I think I would remove these gXXX.cu links to XXX.cc, and instead use consistently the "nvcc -x cu" approach: it is already there in the Makefile for some files, but not for all
- to clarify better (eventually): the AVX business and multi-SIMD fat binarues #177 

Also related to cmake #271 of course... but I would rather start by keeping Makefiles and cleaning them up, then by moving to cmake

valassi:
I am about to commit PR #367 which contains a significant cleanup of Makefiles - although of course more can be done...

Issue #356, opened 2022-01-26
Clarify roles of CPPProcess and MatrixElementKernels


valassi:
This is an important "internal API" issue in the line of those discussed in #323.

I would like to understand how to restructure the collaboration between the code in CPPProcess and MatrixElementKernels. Should these actually be merged? If possible, maybe I would avoid that, as there is some boilerplate-like stuff that can go in MatrixElementKernels independent of the physics process.

@oliviermattelaer, @roiser, copying you explicitly on this one if you have feedback....

The CPPProcess class could probably be renamed as "PhysicsProcess", to give "process" a different meaning: not a CPU process, but a physics process. In fact, CPPProcess.h/cc are now the files where most of the physicsprocess-dependent code resides (essentially, how to assemble the various FFV/XXX to make a matrix element calculation, and also the reading of parameters, and also the color algebra).

CURRENTLY, note that this is quite peculiar:
- check_sa creates a CPPProcess instance (one in the app, but not a singleton pattern with static methods)
- this CPPProcess actually reads the physicvs process parameters from file, and then copies them to C++ static memory or to CUDA global memory (unless they are hardcoded, see PR #306): one problem however is that this is not yet done for EFTs (#351)
- from this point on, essentially no one needs the CPPProcess instance at all: the physics parameters are in static/constant (essentially the parameters are singletons, which makes sense), and it is from there that calculate_wavefunctions itself reads it, to put it in FFVs
- it is kind of an accident that calculate_wavefunctions is also in the CPPProcess FILE, in the sense that in any case it does not read member functions of CPPProcess: it actually is not a memeber function of the CPPProcess class, it is a standalone function (as if it was a static function of the CPPProcess class)
- similarly, claculate_wavefunction is only used by the sigmakin functions, which are themselves also standalone c++ functions, or standalone CUDA kernels, not belonging to a class

I would say there are two/three things that can/must be fixed/clarified
- whether cpprprocess needs to be a singleton or a class with many instances (I guess the former)
- the handling of parameters, also to fix EFTs (one way could be to make parameters available from the cppprocess singleton, so that in cases like EFT where they have not - YET - been copied to constant memory, there is an easy way to retrieve them)
[...]

Issue #329, opened 2022-01-13
Keep Bridge state across several calls?


valassi:
Hi @roiser, @oliviermattelaer 

After our chat yesterday at some point I realised there may be one issue with the current Bridge interface: we keep creating a new object
https://github.com/madgraph5/madgraph4gpu/blob/774a90088168e96bf37086fff6ce478c19861423/epoch2/fortran-cuda/pp_ttx/SubProcesses/fbridge.cu#L30
https://github.com/madgraph5/madgraph4gpu/blob/774a90088168e96bf37086fff6ce478c19861423/epoch2/fortran-cuda/pp_ttx/SubProcesses/bridge.h#L115

This would not leak memory because it is stack allocated, however it has two issues
- Most importantly, it would calculate good helicities over and over, at each call
- I also have the impression that it would read the parameter file every time
- In addition, it is a bit silly to create a Bridge and delete it at every iteration of matrix element calculations

Since it is a bad idea to keep static instances in the cuda/c++ module, I think that it is somehow unavoidable to pass the knowledge of the Bridge C++ pointer to Fortran. I made a couple of quick tests and this looks quite easy. 

For instance, look at this code

File Bridge.h
```cpp
#ifndef BRIDGE_H 
#define BRIDGE_H 1
#include <iostream>
class Bridge {
public:
  Bridge() : m_init( false ){ std::cout << "Creating Bridge " << this << std::endl; }
  virtual ~Bridge(){ std::cout << "Deleting Bridge " << this << std::endl; }  
  void doStuff( const double x ){
    if ( ! m_init ){ std::cout << "Initialising" << std::endl; m_init = true; }
    std::cout << "Doing stuff " << x << std::endl;
  }  
private:
  bool m_init;
};
#endif // BRIDGE_H
```

File prog.f
```fortran
c See https://stackoverflow.com/a/24762346
c (See also https://docs.oracle.com/cd/E19059-01/stud.9/817-6694/11_cfort.html)
c (See also https://stackoverflow.com/a/29154061)
      PROGRAM FPROG
      IMPLICIT NONE
      INTEGER*8 BRIDGE
      INTEGER ILOOP, NLOOP
      DOUBLE PRECISION XLOOP
      CALL FCREATEBRIDGE( BRIDGE )
      NLOOP = 5
      DO ILOOP = 1, NLOOP
        XLOOP = 1.1 * ILOOP
c       WRITE(*,*) "XLOOP =", XLOOP
[...]

Issue #212, opened 2021-06-14
Single precision average ME is not the same for CUDA and C++ in single-precision (ggttgg and eemumu)


valassi:
As discussed in PR #211, single precision average ME is not the same for CUDA and C++ in single-precision ggttgg

See for instance https://github.com/valassi/madgraph4gpu/commit/a75ee3b6ba38d0be49f294c241a5e8b0682c84df#diff-45e40fdc2f6b7c71419c9f5e7e36267d7951e21c32488d6ecf35de3ec28ced57
```
perf stat -d ../../../../../epoch2/cuda/gg_ttgg/SubProcesses/P1_Sigma_sm_gg_ttxgg/gcheck.exe -p 64 256 1 |& egrep '(Process|fptype_sv|OMP threads|EvtsPerSec\[MECalc|MeanMatrix|FP precision|TOTAL       :|EvtsPerSec\[Matrix|CUCOMPLEX|COMMON RANDOM|ERROR|instructions|cycles|elapsed)' | grep -v 'Performance counter stats'
FP precision               = FLOAT (nan=0)
EvtsPerSec[MatrixElems] (3)= ( 6.610975e+05                 )  sec^-1
MeanMatrixElemValue        = ( 4.059594e+00 +- 2.368052e+00 )  GeV^-4
TOTAL       :     5.920932 sec
    15,536,792,487      cycles                    #    2.654 GHz
    28,689,538,755      instructions              #    1.85  insn per cycle
       6.207201648 seconds time elapsed

perf stat -d ../../../../../epoch2/cuda/gg_ttgg/SubProcesses/P1_Sigma_sm_gg_ttxgg/check.exe -p 64 256 1 |& egrep '(Process|fptype_sv|OMP threads|EvtsPerSec\[MECalc|MeanMatrix|FP precision|TOTAL       :|EvtsPerSec\[Matrix|CUCOMPLEX|COMMON RANDOM|ERROR|instructions|cycles|elapsed)' | grep -v 'Performance counter stats'
FP precision               = FLOAT (nan=0)
EvtsPerSec[MatrixElems] (3)= ( 1.786471e+03                 )  sec^-1
MeanMatrixElemValue        = ( 4.060118e+00 +- 2.367901e+00 )  GeV^-4
TOTAL       :     9.183867 sec
    24,604,689,155      cycles                    #    2.677 GHz
    73,872,471,813      instructions              #    3.00  insn per cycle
       9.193035302 seconds time elapsed
```

[...]

PR #367, opened 2022-02-01
Shared libraries + Bridge + Cleaner Makefiles

valassi:
This is a WIP PR to follow up on
- shared libraries (replacing PR #361 ?)
- cleaning up Makefiles a bit (#362)
- complete Bridge (#329, #356 etc)

For the momente there is only one commit, ficing the conflicts in #361.


valassi:
The dependency of check.exe and runTest.exe on the shared libraries is not ideal now.

First problem, the executables are now not statically built so one needs an LD_LIBRARY_PATH (ok maybe that's what we want...)
```
[avalassi@itscrd70 gcc10.2/cvmfs] /data/avalassi/GPU2020/madgraph4gpuX/epochX/cudacpp/gg_tt/SubProcesses/P1_Sigma_sm_gg_ttx> ./check.exe 
./check.exe: error while loading shared libraries: libmodel_sm.so: cannot open shared object file: No such file or directory
[avalassi@itscrd70 gcc10.2/cvmfs] /data/avalassi/GPU2020/madgraph4gpuX/epochX/cudacpp/gg_tt/SubProcesses/P1_Sigma_sm_gg_ttx> ./runTest.exe 
./runTest.exe: error while loading shared libraries: libmodel_sm.so: cannot open shared object file: No such file or directory
```

Second problem, there seem to be some hardcoded paths
```
[avalassi@itscrd70 gcc10.2/cvmfs] /data/avalassi/GPU2020/madgraph4gpuX/epochX/cudacpp/gg_tt/SubProcesses/P1_Sigma_sm_gg_ttx> ldd ../../lib/libmg5amc_cu.so 
        linux-vdso.so.1 =>  (0x00007ffcf6796000)
        ../../lib/libmodel_sm.so (0x00007f6e95b5e000)
        librt.so.1 => /lib64/librt.so.1 (0x00007f6e95817000)
        libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f6e955fb000)
        libdl.so.2 => /lib64/libdl.so.2 (0x00007f6e953f7000)
        libstdc++.so.6 => /cvmfs/sft.cern.ch/lcg/releases/gcc/10.2.0-c44b3/x86_64-centos7/lib64/libstdc++.so.6 (0x00007f6e95222000)
        libm.so.6 => /lib64/libm.so.6 (0x00007f6e94f20000)
        libgcc_s.so.1 => /cvmfs/sft.cern.ch/lcg/releases/gcc/10.2.0-c44b3/x86_64-centos7/lib64/libgcc_s.so.1 (0x00007f6e95b23000)
        libc.so.6 => /lib64/libc.so.6 (0x00007f6e94b52000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f6e95a1f000)
[avalassi@itscrd70 gcc10.2/cvmfs] /data/avalassi/GPU2020/madgraph4gpuX/epochX/cudacpp/gg_tt/SubProcesses/P1_Sigma_sm_gg_ttx> ldd ./check.exe 
        linux-vdso.so.1 =>  (0x00007ffcc37b7000)
        ../../lib/libmg5amc_cxx.so (0x00007f7fd9dcb000)
        libdl.so.2 => /lib64/libdl.so.2 (0x00007f7fd99bf000)
        libmodel_sm.so => not found
        libcurand.so.10 => /usr/local/cuda-10.2/lib64/libcurand.so.10 (0x00007f7fd591c000)
        libstdc++.so.6 => /cvmfs/sft.cern.ch/lcg/releases/gcc/10.2.0-c44b3/x86_64-centos7/lib64/libstdc++.so.6 (0x00007f7fd5747000)
        libm.so.6 => /lib64/libm.so.6 (0x00007f7fd5445000)
        libgcc_s.so.1 => /cvmfs/sft.cern.ch/lcg/releases/gcc/10.2.0-c44b3/x86_64-centos7/lib64/libgcc_s.so.1 (0x00007f7fd9d90000)
        libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f7fd5229000)
        libc.so.6 => /lib64/libc.so.6 (0x00007f7fd4e5b000)
        ../../lib/libmodel_sm.so (0x00007f7fd9d7d000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f7fd9bc3000)
        librt.so.1 => /lib64/librt.so.1 (0x00007f7fd4c53000)
```

I would be tempted to have the libmodel embedded inside the libcxx/libcuda. This actually means replicating some symbos, but for the moment I guess we WITHER use one OR the other? Eventually if we have all of them together we should have them inside a single library?... Something to be clarified related to heterogeneous apps #318

roiser:
Hi @valassi concerning the having one shared library or many, I kept them on purpose separately as the duplication of the symbols didn't make too much sense to me. Best

valassi:
A first cleanup of Makefiles and the move to shared libs is completed - the Bridge changes are all to be done.

Interestingly, using shared libraries is 1-2% slower in C++ for eemumu, but cannot see a difference for ggttgg. (Cuda may even be a tiny bit faster in the latter case? Probably just a coincidence not reproducible)

valassi:
Ok also need DYLD_LIBRARY_PATH on mac....

valassi:
A few failures, need to fix FPTYPE=f for the fbridge

valassi:
I am adding a full fortran test of the bridge - WIP to be completed

valassi:
The full fortran test is still WIP. Note that I removed LD_LIBRARY_PATH and I added rpath to the three executables check, gcheck, runTest, this is much more practical for simple tests. From MadEvent some LD_LIBRARY_PATH may be needed instead (or maybe not)

valassi:
Of course mac has to be different, does not like rpath origin... will check what to do

valassi:
I added a temporary patch for mac in #375, to be fixed (can be outside this PR)

I have added a draft fortran executable - runtime issue for the c++ version, does not build for cuda (must move from Cpu to Gpu versions of random and sampler, as in check.exe bridge test)

valassi:
I have completed a standalone Fortran test in PR #367. This is a Fortran program that fills a Fortran momenta array using a c++ sampler (common host + rambo sequence), and then computes MEs using a c++ or cuda bridge sequence. The results are the same in this Fortran program as they are in the standalone c++/cuda program.

valassi:
Ok all tests are finally succeeding in the CI, after fixing a few CI related issues.

Amongst the latest changes in this PR in the last few days:
- I implemented a test (also run through the CI make check) that computes an ME average for a fixed small number of events both through the C++ standalone executable (check or gcheck) and through the Fortran+Bridge standalone executable (fcheck or fgcheck) and compares the outputs using some basic python. The results have to agree within a given tolerance.
- In this test, the Fortran always uses double precision, but I also implemented the option of having the MEs in CUDA/C++ in single precision (see #5 and #212)
- The tolerance had to be adapted, because especially with float precision the results of the two chains can differen quite a lot. I think it is now a 2E-4, which is already huge. This is enough for float ggttgg tests, but would fail float ggttggg tests where the difference is even larger. The root causes of this are probably a combination of the fact that momenta are converted float to double back to float in the Fortran version, and maybe also of the more complex (and hopefully more precise) way in which the c+ executable computes ME averages uing EventStatistics.h

I am just running a few more tests to commit some logs, but otherwise this is ready to go

valassi:
Hi @roiser @oliviermattelaer this is now complete, I am about to merge it - sorry for the delay

PR #370, opened 2022-02-02
[alpaka] Add the code generation plugin for Alpaka to epochX.

smithdh:
This is a PR to add an Alpaka (epochX) plugin.


smithdh:
Hi. As this is adding new code (which nothing else should depend on yet) I'll merge it myself.

PR #377, opened 2022-02-07
[CI] Add gfortran install step for Mac.

### LINKED ISSUE #376: Enable/install a Fortran compiler for Mac in the CI
Hi @hageboeck would it be possible to install a Fortran compiler in the CI please?

I have now added a standalone Fortran test in PR #367, this is meant to mimic even more closely what will go on in the MadEvent integration, so it includes a Fortran program that calls two of C++/CUDA components, not only a ME calculation but also a sample (random+rambo). 

I have included this in the basic build, so now it is failing
https://github.com/madgraph5/madgraph4gpu/runs/5090307331?check_suite_focus=true
```
f77 -I. -I../../include -c fcheck_sa.f -o fcheck_sa.o
make: f77: No such file or directory
``` 
(I am using $(FC) in the makefile for the Fortran compiler, just like $(CXX) is the c++ compiler, in case you want to set it up to something else)

By the way, eventually I would like to create a basic test that also checks that the Fortran and c/cuda executables give the same results. I can add that as a target. Anyway, I was wondering, maybe I would rename the 'make check' target that is now used to run the test in the CI - the reason is that 'check.exe' is the name of the executable (well we could change that too, eventually), so I would prefer something a bit more explict like 'make civalidation'. 'make citests' or something similar. Suggestions/feedback?

Thanks! Andrea

hageboeck:
Try to
Fix #376

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -44,7 +44,9 @@ jobs:
           brew install gfortran
           brew link --overwrite gcc
     - name: Invoke gfortran
-      run: gfortran -v
+      run: |
+          ls /usr/local/Cellar/gcc/*/bin
+          gfortran -v
     - name: make info
       run: make -C ${{ matrix.folder }} info
     - name: make
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -42,7 +42,7 @@ jobs:
       run: |
           brew update
           brew install gfortran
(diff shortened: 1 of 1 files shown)


hageboeck:
As per [Github Docs](https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md), one can already invoke it via `gfortran-11`; confirmed in test CI run here.

PR #379, opened 2022-02-09
(nochange) Throughput logs in double/float eemumu with fast math disabled

valassi:
This is a no-change PR.

It includes throughput logs in double/float eemumu with fast math disabled. It is related to the study of nans and numerical stability in #212.


valassi:
This has no change. I am self merging.

Issue #239, opened 2021-07-23
Build and test on AMD CPUs


valassi:
It would be useful to build and test on AMD x86 CPUs and not only Intel.

Thanks to @lfield and his colleagues I have had access to an AMD EPYC at CERN.
The results are in PR #238.

Note that that node does not yet support AVX512.
It is an AMD EPYC 7302, so apparently a Zen2
https://en.wikichip.org/wiki/amd/epyc/7302

Instead AVX512 will be supported in 2021 by Zen4
https://www.techpowerup.com/279129/amd-zen-4-microarchitecture-to-support-avx-512

Anyway, the results on this older Zen2 are already quite interesting. Thanks again Laurence!

valassi:
Note: also Juwels Booster has Zen2 EPYC7302, with only AVX2, see PR #381

valassi:
A summary of results at Juwels Booster is here
https://github.com/madgraph5/madgraph4gpu/blob/a69d7f9ea37dd6445cd375e6b29a33f6a884e681/epochX/cudacpp/tput/summaryTable_juwels.txt#L50
```
*** FPTYPE=d ******************************************************************
+++ REVISION df441ad +++
On jwb0085.juwels [CPU: AMD EPYC 7402 24-Core Processor] [GPU: 4x NVIDIA A100-SXM4-40GB]:

[nvcc 11.5.50 (gcc 11.2.0)] 
HELINL=0 HRDCOD=0
            eemumu      ggtt        ggttg       ggttgg      ggttggg     
CUD/none    1.57e+09    1.69e+08    2.37e+07    9.45e+05    2.04e+04    
CPP/none    2.26e+06    2.63e+05    2.89e+04    2.17e+03    9.11e+01    
CPP/sse4    4.34e+06    3.94e+05    5.41e+04    4.15e+03    1.72e+02    
CPP/avx2    8.58e+06    8.69e+05    1.26e+05    9.92e+03    3.68e+02    

*** FPTYPE=f ******************************************************************
+++ REVISION df441ad +++
On jwb0085.juwels [CPU: AMD EPYC 7402 24-Core Processor] [GPU: 4x NVIDIA A100-SXM4-40GB]:

[nvcc 11.5.50 (gcc 11.2.0)] 
HELINL=0 HRDCOD=0
            eemumu      ggtt        ggttg       ggttgg      ggttggg     
[...]

PR #238, opened 2021-07-23
Tests on an AMD EPYC node

valassi:
Minor patches for AMD (only in the throughput script essentially)

Interesting performance, even in inline mode.
This AMD does not yet have AVX512 (it is a zen3 not a zen4?)

```
On b7s01p3272.cern.ch [CPU: AMD EPYC 7302 16-Core Processor] [GPU: none]:
=========================================================================
Process                     = EPOCH1_EEMUMU_CPP [gcc 9.2.0] [inlineHel=0]
FP precision                = DOUBLE (NaN/abnormal=0, zero=0)
Internal loops fptype_sv    = SCALAR ('none': ~vector[1], no SIMD)
Random number generation    = COMMON RANDOM (C++ code)
OMP threads / `nproc --all` = 1 / 64
EvtsPerSec[MECalcOnly] (3a) = ( 1.790062e+06                 )  sec^-1
MeanMatrixElemValue         = ( 1.372113e-02 +- 3.270608e-06 )  GeV^0
TOTAL       :     5.262333 sec
    18,640,514,529      cycles                    #    3.271 GHz                      (60.00%)
     1,432,917,809      stalled-cycles-frontend   #    7.69% frontend cycles idle     (60.11%)
     1,003,671,602      stalled-cycles-backend    #    5.38% backend cycles idle      (60.13%)
    48,606,388,470      instructions              #    2.61  insn per cycle
                                                  #    0.03  stalled cycles per insn  (60.14%)
       5.267442975 seconds time elapsed
=Symbols in CPPProcess.o= (~sse4:  614) (avx2:    0) (512y:    0) (512z:    0)
-------------------------------------------------------------------------
Process                     = EPOCH1_EEMUMU_CPP [gcc 9.2.0] [inlineHel=0]
FP precision                = DOUBLE (NaN/abnormal=0, zero=0)
Internal loops fptype_sv    = VECTOR[2] ('sse4': SSE4.2, 128bit) [cxtype_ref=YES]
Random number generation    = COMMON RANDOM (C++ code)
OMP threads / `nproc --all` = 1 / 64
EvtsPerSec[MECalcOnly] (3a) = ( 3.400763e+06                 )  sec^-1
MeanMatrixElemValue         = ( 1.372113e-02 +- 3.270608e-06 )  GeV^0
TOTAL       :     3.684801 sec
    13,394,877,121      cycles                    #    3.248 GHz                      (59.95%)
[...]

Issue #156, opened 2021-04-06
Detailed comparisons of Nvidia A100 and V100


valassi:
Hi @ingvildh, 

as discussed this morning, one thing that could be quite useful, for studying A100s, would be to start by comparing the performance to our V100 baseline.

I would suggest to start by taking the gg to ggtt code of epoch2 (issue #146, e.g. commit https://github.com/madgraph5/madgraph4gpu/commit/dd8711d0aa22c85429802b23de3169a68a97f298), run nsight compute and get a baseline for V100, then rerun for A100 and see if you can compare the various metrics. I have never used nsight compute to compare profiles from two different systems, but I assume it is possible.

Later on, one idea may be to look at tensor cores (#118), maybe for color algebra (#155)

Thanks! Andrea

valassi:
This can be useful for tensor cores #118 and specifically for color algebra on tensor cores #155. But we need to find the A100 cards first...

valassi:
Thanks to Stefan we now have A100, see PR #381

PR #236, opened 2021-07-20
CORI tests - first performance benefit of AVX512/zmm

valassi:
Interesting, first time I see AVX512z better than AVX512y...
Unfortunately I have no perf to chech better

```
On cori04 [CPU: Intel(R) Xeon(R) Gold 6148 CPU] [GPU: 1x Tesla V100-SXM2-16GB]:
=========================================================================
Process                     = EPOCH1_EEMUMU_CUDA [nvcc 11.4.48 (gcc 10.1.0)]
FP precision                = DOUBLE (NaN/abnormal=0, zero=0)
EvtsPerSec[MatrixElems] (3) = ( 7.520234e+08                 )  sec^-1
EvtsPerSec[MECalcOnly] (3a) = ( 1.425600e+09                 )  sec^-1
MeanMatrixElemValue         = ( 1.371706e-02 +- 3.270315e-06 )  GeV^0
TOTAL       :     0.542458 sec
real    0m0.663s
==PROF== Profiling "sigmaKin": launch__registers_per_thread 122
==PROF== Profiling "sigmaKin": sm__sass_average_branch_targets_threads_uniform.pct 100%
-------------------------------------------------------------------------
Process                     = EPOCH1_EEMUMU_CUDA [nvcc 11.4.48 (gcc 10.1.0)]
FP precision                = FLOAT (NaN/abnormal=2, zero=0)
EvtsPerSec[MatrixElems] (3) = ( 1.636039e+09                 )  sec^-1
EvtsPerSec[MECalcOnly] (3a) = ( 3.453580e+09                 )  sec^-1
MeanMatrixElemValue         = ( 1.371686e-02 +- 3.270219e-06 )  GeV^0
TOTAL       :     0.580860 sec
real    0m0.685s
==PROF== Profiling "sigmaKin": launch__registers_per_thread 48
==PROF== Profiling "sigmaKin": sm__sass_average_branch_targets_threads_uniform.pct 100%
=========================================================================
Process                     = EPOCH1_EEMUMU_CPP [gcc 10.1.0]
FP precision                = DOUBLE (NaN/abnormal=0, zero=0)
Internal loops fptype_sv    = SCALAR ('none': ~vector[1], no SIMD)
OMP threads / `nproc --all` = 1 / 80
EvtsPerSec[MECalcOnly] (3a) = ( 1.728110e+06                 )  sec^-1
MeanMatrixElemValue         = ( 1.371706e-02 +- 3.270315e-06 )  GeV^0
[...]

PR #381, opened 2022-02-17
Juwels tests (AMD EPYC and Intel Gold CPUs, NVidia A100 GPUs) + Makefile fixes

valassi:
A few tests on Juwels booster (thanks to @roiser for the access!)

This is AMD EPYC CPU and NVidia A100 GPU

A few (hopefully minor) issues to fix
- the AMD EPYC does not support AVX512, check if anything else can be done for vectorization (ok known issue, this is also a Zen2 (7402, not 7302), see #239 and #238)
- the A100 tells me it has the wrong ptx, I gess I need to fix the arch options (also #156)


valassi:
Hi @roiser thanks I had made a mistake, I tried to use "compute_80" but indeed "sm_80" is needed.

For reference, on our V100s "compute_70" is enough.

I made a patch using a different more complex structure for both.
For the moment this is a hack, I use 70 (V100) ALWAYS, and I also add 80 (A100) only if the hostname is juwels.

It can be improved later on... probably there is some build overhead by having both, to be studied. But to get going on A100 it will be ok I think.

Note by the way that JUwels Cluster has also V100s (which is why I did not want to remove them), while Juweles Booster has A100s. But our project for the moment only has access to Juwels booster.

I plan to complete this PR by ficing code generation and porting to all processes. Let me know if you have suggestions.

Thanks
Andrea

valassi:
I have included here some Makefile fixes to avoid crashes of parallel 'make -j' builds #382

valassi:
This is now complete
- the build options for A100 are fixed in codegen and all processes (in the end I am building EITHER for V100 OR for A100, as presently at juwels we cannot use Juwles cluster with V100)
- I have produced throughput logs and will comment in the relevant tickets (for A100 on #156, for AMD on #239, for INtel Gold vectorization on #173)
- interestingly, in fact, even on a Juwels Cluster login node I used some Intel Golds where I was able to see AVX512/zmm benefits (as previously observed at CORI, #236)
- I have produced summary tables for all these tests
- en passant I fixed a few issues with paralle builds 'make -j'

valassi:
Self merging

PR #386, opened 2022-02-24
Disable fail-fast in the github CI

valassi:
Do not cancel jobs in a matrix if another job failed
See https://github.community/t/error-the-operation-was-canceled-in-ci/166506/4
See https://stackoverflow.com/questions/61070925/github-actions-disable-auto-cancel-when-job-fails

This is important for the 'shared' PR #367, where I was getting several uncorrelated failures from different sources and I want to see them all


valassi:
Tets passed, self merging

@hageboeck : for info, note I disabled fail-fast in the matrix

hageboeck:
> Tets passed, self merging
> 
> @hageboeck : for info, note I disabled fail-fast in the matrix

@valassi acknowledged. You're mostly wasting github's resources if something fails.

Issue #360, opened 2022-01-27
Port epochX CODEGEN from 270gpu to the 311lovec branch (and pick up the new features there!)


valassi:
Hi @oliviermattelaer I assign this on myself - then we need to discuss when I should start.

All of my epochX CODEGEN is now based on the branch 270gpu of launchpad and specifically on revision 370
https://github.com/madgraph5/madgraph4gpu/blob/aa65924c34d1366f9652f427858e95465d2667f3/epochX/cudacpp/CODEGEN/MG5aMC_patches/2.7.0_gpu/revision.BZR#L1

This is this revision
https://bazaar.launchpad.net/~maddevelopers/mg5amcnlo/2.7.0_gpu/revision/370

I understand I should be moving instead to this other branch
https://code.launchpad.net/~maddevelopers/mg5amcnlo/3.1.1_lo_vectorization

For instance this includes the patch for issue #343 about nprocesses:
https://bazaar.launchpad.net/~maddevelopers/mg5amcnlo/3.1.1_lo_vectorization/revision/990

(En passant this is also eventually needed as a prerequisite for helicity recycling I guess, #279).

Details to be discussed, thanks Andrea

valassi:
In doing this upgrade, I should try to stick closer to the original code - adressing issues such as #341 reported by Olivier

valassi:
As discusset today with Olivier, note that the 311_lo_vec branch includes multichannel outputs. IIUC there is one more python function with one more argument, which creates the C++ with the additional argument needed for multichannel.

Note also Olivier's suggestion to try out this process to check all looks good
```
add process p p  > t t~
output PROC_vec
--vector_size=16
--me_exporter=standalone_gpu
```

(The "vector_size" may be renamed to something else, this is the number of events in one "iteration" of our standalone app).

valassi:
Hi @oliviermattelaer I have started to work on this in PR #391.

I changed the upstream MG5AMC that I am using with my plugins, from 2.7.0_gpu to 3.1.1_lo_vectorization. Surprisingly (or maybe not?), I see no changes at all in source code except for comments.

Maybe I should actually change the templates I am using to realign them to those from the new branch?...

valassi:
I am splitting this into two separate parts
- In PR #391, which I am about to merge, I am just moving my scripts to the 3.1.1 branch, but I am not modifying th eplugin
[...]

PR #391, opened 2022-03-03
Use the upstream 3.1.1 branch for code generation (without picking up new features yet)

valassi:
This is a WIP PR for #360 

I changed the upstream MG5AMC that I am using with my plugins, from 2.7.0_gpu to 3.1.1_lo_vectorization. Surprisingly (or maybe not?), I see no changes at all in source code except for comments.

Maybe I should actually change the templates I am using to realign them to those from the new branch?...


valassi:
For simplicity, I will already merge these first few changes. THis is teh first part of #360. I have moved code generation to a local installation of the 3.1.1 branch. 

But I have not yet picked up any new features from the 3.1.1 branch. This will be a next PR

Issue #389, opened 2022-02-25
Remove m_masses?


valassi:
While working on clang-format, I am reviweing code generation.

There are four hardcoded lines m_masses.push_back(0). I guess these must come from eemumu.

These lines should be removed in general code generation.

Should probably review if m_masses is needed at all, else better remove that...

valassi:
Fixed in PR #394, closing

PR #394, opened 2022-03-04
Fix for hardcoded m_masses and _sm

valassi:
This is mainly a fix for #389
- Remove the four hardcoded zero masses, include instyead n masses with the right values for HRDCOD=1: note that this has NO EFFECT in any case, because m_masses are not used anywhere... maybe would be even better to remove them, but for now I left them there
- I also replaced a hardcoded "_sm" by the approriate model name


valassi:
self merging

PR #395, opened 2022-03-05
Fixes and improvements for alpaka + Summary table of results

valassi:
Hi @smithdh thanks again for all your work on Alpaka!

It is in good shape to have some performance comparisons in my opinion. It’s especially nice that you also kept the standard c++ and CUDA builds using the same code base, so it’s possible to do a comparison directly there rather than rely on external numbers from a different build/tag.

As discussed, I am preparing a few fixes and improvements for Alpaka over your #370. Here's the first batch of changes
- adapt my generateAndCompare.sh script and add a symlink in your alpaka/CODEGEN directory: this works nicely, so I can produce ggtt for instance
- add the generated ggtt.auto to the repo, for reference: we discussed this in the past also for cuda, and while not strictly needed, I find it convenient (at least temporarely, then maybe this can be removed)
- fix the build of tests, which was failing: this means, port runTest.cc in your code to a few changes in the test infrastructure, namely the move to header-only MadgraphTest.h (#280)
- fix the tests at runtime: this mainly involved changing the path to the reference file path (#268), now all tests succeed
- backport all these changes to code generation

This is WIP. Within this same PR I would still like to add a few changes such as
- probably add a few sanity checks in the Makefile to choose whether an alpaka build is possible, or should be dropped (eg if BOOSTINC does not exist or similar)
- create a tput directory in alpaka for throughput tests: I would like to add symlinks and/or modified versions of my various scripts there
- these scripts would probably run check.exe AND gcheck.exe AND alpcheck.exe from your alpaka build, so this gives the difference between gcheck.exe/CUDA and alpcheck.exe/ALPAKA, on this ~goldenepochX2 tag
- add some logs for alpaka runs, as I did for cudacpp
- add a few scripts to create summary tables too (as I have in cudacpp), which can be used then for a paper
This essentially would complete a very good milestone for a paper/presentation

Would this sound ok for you? Can you maybe have a look already even if this is WIP and see how it seems?

Thanks! Andrea


valassi:
Hi @smithdh this is now ready for review and ready to be merged.

I have completed the things I wanted
-    created a tput directory in alpaka for throughput tests, with symlinks and/or modified versions of my various scripts there
-    these scripts run check.exe AND gcheck.exe AND alpcheck.exe from your alpaka build, so this gives the difference between gcheck.exe/CUDA and alpcheck.exe/ALPAKA, on this ~goldenepochX4 tag
-    added some logs for alpaka runs, as I did for cudacpp
-    modified the scripts to create summary tables too (as I have in cudacpp), which can be used then for a paper

Not done, can be done later on:
-    add a few sanity checks in the Makefile to choose whether an alpaka build is possible, or should be dropped (eg if BOOSTINC does not exist or similar), I have instead added a few sanity checks in my scripts

I think that this essentially completes a very good milestone for a paper/presentation. The summary table of results is the folowing: https://github.com/valassi/madgraph4gpu/blob/alpaka/epochX/cudacpp/tput/summaryTable_alpaka.txt

```
*** FPTYPE=d ******************************************************************

+++ cudacpp REVISION 09e482e +++
On itscrd70.cern.ch [CPU: Intel(R) Xeon(R) Silver 4216 CPU] [GPU: 1x Tesla V100S-PCIE-32GB]:

[nvcc 11.6.112 (gcc 10.2.0)] 
HELINL=0 HRDCOD=0
                eemumu          ggtt         ggttg        ggttgg       ggttggg
         [2048/256/12]  [2048/256/1]  [2048/256/1]  [2048/256/1]    [64/256/1]
CUD/none      1.33e+09      1.42e+08      1.44e+07      5.19e+05      1.22e+04
ALP/none      --------      --------      --------      --------      --------
         [2048/256/12]  [2048/256/1]    [64/256/1]    [64/256/1]     [1/256/1]
CPP/none      1.63e+06      2.00e+05      2.47e+04      1.80e+03      7.02e+01
CPP/sse4      3.10e+06      3.17e+05      4.50e+04      3.33e+03      1.30e+02
CPP/avx2      5.42e+06      5.63e+05      8.84e+04      6.82e+03      2.58e+02
CPP/512y      5.89e+06      6.15e+05      9.81e+04      7.49e+03      2.85e+02
CPP/512z      4.77e+06      3.84e+05      7.20e+04      6.54e+03      2.95e+02

+++ alpaka REVISION f5a44ba +++
On itscrd70.cern.ch [CPU: Intel(R) Xeon(R) Silver 4216 CPU] [GPU: 1x Tesla V100S-PCIE-32GB]:

[nvcc 11.6.112 (gcc 10.2.0)] 
HELINL=0 HRDCOD=0
                eemumu          ggtt         ggttg        ggttgg       ggttggg
         [2048/256/12]  [2048/256/1]  [2048/256/1]  [2048/256/1]    [64/256/1]
CUD/none      1.32e+09      1.43e+08      1.41e+07      3.09e+05      6.22e+03
ALP/none      6.29e+08      9.98e+07      4.69e+06      2.67e+05      (FAILED)
         [2048/256/12]  [2048/256/1]    [64/256/1]    [64/256/1]     [1/256/1]
CPP/none      1.34e+06      2.00e+05      2.46e+04      1.81e+03      6.71e+01
CPP/sse4      2.55e+06      3.13e+05      4.51e+04      3.31e+03      1.21e+02
CPP/avx2      4.57e+06      5.73e+05      9.00e+04      6.73e+03      2.42e+02
CPP/512y      5.01e+06      6.28e+05      9.99e+04      7.51e+03      2.63e+02
CPP/512z      3.54e+06      3.82e+05      7.24e+04      6.41e+03      2.76e+02
```

This allows a few interesting observations about alpaka:
- The alpaka implementation has some overhead over native CUDA, it is 10% to a factor 3 slower (but this looks really peculiar, I should maybe rerun all tests, also because there is no clear pattern)
- For the most complex process ggttggg, the alpaka implementation fails (not enough memory somewhere)

This also allows a few interesting observations about cudacpp on goldenX4 vs the current latest baseline (log REVISION 09e482e)
- The current CUDA is up to a factor 2 faster than golden epochX4, but only for the complex ggttgg and ggttggg, while the other processes look very similar
- The current CPP is very similar in performance to the golden epochX4, just a tiny bit faster, but not very remarkable

All these observations are very preliminary and should be repeated, but it looks like some interesting results are already appearing.

Is it ok if I merge this? Would you have some feedback? We can certainly repeat more tests, but these can come in a later PR.

Thanks!
Andrea

valassi:
I have rerun twice all tests, and within a few % of jitter I do get the same results, these are nidely reproducible.

valassi:
Hi @dhsmith, thanks for your feedback during the meeting. As discussed there, I will just go on and merge. We can always add any additional fixes later on.

Self merging.

PR #396, opened 2022-03-08
Disable mirror processes (nprocesses=2 in pptt) + Add process name to library names

### LINKED ISSUE #343: Fix "nprocesses>1" code generation (example: uutt within pptt)
Hi @oliviermattelaer as discussed today at the meeting I am assigning this to you.

This is a followup to #337 and to #272. The problem: code generation for pptt now succeeds (after fixing the assert in #337), however the code does not build in the uutt subdirectory, because nprocesses=2 there (#272).

 I checked that the uutt standalone and the uutt within pptt differ in the following way for instance:

```
/data/avalassi/GPU2020/madgraph4gpuX/epochX/cudacpp> diff -r pp_tt.auto/ uu_tt.auto/
...
diff -r pp_tt.auto/SubProcesses/P1_Sigma_sm_uux_ttx/CPPProcess.cc uu_tt.auto/SubProcesses/P1_Sigma_sm_uux_ttx/CPPProcess.cc
...
502c497
<     const int denominators = 36,36; // FIXME: assume process.nprocesses == 1 for the moment (eventually denominators[nprocesses]?)
---
>     const int denominators = 36; // FIXME: assume process.nprocesses == 1 for the moment (eventually denominators[nprocesses]?)
```
From our discussions, I understand that this is because "uutt" within pptt is not only "uu" proper, but also some other combinations of quarks (as you mentioned today, probably it is that u+ubar to t+tbar and ubar+u to t+tbar are not exactly the same?). Anyway, what I mean is that the ggtt within pptt and the ggtt standalone instead are essentially the same code, so it is really the "uutt" part which is different.

As discussed today, this probaby comes from the base python that you developed, rather than from my plugin modifying your python. So best if you take a look. Thanks! Andrea

PS I am instead merging PR #340 that contains ONLY the fix for the assert in #337

valassi:
This is a WIP PR for issue #343
- It integrates the latest patch by @oliviermattelaer in bazaar version 311, where mirror processes are disabled (so that nprocesses becomes 1 instead of 2 in the uutt subprocess of pptt)
- denominators becomes an array[nprocesses], as in bazaar version 311: the fix ensures that "int denominators = 36,36;" becomes "int denominators[1] = 36;"
- HOWEVER, the code still assumes that nprocesses==1: a static assert is even added about that, and the scalar "denominators" is replaced by "denominators[0]" everywhere...

This fixes #343, but we should still find an example with nprocesses>1 (#272) even in the absence of mirror processes (is there any such thing? otherwise can the "denominators[1]" array become a scalar again?)

Later on I might also add a patch for pptt library names as discussed with @roiser


valassi:
This is now complete. It addresses two issues for pptt
- issue #343, it removes mirror processes (nprocesses==2 for uutt)
- issue #397, it adds the process name to the library names

valassi:
All checks passed, self merging

PR #410, opened 2022-03-29
Alphas part1

roiser:
State of play of alphas, i.e. the bridge interface is there and the passing of the Gs up to CPPProcess. Missing is the calculation of the GC_XX variables as vectors and the ME calculations. I'll create a separate branch for those.

Note it likely will create a lot of merge conflicts as this is the code BEFORE the clang-format'ing of the upstream/master


roiser:
This is superseded by #420 which is the same but fixed conflicts on top of the latest master

PR #411, opened 2022-03-30
Fix warnings on signedness of mostly integer variables

roiser:
Some fixes in the code where signed and unsigned integers were compared or assigned. In practice it probably won't make much of a change b/c the values of the variables would not go that high but probably better to fix those.


valassi:
Hi @roiser, review is all good. However there were a few issues, and replaced this #411 by a new PR #431. The difference is fixing a few build errores and warnings, the backport to code generation and the regeneration of all eemumu and ggtt*.

I am closing this.

PR #412, opened 2022-03-31
CI updates for vGPU machine

hageboeck:


valassi:
Hi @hageboeck is this still needed? Your new vGPU CI is actually working very well already, much less time to wait :-)

I guess it has more slots... I am using gcc10.2 for development anyway, I do not think we need gcc11 in the CI.

Thanks again for the work on this, very nice!

hageboeck:
Not needed after reinstalling

PR #420, opened 2022-04-04
Alpha s part1 newmaster

roiser:
As I'm continuing to work on alphaS I had merged the new master (with clang-format) with the alphas_part1 status. So this PR is the same as #410 but with the merging work already done. Should be straight forward to merge into upstream.


roiser:
Superseded by #422.

Issue #221, opened 2021-06-29
Build and test using ARM vector extensions 


valassi:
This was discussed during the June 28 meeting https://indico.cern.ch/event/1053713/
This is mainly about ARMv9 Scalable Vector Extension 2 (SVE2)... of course we need to find a test system first.

What would this give on the AVX512 issue #173 ?

In theory, vector widths up to 2048 bits could be supported? See for instance
https://levelup.gitconnected.com/armv9-what-is-the-big-deal-4528f20f78f3

valassi:
A few interesting links about this issue
- https://arxiv.org/abs/1803.06185 The ARM Scalable Vector Extension (2018)
- https://levelup.gitconnected.com/armv9-what-is-the-big-deal-4528f20f78f3 (2021)
- https://developer.arm.com/documentation/100891/0612/sve-overview/introducing-sve

roiser:
That's interesting, I have some connections to UK ARM company and HPC centers. Let's discuss next time.

valassi:
Hi @roiser I dump here a few comments, which wer etriggered by your PR #421 but go beyond that. (This even goes beyond ARM so maybe I should open a new issue, but it is easier to discuss it here).

Until now, before your PR #421 for mac M1 ARMv8, we had only two types of SIMD:
- the full spectrum of Intel, where I initially developed this: none, sse4 (128bit), avx2 (256bit), 512y (avx512 limited to 256bt ymm registers), 512z (avx512 with 512bit zmm registers)
- Power9 from CINECA (PR #223), which only has two options: none, or 128bit VSX

The important points here in what I designed are two
- The SIMD level must be strictly configurable: I want to be able to test the same platform with no simd and compare it to several levels of SIMD. This means I want to be able to have a 'none' option.
- SIMD is implemented via compiler vector extensions that are fully controlled in the code. Depending on some ifdefs, the length neppV of the fptype_v vector is set. There is no autovectorization in the matrix element calculation (but there may be some elsewhere, for the moment...).

The patch I added for Power9 is quite ugly, but that's what it is. Since Power9 has no SSE4, but the logic of the code internally depends on -D__SSE4_2__, in the Makefile I am simply adding this flag for Power9 if AVX=sse4 is set
https://github.com/madgraph5/madgraph4gpu/blob/898e8b4c98356c5e31b8bd4755b9f8cdd8d98865/epochX/cudacpp/ee_mumu/SubProcesses/Makefile#L216

[...]

PR #421, opened 2022-04-04
enable ARM NEON (128 bit) vector registers via compiler defined macros

roiser:
enable ARM NEON

### DIFF
--- a/epochX/cudacpp/gg_tt/SubProcesses/P1_Sigma_sm_gg_ttx/check_sa.cc
+++ b/epochX/cudacpp/gg_tt/SubProcesses/P1_Sigma_sm_gg_ttx/check_sa.cc
@@ -766,6 +766,8 @@ main( int argc, char** argv )
 #else
   wrkflwtxt += "/sse4";
 #endif
+#elif defined __ARM_NEON__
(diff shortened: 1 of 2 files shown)


valassi:
Hi @roiser I added several comments about this in #221.

Bottom line:
- it presently does not support a way to disable SIMD (ie the 'none' option)
- it addresses the sam eissues as 128-bit vsx on Power9, but in a different way
- for these reasons I would modify this to use the same approach as on Power9 (ie add -D__SSE42__ if you want 128bit neon vectors, and add nothing if you want to keep no simd)
- later on this should probably be changed to use a define like -Dmg_prefer_vector_width128 instead of -D__sse42, and use that on all of intel, power9, arm

valassi:
Hi @roiser I am closing this one and replacing it by PR #425. Please have a look (but still in WIP)... thanks!

valassi:
I have now merged #425 which includes and replaces this #421.

roiser:
* later on this should probably be changed to use a define like -Dmg_prefer_vector_width128 instead of -D__sse42, and use that on all of intel, power9, arm

On this point, yes I think we should go away from Intel specific names and use generic ones. All very fine with me, thanks

PR #387, opened 2022-02-24
Basic CMake configuration

roiser:
This PR produces a functionally working CMake configuration for CPU builds. For the code generation one modification upstream in the MG5 code proper would be needed 

```
=== modified file 'madgraph/iolibs/export_cpp.py'
--- madgraph/iolibs/export_cpp.py	2022-01-27 14:10:48 +0000
+++ madgraph/iolibs/export_cpp.py	2022-02-24 10:29:32 +0000
@@ -2498,7 +2498,7 @@
         with misc.chdir(self.dir_path):
             logger.info('Creating subdirectories in directory %s' % self.dir_path)
 
-            for d in ['src', 'lib', 'Cards', 'SubProcesses']:
+            for d in ['src', 'lib', 'Cards', 'SubProcesses', 'CMake']:
                 try:
                     os.mkdir(d)
                 except os.error as error:
@@ -2982,7 +2982,3 @@
         return  ProcessExporterMatchbox(cmd._export_dir, opt)
     else:
         return cmd._export_plugin(cmd._export_dir, opt)
```

PR #423, opened 2022-04-06
Basic CMake configuration with additional fixes

valassi:
Hi @roiser this includes your PR #387 but also adds some additional fixes

In particular:
- I had to fix a conflict in your branch, it was no longer merging faultlessly on upstream/master
- There was a bug in codegen, the CMake directory was not created (the three files were copied to the same file CMake)
- For consistency with everything else I have done so far, I have committed the latest codegen-ed code for the four proceses eemumu and ggtt*, in both auto and manual (you had only included that in gg_tt manual)

This replaces PR #387, that I will close. I will merge this one instead.


valassi:
All checks have passed, I am merging (as agree with @roiser)

PR #424, opened 2022-04-06
Further improvements for CMake

valassi:
To be consistent with my previous changes, I have moved also the CMakeLists_P.txt to the CMake/SubProcesses template subdirectory


valassi:
There is no change in the code - self-merging without waiting for CI cjhecks

Issue #429, opened 2022-04-12
Bug in on-device curand reproducibility (was: differences for GPU execution bridge / no bridge)


roiser:
With the current branch I'm observing a strange behaviour when running (g)check.exe. 

The produced values for check.exe (CPU) with or without the bridge option and also gcheck.exe with the bridge option are the same. 

Running with gcheck.exe without bridge option I can currently produce different output results. E.g. I'm showing below the command lines and the last result

Executing
```
roiser@itscrd02:P1_Sigma_sm_gg_ttx (master)$ ./gcheck.exe -v 32 32 2
```
produces last result 
```
Momenta:
   1  7.500000e+02  0.000000e+00  0.000000e+00  7.500000e+02
   2  7.500000e+02  0.000000e+00  0.000000e+00 -7.500000e+02
   3  7.500000e+02  1.977175e+02 -5.098854e+01 -7.216702e+02
   4  7.500000e+02 -1.977175e+02  5.098854e+01  7.216702e+02
-------------------------------------------------------------------------------
 Matrix element = 10.8827 GeV^0
-------------------------------------------------------------------------------
```

while executing 

```
roiser@itscrd02:P1_Sigma_sm_gg_ttx (master)$ ./gcheck.exe -v --bridge 32 32 2
```
produces last result

```
Momenta:
   1  7.500000e+02  0.000000e+00  0.000000e+00  7.500000e+02
   2  7.500000e+02  0.000000e+00  0.000000e+00 -7.500000e+02
   3  7.500000e+02 -6.264599e+02 -4.169666e+01 -4.102552e+02
   4  7.500000e+02  6.264599e+02  4.169666e+01  4.102552e+02
-------------------------------------------------------------------------------
 Matrix element = 0.862187 GeV^0
-------------------------------------------------------------------------------
```

The binaries are produced with a plain "make" in the P1 directory. 

Also important to note is that the above difference does not show up with 1 iteration, e.g. executing

```
roiser@itscrd02:P1_Sigma_sm_gg_ttx (master)$ ./gcheck.exe -v --bridge 32 32 1
```

and

```
roiser@itscrd02:P1_Sigma_sm_gg_ttx (master)$ ./gcheck.exe -v 32 32 1
```

produces same results for me.

valassi:
Hi Stefan, thanks, I am looking into this. This is very peculiar.
```
for args in "-p 32 32 1" "-p 32 32 2" "-p 32 256 2"; do
[...]

PR #430, opened 2022-04-14
Workaround for reproducibility issue in curand-on-device

valassi:
This PR is a workaround for reproducibility issue #429 in curand-on-device


valassi:
All checks have passed. I am self merging

Issue #341, opened 2022-01-21
cudacpp plugin hardcode some change that modifies code outside of the sandbox of the plugin


oliviermattelaer:
Hi Andrea,

I was implementing the multi-channel for the standalone_gpu class, and was surprised that some object were pointing to your PLUGIN. looks like model_hangling.py is directly overwritting references to class in other file of mg5amc, which creates side effect to other codes.

Are those for debugging? if not then we should find a better way for you to specify the class that you want (via class attribute and subclassing)

Cheers,

Olivier

valassi:
Hi Olivier,
no some things may have been for debugging, but some are not.

For instance this one was about disabling formatting (but ok, still aesthetics)
https://github.com/madgraph5/madgraph4gpu/blob/5ac285cebc04cd7eb4d6679c6a3a5e2645803d30/epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/model_handling.py#L47

We should probably sit down (even in zoom for the moment) and discuss the issues one by one (or start with the general ideas).

I agree, eventually the plugin stuff should get better integrated into the standard things, we need to discuss how to do that.

By the way one thing I can think of is that I would (and have not done yet) change very significantly CPPProcess, both in the class itself and in how it is generated. 

Anyway, its good that we are already at the level of converging and cleaning up, somehow :-)

You are right in assigning this to me, but for th emoment I cannot do much. I need to discuss with you the details first to understand. Or maybe I can start to have a look when/if you tell me that I should port my stuff to a specific more recent branch and commit of the code - when this is possible, let me know!

Andrea

PS This is also vaguely related to helicity recyling #279... more generally, I am well aware that my plugin works on top of a specific version of 270gpu, while it should be ported to more recent versions of the code (and be integrated in a more maintainable way).

valassi:
To be discussed within #360 - upgrading from launchpad 270gpu to 311lovectorization

valassi:
Hi @oliviermattelaer thanks for the discussion today.

Ok I understand better your concern now: the problem is that, if my plugin is present in a Madgraph installation, this presently affects other outputs even if my output is not requested at all. Clearly this isa problem.

I guess there are two ways to address this, both of which should be looked at

1. I can try to make the plugin more encapsulated. For instance, replace export_cpp/misc etc components only if my plugin is used, not all the time if it is present. 

[...]

PR #433, opened 2022-04-20
WIP fix to prevent the cudacpp plugin from modifying the global environment (issue #341)

valassi:
This is a WIP fix for #341 

It includes several types of modifications
- some function redefinitions have been removed (they seem to be not really necessary?)
- as suggested by @oliviermattelaer, the modified HelasCallWriter is now set via the aloha_exporter hook (this seems to be the main change required to remove the bug for standalone_gpu, however now the standalone_cudacpp code is badly formatted: this is WIP, I still need to fix this)
- for some modules, like export_cpp, a second "deep copy" is made (actually this is not really a deep copy, it is a second loading of the module, as in https://stackoverflow.com/a/11285504)

Still some issues to sort out, but this issue seems possible to solve


valassi:
The patch includes some changes that require the fix for https://github.com/oliviermattelaer/mg5amc_test/issues/11 in revision 1001

PR #428, opened 2022-04-12
alphaS w/ buffers and calculation kernel

roiser:
A new PR which supersedes #422, this one includes the move to MemoryBuffers for the GC_ variables and the calculation of the GC_ values via a GPU kernel

PR #434, opened 2022-04-21
AlphaS against latest master, including code generation and all processes generated

valassi:
Hi @roiser I am creating here a PR superseding your #428. The main differences will be
- porting to the current latest upstream/master
- backporting your changes (in ggtt only for now) to code generation, and generating all five processes I use


valassi:
I am making progress on integrating this, but there is work to be done.

I would say the main problems were:
- while check.exe and gcheck.exe were ok, both runTest and the fcheck test were failing https://github.com/madgraph5/madgraph4gpu/pull/434/commits/78c4ed33e412e61ce6854d9e1f892e77f9ebfc54
- code generation was not there, and the design of the interface was using gc10/gc11 which are process-dependent parameters that should be removed from the interface (in eemumu they are not needed, for instance)

What I have done so far:
- I have removed gc10/gc11 from the bridge interfaces, and only kept gs (NB: gs are actually also not needed in some processes like eemumu as this is a QCD coupling, but the whole Madgraph machinery distinguishes between couplings that are dependent or not on alphas, so for the moment I would keep the gs as-is in the interfaces)
- gc10 and gc11 are now allocated within MEK, no longer within check_sa: they are an internal detail of MEK, they do not need to be exposed outside it
- there was no way to pass gs through fbridge.inc from fortran code, I have now added that
- partly thanks to the above, the fcheck tests are now succeding (amongst other things, the hardcoded value of gs that is set in check_sa.cc also needed to be set in fcheck_sa.f, and is now passed through fbridge.inc)
- I have removed the computeDependentCouplings method from the MEK and from the Bridge: given that only Gs (not gc10, gc11) are passed through the bridge interface, I aminclined to treat thecomputation of dependent couplings as one instance of 'splitting kernels'... I will cook up something for now (possibly moving it to calculate_wavefunction), eventually this should be reviewed together with splitting kernels
- I have started to rename and move the actual computation of dependent functions: I moved it to HelAmps for now, as it is a process-dependent computation (essentially it should only appear in HelAmps.h, or CPPProcess, or Parameters_sm: these are the only three files with process dependent stuff, all the rest I am trying to keep process independet)

Amongst the things tha are not yet ok
- runTest still fails
- remove gc10 and gc11 from sigmakin and similar interfaces, just keep g
- (minor reminder: copying Gs inside the bridge must be fixed if fptype is not FORTRANFPTYPE)
- (minor reminder: check timers in check_sa, find the right spot for copying gs)

valassi:
Ok I have now
- fixed runtest (was missing hardcoding gs)
- fixed fptype!=FORTRANFPTYPE (and the fgcheck test in float mode now succeeds, it was failing)

Still t do amongst other things
- remove gc10 and gc11 from sigmakin and similar interfaces, just keep g (this is necessary for code generation)
- (minor reminder: check timers in check_sa, find the right spot for copying gs)

In addition
- investigate performance? I saw the number of registers has increased from 170 to 200 in ggtt double

valassi:
I need to understand ho wto do gc10/gc11 with respect to code generation. I am dumping my main ideas in #356, as this is mainly about how to split the roles of MEK and CPPProcess (very much relatedto splitting kernels #310)

roiser:
My idea here was to let the code generation do the work, but if you want to change this its also fine with me.

valassi:
Well... I am doing it precisely because this will make it easier for code generation to do the work ;-)

valassi:
(I mean, we need to write the code generation... if we start having a variable number of parameters in the functions because sometimes we need gc10 and gc11, sometimes one more, sometimes none, writing the code generation becomes much more complicated, IMO)

valassi:
I am looking at memory access functions more in detail now. I think that those for Gs and for couplings were using my 'trivial' version for wavefunctions that was always accessing one same address, so I changed them and it is tricky (I am not really sure why it worked before, but we never tested it with anything else than a constant hardcoded G anyway?)

valassi:
I have improved the machinery of memory access so that more cases than before are vectorised. I need this for Gs and couplings. In particular I started from matrix elements, now the writing of non-const MEs is also vectorised (fixing #435). This even has a tiny ~1% speedup as expected, but that's not the point - the point is that now the code is more flexible also for couplings (and eventually for wavefunctions in splitting kernels).

roiser:
Hi @valassi, I think there was a misunderstanding about the code generation part. The idea of this PR was to check the code and agree on a version of it. The code generation should then be created once we agree on it. I see you have some ideas on how to implement things differently. I will have a look at those more in depth next week. So its good to not implement code generation twice.

valassi:
This is now almost completed:
- the memoryaccess for couplings is fully functional, using the new AOSOA layout
- I have cleaned up Parameters_sm and even easily reenabled the hardcoded option (probably totally useless, anyway)

Just two very minor things to do before code generation:
- move the G2COUP function implementation to Parameters_sm.h as an inline function (the code is there in the hardcoded code branch)
- add some constexpr parameters like idcoup_gc_10=0 and idcoup_gc_11=1and use those instead of my hardcoded 0, 1 in the idcoupAccessBuffer functions

The code will then be in a state where it should be very easy to backport it to code generation. Then regenerate eemumu, ggttg*, rerun all tests, and merge...

valassi:
ah, one small reminder more, for the hardcoded Gs take them from the existing alphas maybe instead of having a hardcoded number

valassi:
I finally completed the first draft of code generation for alphas: I am able to regenerate the full ggtt code.

The problem is that the generated code in eemu and ggttg fails to build:
- for eemumu see https://github.com/madgraph5/madgraph4gpu/pull/434/commits/c694c98807f5580385abfc36e054325f4fb8a951 (I am assuming that the old cIPC were all alphas dependent, but probably this is not the case?)
- for ggttg see https://github.com/madgraph5/madgraph4gpu/pull/434/commits/b57a1a369e6d883b6ba3c7dd93783c6c6cf85282 (here the problem is a bit simpler - but I will look at it tomorrow or later - I was assuming a calculation based only on G, but I am using code where intermediate variables are used)

valassi:
Indeed for eemumu the probalem is that there are cIPC and they are not aS dependent.

This is quite annoying. It means that for eemumu we must keep the cIPC machinery (and for a process that has electroweak and also QCD we will mix cIPC and the new Gs)

See the diffs in CPPProcess.cc here
https://github.com/madgraph5/madgraph4gpu/pull/434/commits/c694c98807f5580385abfc36e054325f4fb8a951#diff-15b7c54622029b2b90bd604f81bd6ecdc9dda30bf417f5b9655f219540bd9a3f
```
#ifdef MGONGPU_HARDCODE_CIPC
  __device__ const fptype cIPC[6] = { (fptype)Parameters_sm::GC_3.real(), (fptype)Parameters_sm::GC_3.imag(), (fptype)Parameters_sm::GC_50.real(), (fptype)Parameters_sm::GC_50.imag(), (fptype)Parameters_sm::GC_59.real(), (fptype)Parameters_sm::GC_59.imag() };
  __device__ const fptype cIPD[2] = { (fptype)Parameters_sm::mdl_MZ, (fptype)Parameters_sm::mdl_WZ };
#else
#ifdef __CUDACC__
  __device__ __constant__ fptype cIPC[6];
  __device__ __constant__ fptype cIPD[2];
#else
  static fptype cIPC[6];
  static fptype cIPD[2];
#endif
#endif
```
vs
```
#ifdef MGONGPU_HARDCODE_CIPD
  __device__ const fptype cIPD[2] = { (fptype)Parameters_sm::mdl_MZ, (fptype)Parameters_sm::mdl_WZ };
#else
#ifdef __CUDACC__
  __device__ __constant__ fptype cIPD[2];
#else
  static fptype cIPD[2];
#endif
#endif
```

valassi:
In this case, eemumu needes gc3, gc50 and gc59 (I guess ffZ, ffgamma and in any cas ffW couplings?)

valassi:
I have fixed the ggttg issue and am now running tests for all of ggtt, ggttg, ggttgg, ggttggg.

The eemumu issue is MUCH more complex. The problem is that effectively we have now introduced two radically different ways to handle QCD and non-QCD couplings (those that depend on aS and those that do not). I have moved code generation to use the QCD style, but now this fails for eemumu non-QCD style. Even recovering bits and pieces of code from upstream/master before this PR is not enough (eg the cIPC and copy to constant GPU memory), simply because the interfaces have now all changed: the FFV functions were expecting a single cxtype for a fixed coupling for eemumu, they are now expecting an fptype pointer to global memory from where event-by-event QCD couplings can be taken for ggtt*.

There are various options, which are either unstaisfactory or demand quite some work, or both:
- 1. One could imagine having the old FFV interface with cxtype for non-QCD couplings and the new FFV interface with fptype* for QCD couplings, but this would demand a lot of work (may be a code generation nightmare, especially for me as I do not know all the inetrnals of the code)... and in the end it also results in rather non-elegant code. It is probably the most performing, however.
- 2. The opposite end of the spectrum is maybe the simplest, but the least performing: one can treat non-QCD couplings as if they were also running couplings, ie store them in an fptype array where one can get theoretically a different non-QCD coupling event by event (maybe this is actually needed at some point for precise electroweak physics, @oliviermattelaer?...)
- 3. One option in between, which is probably also quite complex for codegen, but results in simpler code overall and should not have big performance overheads (apart from the fact that couplings would be in global memory and not constant memory or hardcoded) is the following. One can keep the same fptype* API for FFV functions for both QCD and non-QCD couplings. However then one can have two parallel sets of Coulings, with two tyes of Buffers and two types of MemoryAccessCouplings, for fixed and running couplings. This means having a CF_ACCESS for fixed couplings and a CR_ACCESS for running couplings. The running coupling version decodes a large AOSOA in global memory, retrieves an event by event coupling. The fixed coupling version would do the 'TRIVIAL' access (exactly what was there in this PR at the very beginning), and essentially point to a small buffer that contains only two floats, ie one complex, always the same for all events.

I would totally rule out option 1, because having two different signatures for fixed and running couplings sounds like a nightmare. Option 2 may be faster to implement than option 3? And in any case it loses performance for non-QCD processes, ie typically it does not affect LHC physics processes. Option 3 does seem attractive, but looks like a lot of work, and maybe we do not really need it?

I would probably be tempted to go for option 2... then we can check how much performance we lose in eemumu, and reassess whether we need to go to option 3 (or something else still...).

oliviermattelaer:
Hi,

I agree that "i" is a nightmare for you but likely not something too complicated for me.
The only remark is that for your example, one could switch the signature from one scheme to the other since one process is pure QCD and the other pure QED, however some process (like p p > t t~ a --where "a" means photon) we will need both version of the FFV vertex (which is not a problem per say, just need to add a flag).

For "ii" you are not wrong that maybe one day every coupling would be running and that not all of those running will be related to the strong coupling. But this is likely not for LHC.

valassi:
Thanks Olivier for the comments :-)

In the end I quickly tried option 3, and it was easier than I thought to get some code up and running. Now my emumu succeeds gpu and simd/none, so I just need to fix the SIMD part (it gives wrong MEs). Then doing the codegen should be realtively easy, I tried to keep the code in such a way that codegen can be reused from what I have (in this PR or before). Will keep you psted.

valassi:
I am having some trouble to understand this now, putting stuff of the code together
- coups_dep is in class PLUGIN_UFOModelConverter(PLUGIN_export_cpp.UFOModelConverterGPU), here I can distinguish between depdent and independent couplings
- couplings2order is in class PLUGIN_OneProcessExporter(PLUGIN_export_cpp.OneProcessExporterGPU), here I have a list of all couplings, but I need to separate dep and indep, how do I get this here?
- the ufomodelconverter is set in PLUGIN_ProcessExporter.create_model_class in output.py, the oneprocessexporter is set in PLUGIN_ProcessExporter.oneprocessclass in output.py: is there any way I can find the ufomodelconverter (to have the distinction between dep and indep) inside the oneprocessesxporter? from PLUGIN_ProcessExporter top class I can go down to both ufomodel and oneprocess, but can I go back up to PLUGIN_ProcessExporter from either ufomodel or oneprocess?

oliviermattelaer:
Well, the conversion of the model from the UFO format is done really at the end of everything.
The "reason" is that like that we can export to Fortran/CPP only what is needed and not the full model.

This means that when you are "in" PLUGIN_OneProcessExporter, they are NO instance of PLUGIN_UFOModelConverter
instantiated.

But the information about the model should be available, I'm looking into that

valassi:
Thanks Olivier, I managed to find it! :-)

 

        for coup in coupling:

            keep = True

            # Use the same implementation as in UFOModelConverterCPP.prepare_couplings (assume self.model is the same)

            for key, coup_list in self.model['couplings'].items():

                if "aS" in key and coup in coup_list: keep = False

            if keep: coupling_indep.append( coup ) # AV only indep!

 

Seems to work...

Andrea

oliviermattelaer:
Yes should be working. (Likely way to have nicer code.)

like:
'''
       all_coupling_indep = sum([coup_list for  key, coup_list in self.model['couplings'].items() if "aS" not in key])
       coupling_indep = [coup for coup in coupling if coup in all_coupling_indep]
'''

valassi:
This is FINALLY completed
- code generation is complete
- both eemumu with fixed couplings and ggtt* with running couplings are handled in codegen (I have stripped off testing one process with both of them to #438)
- I have regenerated and tested all 5 processes and created a nice summary table
- WARNING! I have regenerated heft_gg_h and this generates ok, but it fails to build at runtime for the rather complex issue #439... I will NOT fix this now, but I will merge this anyway as I think that SM physics for LHC takes priority over BeyondSM generations
- WARNING! I have regenerated uudd and this generates ok, but it fails to build with another issue #440. Hopefully this is rather simple. Anyway again I will merge this as is because this has been dragging on for far too long...
- WARNING! I have also regenerated eemumu.mad, ggtt.mad, ggtt.madonly. While ggtt.madonly builds fine as expected, the two .mad builds fail. This is expected, because I have not integrated the alphas from madevent into cudacpp. This will be one of the next priorities, #441. I will merge this standalone cudacpp anyway.

valassi:
Note the performance difference between pre and post alphas:
https://github.com/madgraph5/madgraph4gpu/blob/7d1a6d9b604578c0957ebbf566539d4fb6121aac/epochX/cudacpp/tput/summaryTable_alphas.txt

Example
```
-------------------------------------------------------------------------------

+++ cudacpp REVISION 88fe36d1 +++
On itscrd70.cern.ch [CPU: Intel(R) Xeon(R) Silver 4216 CPU] [GPU: 1x Tesla V100S-PCIE-32GB]:

[nvcc 11.6.124 (gcc 10.2.0)] 
HELINL=0 HRDCOD=0
                eemumu          ggtt         ggttg        ggttgg       ggttggg
         [2048/256/12]  [2048/256/1]  [2048/256/1]  [2048/256/1]    [64/256/1]
CUD/none      1.32e+09      1.42e+08      1.44e+07      5.13e+05      1.21e+04
         [2048/256/12]  [2048/256/1]    [64/256/1]    [64/256/1]     [1/256/1]
CPP/none      1.66e+06      2.00e+05      2.48e+04      1.80e+03      7.15e+01
CPP/sse4      3.09e+06      3.15e+05      4.50e+04      3.34e+03      1.30e+02
CPP/avx2      5.47e+06      5.69e+05      8.82e+04      6.79e+03      2.60e+02
CPP/512y      5.87e+06      6.12e+05      9.81e+04      7.46e+03      2.83e+02
CPP/512z      4.67e+06      3.85e+05      7.22e+04      6.56e+03      2.95e+02

+++ cudacpp REVISION bae5c248 +++
On itscrd70.cern.ch [CPU: Intel(R) Xeon(R) Silver 4216 CPU] [GPU: 1x Tesla V100S-PCIE-32GB]:

[nvcc 11.6.124 (gcc 10.2.0)] 
HELINL=0 HRDCOD=0
                eemumu          ggtt         ggttg        ggttgg       ggttggg
         [2048/256/12]  [2048/256/1]  [2048/256/1]  [2048/256/1]    [64/256/1]
CUD/none      1.10e+09      1.36e+08      1.17e+07      4.91e+05      1.11e+04
         [2048/256/12]  [2048/256/1]    [64/256/1]    [64/256/1]     [1/256/1]
CPP/none      1.65e+06      1.99e+05      2.46e+04      1.79e+03      7.21e+01
CPP/sse4      3.10e+06      3.13e+05      4.43e+04      3.33e+03      1.32e+02
CPP/avx2      5.35e+06      5.61e+05      8.80e+04      6.76e+03      2.61e+02
CPP/512y      5.96e+06      6.11e+05      9.91e+04      7.58e+03      2.91e+02
CPP/512z      4.60e+06      3.80e+05      7.15e+04      6.48e+03      2.93e+02

-------------------------------------------------------------------------------
```

There is almost no difference

valassi:
(Well, about performance: not sure at all why eemumu would be slower... I opened #442 to investigate further if someone has the courage).

This is now complete. I am self merging.

PR #444, opened 2022-05-05
[CI] Introduce environment variable REQUIRE_CUDA.

hageboeck:
If REQUIRE_CUDA is set (to an arbitrary value), the presence of a CUDA
installation is checked at build time.

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -53,6 +53,7 @@ jobs:
     runs-on: self-hosted
     env:
       FC: gfortran
+      REQUIRE_CUDA: 1
     strategy:
       matrix:
         folder: [ epochX/cudacpp/ee_mumu/SubProcesses/P1_Sigma_sm_epem_mupmum , epochX/cudacpp/gg_ttgg/SubProcesses/P1_Sigma_sm_gg_ttxgg , epoch2/cuda/ee_mumu/SubProcesses/P1_Sigma_sm_epem_mupmum ]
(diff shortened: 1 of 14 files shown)


hageboeck:
@valassi 

I guess the environment variable works.

### 1st step
- Which folders should we port it to (or should we move it to a different place)?
- Another option is to try to it later only during the invocation of the tests, so you would at least see the CPU build. That might not be too useful, though, since you won't see the CUDA build.

### 2nd step
After agreeing where we should check the availability of CUDA, we restart the CI runners, so they pick up the new environment that tells them where CUDA is. That should make all tests green that are now failing.

### 3rd step
Merge

hageboeck:
> Actually it looks good as it is now, do you need anything else (apart from restarting the CI so that CUDA is actually found and the tests do not fail?)
> 
> I mean, I think I would keep this change you made physics-process-agnostic: as long as you have named a test "something-GPU", whether the something is eemumu or ggtt or epoch2 or epochX, I would say that they should all fail if CUDA is not found. Does this answer your question?

Yes, partly, but this would mean that the check of the environment variable has to be ported to all Makefiles of epochX + maybe also the code generator. Or do you see a way to do it only in one single place?

valassi:
Hi Stephan, you are completely right, I went too fast.
- I did not notice that you achieved the failure by modifying Makefiles, so we need to modify the others (and/or codegen)
- I actually also did not notice that there is a bug in your patch, now it fails all the time if REQUIRE_CUDA is set, even if nvcc is there ;-)

I have made a few changes in my fork of your branch
https://github.com/valassi/madgraph4gpu/tree/CI_vGPU
If you agree with them, can you merge them into yours an dmerge to master?
(Or I can replace the PR by another PR on my branch)

In my patch I fix both issues, the failure is only if nvcc is not there, and also the patch is in codegen and then in all regenerated processes

Thanks!
Andrea

valassi (commented):
See my other general comment, there were a few issues I fixed in https://github.com/valassi/madgraph4gpu/tree/CI_vGPU

hageboeck:
@valassi Now that cuda drivers and environment issues are taken care of, we are now running successfully the GPU tests. Merging.

valassi:
Very good, thanks!

Issue #439, opened 2022-04-28
Fix heft generation after alphas PR: more fundamentally, remove reading of param_card.dat?


valassi:
Hi @oliviermattelaer @roiser I am finally about to merge the alphas PR #434.

As one last check, I wanted to check that heft models also work fine - they do not.

I have fixed the simple stuff (like _sm vs _heft suffixes), but there is one fundamental issue: the code to compute the dependent couplings depends itself on several parameters that I had not exported there. See below a full example.

The problem here is quite fundamental, in the sense that normally these parameters are available through the Parameters_sm instance, somehow available through the CPPProcess instanc (things that I think we should review anyway). Anyway, the computation is now part of a device function (__host__ __device__ inline const DependentCouplings_sv computeDependentCouplings_fromG) which can be in a kernel. If all those other parameters must be available in a kernel, one must copy them there to device memory, especially if they are read at runtime from a runcard file. The alternative is that they are taken from somewhere hardcoded in the code (where they are sent at generation time - this is what I did in the HARDCODED ifdef branch, so you see that in the full files I attach these are available.

My question is,
- should we continue supporting reading of runcards at runtime?
- or can we assume (as it is done in the fortran code in one step during the make process) that these end up into a hardcoded code file instead?
The latter solution would considerably simply things.

* In practice, I am asking whether we should use my 'hardcoded' branch as default and drop the runtime reading of runcards. *

This is certainly not a priority, but I raise the issue here for reference. I will merge the alphas PR as is, without support for heft...

Thanks!
Andrea

PS Example from my autogenerated code
```
namespace Parameters_heft_dependentCouplings
{
  constexpr size_t ndcoup = 1; // #couplings that vary event by event because they depend on the running alphas QCD
  constexpr size_t idcoup_GC_13 = 0;
  struct DependentCouplings_sv
  {
    cxtype_sv GC_13;
  };
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wunused-variable"  // e.g. <<warning: unused variable ‘mdl_G__exp__2’ [-Wunused-variable]>>
#pragma GCC diagnostic ignored "-Wunused-parameter" // e.g. <<warning: unused parameter ‘G’ [-Wunused-parameter]>>
#ifdef __CUDACC__
#pragma nv_diagnostic push
#pragma nv_diag_suppress 177 // e.g. <<warning #177-D: variable "mdl_G__exp__2" was declared but never referenced>>
#endif
[...]

PR #446, opened 2022-05-06
partial fix for EFT processes after merging alphas (only workd with HRDCOD=1 for now)

valassi:
This is a WIP PR for #439.

Now it is possible to build heftggh after the alphas merge, but ONLY in the hardcoded HRDCOD=1 mode.

The problem with the non hardcoded mode is that several EFT parameters like mdl_MH__exp__6 which are needed in the calculation of couplings from alphas are not available in the G2COUP kernel. Various options
- force the use of HRDCOD=1 (this works, because the parameters are constexpr in a namespace)
- compute the Gs on the host rather than on the device (just get mdl_MH__exp__6 from the singleton Parameters class on the host)
- (probably the best, but most complex) copy frm host to device the parameters like mdl_MH__exp__6 from the singleton Parameters class

Note that the last one has some overlap with what we do with cIPC/cIPD. Maybe easiest to generalize the cIPC/cIPD to all parameters? Overkill for the other cases? In any case the amount of global memory (or even constant memory) these would occupy is very little.


valassi:
In a way the question is, what determines what is in self.parms2order and self.coup2order? These are the only pieces of the Parameters.h file that are presently copied to constant memory...

PR #450, opened 2022-05-10
Additional fixes for EFT gg->h in clang

valassi:
This contains two small fixes for EFT in clang:
- a followup to #446 and #449: in clang, there is no bracket operator [] returning a reference in SIMD vectors, so I had to modify the patch for float vectors in EFT
- a followup to #358: I added a small patch using constexpr to remove a build warning if nparf=1, ie for 2 to 1 processes


valassi:
All checks passed, self merging

And thanks @hageboeck for fixing the CI! ;-)

Issue #373, opened 2022-02-04
Running of alphas (QCD coupling)


valassi:
Hi @roiser @oliviermattelaer I dumo here a few questions/ideas that we just discussed with Stefan.

Olivier can you please correct me if I am wrong? :-)

1) First point, what is this about and what this implies in the calculations

(Note: focus on the default stream, reading from cards. I implemented an optional 'hardcoded' stream too, but it may be incomplete, or in any case its not the main point).

Currently
- We currently have one aS value that is a constant 0.118 (read from card or hardcoded in Parameters.h).
- There are some other parameters like G, sqrtaS etc which depend on aS (in Parameters again, either from card or hardcoded constexpr calculations)
- There are then model couplings like GC10, GC11, which depends on all these parameters aS, G, sqrtaS etc
- CPPProcess in an init phase propagates the model parameters GC10, GC11 to the constants cIPC (CUDA constant memory, or static c++ memory... not even sure if there is some hardcoding done here actually?)
- CPPProcess in every event calculation passes the appropriate cIPC[i] constants to the FFV functions
- The FFV functions assume that each coupling is a constant: even if FFV in vector C++ or in CUDA is able to handle several events at the same time, the same constant couplings are always used

What we need because of physics:
```
aS = aS0 * f(Q2)
```

In practice:
- The valuse of aS to be used is different from one event to another: the aS0=0.118 constant needs to be multiplied by a factor f(Q2) which depends on the "scale" Q2.
- The scale Q2 actually depends on the momenta of the external particles (and more? I will come back to this later)
- This means that the values of G, sqrtaS etc should also depend event by event
- This means that GC10, GC11 etc and the cIPC should also depend event by event
- This means that the couplings that enter the FFV should also depend event by event

2) Knowing the running value of aS - i.e knowing the factor f(Q2), how can this be propagated to ME calculations?

2a) The part before FFV: keep the constants as they are to represent as0?

Changing the existing Parameters and cIPC infrastructure would be too complex, and would probably be a wrong idea. I think that I would keep the existing stuff up until the cIPC, which are still constants and can still in constant CUDA memory and static const c++ memory. Essentially the cIPC constants represent the values of the constants based on the reference value aS0.

One would then a function (in Parameters? something that can be easily code generated!) that describes how each of the parameters would change.
- Luckily all of these are multiplicative factors with a given exponent?
[...]

PR #452, opened 2022-05-11
Basic fix to allow ggtt.mad builds with running alphas + Fixes for parallel "make -j"

valassi:

Issue #417, opened 2022-04-04
Physics results are different in fortran and cudacpp MEs (by factors or orders of magnitude)


valassi:
I am progressing in the integration of madevent+cudacpp (#400 and beyond, PR #401 and beyond).

The main issue I am facing now is that the MEs per event are different in fortran and in cuda/cpp. I am using the same type of debugging Stefan had used in his earlier tests, just caculate OUT with Fortran and OUT2 with cuda/cpp and print out the ratio.

In gg_tt, the results were wildly different. The ratios were huge: event number (in iteration with 16), OUT, OUT2, ratio
```
           1   2.7179477360093959E-006   3.8048997495080039        1399916.4513349019     
           2   2.4038671516498269E-003  0.49642094931550251        206.50931103858960     
           3   3.8625573348408305E-005   1.0461548930420270        27084.514282948174     
           4   1.5598005763660177E-003  0.44890337193428720        287.79536226363660     
           5   2.1373159576718779E-005  0.62881610141698518        29420.830325055824     
           6   3.1008689435420177E-005  0.54918963982598723        17710.830409964587     
           7   2.9484864958814827E-003  0.57067969683783715        193.55004597612245     
           8   6.2609580619723756E-004   1.0769120999539372        1720.0436247846069     
           9   6.0233031303646806E-005  0.47369098861178799        7864.3059855947913     
          10   1.6038037166377266E-006   1.1317896927132225        705690.90280321101     
          11   1.0426826876806022E-007   2.1018860475392533        20158443.909860998     
          12   1.1518358808597667E-005   2.5083158683912758        217766.77650630136     
          13   1.1234114251542622E-005  0.98776012624391241        87925.056139452849     
          14   2.4224907882032997E-003  0.51894860539999155        214.22108514389134     
          15   2.9500647392138135E-008   3.0360682133944707        102915308.02824269     
          16   6.2546173728503316E-004  0.32241178922449132        515.47803807151672     
           1   1.7029951282941172E-003  0.53612877811686066        314.81521538696268     
           2   2.9291733453278772E-006   1.2607876351277905        430424.38479743304     
           3   4.9347016092575566E-004  0.32938191546612056        667.48091687691169     
           4   1.6280666234303769E-004  0.49474672397459363        3038.8604302454773     
           5   1.4736053161971475E-006   1.1342960426203157        769742.09454369463     
           6   6.9846696062340651E-006  0.80332578412761235        115012.71060990711     
[...]

PR #453, opened 2022-05-11
Enable properly running alphas again (variable renormalization scale) + Makefile fixes

valassi:
This is a PR for #373 and specifically #417.

We can now finally use the full running alphas machinery and get the same physics results in fortran and cudacpp

It also includes a few bug fixes in Makefiles, for bugs I introduced in the previous PR #452


valassi:
All checks have passed, I am self merging. This closes #373 and the whole saga about running alphas.

I keep #417 open as I want to develop an infrastructure for easier comparisons of physics results (and of performance) for fortran and cudacpp.

PR #457, opened 2022-05-18
Simplify color algebra (A-iB)(M)(A+iB) as AMA+BMB

valassi:
This is a spinoff of #155 about color algebra optimisations.

While discussing with @roiser we realised that one can rewrite (A-iB)(M)(A+iB) as AMA+BMB in the color algebra calculation (where A,B are vectors and M is a matrix, with ncol and ncol*ncol dimensions).

This will in any case simplify tensor core calculations (as there is no need to deal with complex arithmetics, one can just use two real quadratic forms).

I was also naively expecting that this might improve performance by a factor 2 or 4, because some calculations are not needed. It turns out that the physics results are the same, but performance is excatly the same.

I would imagine that this may be because in the older implementation the c++ compiler was in any case optimising away some calculations, refusing to compute calculations whose results are not stored? Essentially the change is this
https://github.com/madgraph5/madgraph4gpu/blob/6f5688bec17f6e88f80dee0d8c08479e72526d51/epochX/cudacpp/gg_ttgg/SubProcesses/P1_Sigma_sm_gg_ttxgg/CPPProcess.cc#L1942
```
      // Sum and square the color flows to get the matrix element
      // (compute |M|^2 by squaring |M|, taking into account colours)
      fptype_sv deltaMEs = { 0 }; // all zeros https://en.cppreference.com/w/c/language/array_initialization#Notes
      for( int icol = 0; icol < ncolor; icol++ )
      {
        cxtype_sv ztemp_sv = cxzero_sv();
        for( int jcol = 0; jcol < ncolor; jcol++ )
          ztemp_sv += cf[icol][jcol] * jamp_sv[jcol];
        // OLD implementation (slower?)
        //deltaMEs += cxreal( ztemp_sv * cxconj( jamp_sv[icol] ) ) / denom[icol];
        // NEW implementation (faster?) 
        // Rewrite the quadratic form (A-iB)(M)(A+iB) as AMA - iBMA + iBMA + BMB = AMA + BMB!
        deltaMEs += ( cxreal( ztemp_sv ) * cxreal( jamp_sv[icol] ) +
                      cximag( ztemp_sv ) * cximag( jamp_sv[icol] ) ) / denom[icol];
      }
```

This is almost a no-op but I would merge it anyway. Maybe I will first rerun a full set of tests just for good habit.

I only tested one log and essentially there is no difference with upstream/master:
https://github.com/madgraph5/madgraph4gpu/commit/ed0890b46b2f26530eb0eb7a8b56c6efc6e76d72

### DIFF
--- a/epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/madgraph/iolibs/template_files/gpu/process_matrix.inc
+++ b/epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/madgraph/iolibs/template_files/gpu/process_matrix.inc
@@ -11,7 +11,11 @@
         cxtype_sv ztemp_sv = cxzero_sv();
         for( int jcol = 0; jcol < ncolor; jcol++ )
           ztemp_sv += cf[icol][jcol] * jamp_sv[jcol];
-        deltaMEs += cxreal( ztemp_sv * cxconj( jamp_sv[icol] ) ) / denom[icol];
+        // OLD implementation: why is this not slower? maybe compiler does not compute imaginary part of "ztemp_sv*cxconj(jamp_sv[icol])"?
+        //deltaMEs += cxreal( ztemp_sv * cxconj( jamp_sv[icol] ) ) / denom[icol];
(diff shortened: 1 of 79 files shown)


valassi:
All tests pased, I am self merging

Issue #461, opened 2022-05-20
Understand performance difference in cudacpp between SA and madevent


valassi:
This is a followup to PR #454 where I am doing some first tests (with logs) of madevent+cudacpp MEs, comparing physics and performance to madevent and to cudacpp.

I see that the cudacpp ME calculation is faster in standalone (bridge!) mode than in madevent. IN principle I thought the standalone bridge mode was exactly equivalent? 
- There is most likely an issue with helicity filtering, which should be excluded from the measurements
- But there may be more than just that, to be understood
- (Note that in principle both are using the same CUDA grids, so any slowdown - #460 - from using too small CUDA grids should not be here)

See for instance
https://github.com/madgraph5/madgraph4gpu/blob/d1fd8b201803b366579f925017f09a85001ce9f8/epochX/cudacpp/tmad/logs_ggttggg_mad/log_ggttggg_mad_d_inl0_hrd0.txt
```
DATE: 2022-05-20_15:00:32

Working directory: /data/avalassi/GPU2020/madgraph4gpuX/epochX/cudacpp/gg_ttggg.mad/SubProcesses/P1_gg_ttxggg

*** EXECUTE MADEVENT (create results.dat) ***
 [XSECTION] Cross section = 80.03
 [COUNTERS] PROGRAM TOTAL          :    4.2070s
 [COUNTERS] Fortran Overhead ( 0 ) :    0.2606s
 [COUNTERS] Fortran MEs      ( 1 ) :    3.9463s for       96 events => throughput is 2.43E+01 events/s

*** EXECUTE CMADEVENT_CUDACPP (create results.dat) ***
 [XSECTION] Cross section = 80.03
 [MERATIOS] ME ratio CudaCpp/Fortran: MIN = 0.99997121 = 1 - 2.9e-05
 [MERATIOS] ME ratio CudaCpp/Fortran: MAX = 1.00014921 = 1 + 0.00015
 [MERATIOS] ME ratio CudaCpp/Fortran: NENTRIES =     96
 [MERATIOS] ME ratio CudaCpp/Fortran - 1: AVG = 1.3e-05 +- 2.9e-06
 [COUNTERS] PROGRAM TOTAL          :    4.2596s
 [COUNTERS] Fortran Overhead ( 0 ) :    0.2727s
 [COUNTERS] Fortran MEs      ( 1 ) :    3.5988s for       96 events => throughput is 2.67E+01 events/s
[...]

PR #463, opened 2022-05-22
Remove helicity filtering from cudacpp ME timers in (c/g)madevent

valassi:
This is a PR addressing #461 - I observed a difference in cudacpp ME throughputs when computed in standalone mode (even in bridge mode) and within cmadevent/gmadevent.

The largest part of the difference (a factor 2, because I only do one iteration in these tests) is due to the fact that I was not removing helicity filtering in the cudacpp timers. So Fortran MEs includes only one ME calculation, while cudacpp includes two (the first one for helicity filtering, the second one for real ME calculation).

This is now fixed in this PR.


valassi:
All tests have passed, I am self merging.

I closed issue #461. Note that I opened #464 to further improve timers in SA bridge mode, but I consider that really low priority.

Issue #342, opened 2022-01-21
multi-channel implementation in cpp/cuda class


oliviermattelaer:
Hi guys,

I have a code that can output the multi-channel information inside the "basic" gpu class.
The code (for gg>tt~) looks like this:

```
VVV1P0_1(w[0],w[1],cxtype(cIPC[0],cIPC[1]),0.,0.,w[4]);
# Amplitude(s) for diagram number 1
FFV1_0(w[3],w[2],w[4],cxtype(cIPC[2],cIPC[3]),&amp[0]);
if(channel_id == 1){multi_chanel_num += conj(amp[0])*amp[0];};
 multi_chanel_denom += conj(amp[0])*amp[0];};
jamp[0] += +cxtype(0,1)*amp[0];
jamp[1] += -cxtype(0,1)*amp[0];
FFV1_1(w[2],w[0],cxtype(cIPC[2],cIPC[3]),cIPD[0],cIPD[1],w[4]);
# Amplitude(s) for diagram number 2
FFV1_0(w[3],w[4],w[1],cxtype(cIPC[2],cIPC[3]),&amp[0]);
if(channel_id == 2){multi_chanel_num += conj(amp[0])*amp[0];};
 multi_chanel_denom += conj(amp[0])*amp[0];};
jamp[0] += -amp[0];
FFV1_2(w[3],w[0],cxtype(cIPC[2],cIPC[3]),cIPD[0],cIPD[1],w[4]);
# Amplitude(s) for diagram number 3
FFV1_0(w[4],w[2],w[1],cxtype(cIPC[2],cIPC[3]),&amp[0]);
if(channel_id == 3){multi_chanel_num += conj(amp[0])*amp[0];};
 multi_chanel_denom += conj(amp[0])*amp[0];};
jamp[1] += -amp[0];
```

@valassi @roiser, I was hesitating to use std:norm(amp[0]). Any preference here?
(I still have to edit template to initialize/use such variable but this way the impact on memory should be quite minimal.

valassi:
Hi @oliviermattelaer thta's great news! :-)

I see how thic connects to the MadEvent algorithm, nice and easy actually.

About std::norm, I would avoid that for the moment. I imagine those amp[0] are cxtype_sv if I remember well. The way you write this explicitly should ensure they are properly vectorized (computed in parallele for the 4 events in one cxtype_v vector), while with std norm I am not sure.

[...]

Issue #419, opened 2022-04-04
Debug fortran vs cpp differences in multichannel (using SD1 to bypass "get_channel_cut" issues and LIMHEL=0 to bypass helicity filtering issues)


valassi:
I am debugging the fact that physics results are different in fortran and cudacpp, #417

I have found that the huge differences that I had initially found come from the single diagram enhancement factor in multichannel. I now realised, however, that a proper implementation in cudacpp of the enhancement factor needs to take into account the cuts. This is matrix1.f https://github.com/valassi/madgraph4gpu/blob/be0ad118905d9d8fa6f584b0b50c2f49d31c7c22/epochX/cudacpp/gg_tt.mad/SubProcesses/P1_gg_ttx/matrix1.f#L253
```
      IF (MULTI_CHANNEL) THEN
        XTOT=0D0
        DO I=1,LMAXCONFIGS
          J = CONFSUB(1, I)
          IF (J.NE.0) THEN
            IF(SDE_STRAT.EQ.1) THEN
              AMP2(J) = AMP2(J) * GET_CHANNEL_CUT(P, I)
              XTOT=XTOT+AMP2(J)
            ELSE
              AMP2(J) = GET_CHANNEL_CUT(P, I)
              XTOT=XTOT+AMP2(J)
            ENDIF
          ENDIF
        ENDDO
        IF (XTOT.NE.0D0) THEN
          ANS=ANS*AMP2(CHANNEL)/XTOT
```

The get_channel_cut function seems to be a relatively simple fortran function, and of course one could call it from the cudacpp if needed (well, it is not vectorized, but ok). Bit maybe this is not needed at all, at least for the main use cases?

@oliviermattelaer do we need get_channel_cut in cudacpp? (By the way before we duplicate effort, do you already have a multichannel patch in cudacpp?)

Thanks
Andrea

oliviermattelaer:
No we do not need that function.
This should be used only if sde_strategy=2 in the run_card.
(and for the moment we should focus on sde_strategy=1)

Cheers,

Olivier

valassi:
I am making progress in multichannel #342 in PR #465

I do see however large differences in ME in fortran and cudacpp on each event.

I would like to debug this and understand what causes it, but I find again this get_channel_cut. Maybe we do need it?

Otherwise it may be a bit the way the loops are done (eg sum over helicities before or after computing numerators and denominators, or it should not make a difefrence?). Or maybe the channel numbering is different (unlikely)?

[...]

PR #465, opened 2022-05-22
First patch to integrate multichannel (single diagram enhancement) into cudacpp

valassi:
This is a WIP PR for issue #342, to integrate multichannel (single diagram enhancement) into cudacpp

Status
- I would say that I have picked what I needed to pick up from Olivier's code (am I firgetting something?)
- I have started to develop on that, but I still have two main issues
- One, channel_id is used in calculate_wavefuntion and sigmakin, but it does not appear higher up (eg not even in the sigmakin signature), so I started to pass the parameter in sigmaking, an dI need to pass it to the calling functions
- Two, the numerators and denominators are currently allocated for a single event or SIMD event page, but I need to allocate them for all events at the same time. This is because sigmakin an dcalculate wavefunctions actually manipulate a full set of nevt events at the same time, so there is no other way than having those numerators and denominators in two large arrays (which them will also need to be passed through sigmakin). Technically, one could still simplify the calling sequences of sigmakin, so that for cpp it only handles one event page, and the loop over event pages is done outside. However, at some point we will move to splitting kernels, and in that case I think (should cross check) that we will need two arrays for numerators and denominators anyway... so probably best to add them immediately.


valassi:
I have fixed both issues in ggtt.mad
- I have arrays for numerators and denominators
- I have percolated them to the internal API
- I have percolated channelID to teh Bridge API and into Fortran

**TODO: check with @oliviermattelaer if channelId in Fortran is 1 to N (N diagrams) and if 0 can be used to mean "disable multichannel**

Also todo: code generation

valassi:
Ok I have also fixed code generation. I am rerunning all tests

valassi:
Just a status update, to kno wehere to resume from next week:
- Basic integration of multichannel, backport to codegen and regeneration of all processes is there
- Now debugging in ggtt.mad the diffreences in multichannel mode between fortran and cudacpp, mainly #419 
- Using SDE=1 helps but is not enough. One needs to set LIMHEL=0 in Fortran (or set an equivalent threshold in cudacpp) to get again good agreement between fortran and cudacpp. An alternative is to do helicity filtering in fortran and pass the list of good helicities. For the moment, I would stick to LIMHEL=0 in Fortran (it will be easy to change later on if needed, but I want to have everything internally consistent)

To do specifically for this PR before merging, for multichannel
- Backport to codegen and regenerate the othre processes with SDE=1 and LIMHEL=0. Run tmad and check that all processes give good agreement (say E-3 max discrepancy) between fortran and cudacpp MEs

Maybe in this PR, maybe better one immediately afterwards, for multichannel:
- Check integration in the bridge API of random choices of color/helicity (fake it if needed), with the goal of completely replacing Fortran MEs by cudacpp. Calculate cross sections based on cudacpp MEs. Check that we get the same xsec with fortran and cudacpp MEs (say within E-3). Check the performance speedups. This can be a first alpha release, with real physisc results, faster.
-  One tiny detail before doing the above: check why gmadevent and cmadevent give different xsecs (even now with fortran MEs). I think it is because the grid sizes are different (hm so what? should not mak ea difference, but it does... hopefully not a different number of helicities in fortran depening on grid size?)

Later on, beyond multichannel, for first unweighted generation for the experiments
- Random choice of color and random choice of helicity. Implementation, bridge API, tests etc.
- Numerical precision: I would still investigate these E-3 deviations between fortran and cudacpp before giving this to the experiments.
- Packaging of cudacpp: agree how to package this (but we seem to have agreed with Olivier that it can be a plugin).
- More complex: backport of mychanges to fortran code generation. Now I am using a set of patches over generated code. This should becomes instead upstream code-generating code. Review the fortran code first with Olivier and Stefan, then port to the code generator for fortran.
- Ah, for unweighted generation: fix the "fail to reduce to color indices" issues. Otherwise ggttg/ggttgg never generate lhe files.
- Along the way (here for unweighted egneration, or even before for xsec alone): launch Mg5amc as  user would do, with the complete setup, survey refine etc and auto updatin browser pages, using cudacpp MEs. Check that all goes through. Eventually, check lhe files event by event, in fortran and in cudacpp.
- (Any other bells and whistles for the final packaging? I cannot think of anything else now)

valassi:
The last changes so far were about issue #473: code generation in standalone cudacpp was failing. Eventually, I had to go for two separate code bases (as in Olivier's original standalone gpu): standalone cudacpp and madevent+cudacpp. Issue #473 is fixed by this PR #465.

Conversely, note that issue #466 will not be fixed here, it will be in a later PR.

This is now essentially ready to be merged...

valassi:
Before merging, I reply to my own comments from https://github.com/madgraph5/madgraph4gpu/pull/465#issuecomment-1135872107 from two weeks ago

> Just a status update, to kno wehere to resume from next week:
> 
>     * Basic integration of multichannel, backport to codegen and regeneration of all processes is there
> 
>     * Now debugging in ggtt.mad the diffreences in multichannel mode between fortran and cudacpp, mainly [Debug fortran vs cpp differences in multichannel (using SD1 to bypass possible issues with "get_channel_cut") #419](https://github.com/madgraph5/madgraph4gpu/issues/419)
> 
>     * Using SDE=1 helps but is not enough. One needs to set LIMHEL=0 in Fortran (or set an equivalent threshold in cudacpp) to get again good agreement between fortran and cudacpp. An alternative is to do helicity filtering in fortran and pass the list of good helicities. For the moment, I would stick to LIMHEL=0 in Fortran (it will be easy to change later on if needed, but I want to have everything internally consistent)

Ok the above was there already two weeks ago

> To do specifically for this PR before merging, for multichannel
> 
>     * Backport to codegen and regenerate the othre processes with SDE=1 and LIMHEL=0. Run tmad and check that all processes give good agreement (say E-3 max discrepancy) between fortran and cudacpp MEs

This is now done
- codegen and all 5 .mad procxesses have been regenerated with SDE=1 and LIMHEL=0
- tmad tests have been done for the 5 processes (with rough settings), eg https://github.com/valassi/madgraph4gpu/blob/mch/epochX/cudacpp/tmad/logs_ggttggg_mad/log_ggttggg_mad_d_inl0_hrd0.txt
- the agreement is indeed within the percent or permille level, so the factors or orders of magnitude differences seen in #417 and #419 are now gone and these issues can be closed: I opened a separate #476 to futher investigate the remaining issues at the percent or permille level

In addition
- integrating multichannel in madevent+cudacpp was breaking code generation for standalone cudacpp: this is now fixed (#473), but there are two separate code bases (which I plan to clean up in the repo, #478)

> Maybe in this PR, maybe better one immediately afterwards, for multichannel:
> 
>     * Check integration in the bridge API of random choices of color/helicity (fake it if needed), with the goal of completely replacing Fortran MEs by cudacpp. Calculate cross sections based on cudacpp MEs. Check that we get the same xsec with fortran and cudacpp MEs (say within E-3). Check the performance speedups. This can be a first alpha release, with real physisc results, faster.
> 
>     * One tiny detail before doing the above: check why gmadevent and cmadevent give different xsecs (even now with fortran MEs). I think it is because the grid sizes are different (hm so what? should not mak ea difference, but it does... hopefully not a different number of helicities in fortran depening on grid size?)
> 
> 
> Later on, beyond multichannel, for first unweighted generation for the experiments
> 
>     * Random choice of color and random choice of helicity. Implementation, bridge API, tests etc. (I think Olivier is working on some aspects of this in the fortran side)
> 
>     * Numerical precision: I would still investigate these E-3 deviations between fortran and cudacpp before giving this to the experiments.
> 
>     * Packaging of cudacpp: agree how to package this (but we seem to have agreed with Olivier that it can be a plugin).
> 
>     * More complex: backport of mychanges to fortran code generation. Now I am using a set of patches over generated code. This should becomes instead upstream code-generating code. Review the fortran code first with Olivier and Stefan, then port to the code generator for fortran.
> 
>     * Ah, for unweighted generation: fix the "fail to reduce to color indices" issues. Otherwise ggttg/ggttgg never generate lhe files.
> 
>     * Along the way (here for unweighted egneration, or even before for xsec alone): launch Mg5amc as  user would do, with the complete setup, survey refine etc and auto updatin browser pages, using cudacpp MEs. Check that all goes through. Eventually, check lhe files event by event, in fortran and in cudacpp.
> 
>     * Probably, move to branch 340 and github-based mg5amc before the final prerelease?
> 
>     * (Any other bells and whistles for the final packaging? I cannot think of anything else now)

These other points are essentially not done.

Issue #474, opened 2022-06-03
Move code generation upstream reference from bazaar launchpad to github (both on the 3.1.1 branch)


valassi:
Thanks to Olivier the github is now available
https://github.com/mg5amcnlo/mg5amcnlo/tree/3.1.1_lo_vectorization

I should move the upstream reference for code generation from bazaar/launchpad to the above github

valassi:
The next step will be #479: move from the 311 branch to the 340 branch

valassi:
This is implemented in my codegen by PR #483 using Olivier's  https://github.com/mg5amcnlo/mg5amcnlo/tree/3.1.1_lo_vectorization . Closing.

PR #482, opened 2022-06-09
Add gridpack generation with 311 (bazaar) branch

valassi:
After alpaka, this PR is another preliminary step to #474. Before moving from bazaar to git, I want to hav eall generatiopn workflows (cudacpp, alpaka and now gridpack) on the 311 branch. This will allow me to drop all support for the old 28x and 29x bazaar branch in the codegen scripts.


valassi:
All checks passed, I will self merge - I am now only using bazaaar latest 311 branch (no 270 branch left). Next step is moving to github instead.

PR #483, opened 2022-06-10
Move codegen from bazaar to github

valassi:
This is the PR for #474. I am dropping bazaar in codegen and moving to github. All looks good.


valassi:
This is complete, closing #474. Self merging now.

Issue #480, opened 2022-06-08
clash between 'makefile' and 'Makefile' on case insensitive systems


oliviermattelaer:
Updating files: 100% (12406/12406), done.
warning: the following paths have collided (e.g. case-sensitive paths
on a case-insensitive filesystem) and only one from the same
colliding group is in the working tree:
  'epochX/cudacpp/ee_mumu.mad/SubProcesses/Makefile'
  'epochX/cudacpp/ee_mumu.mad/SubProcesses/makefile'
  'epochX/cudacpp/ee_mumu.mad/SubProcesses/P1_ll_ll/Makefile'
  'epochX/cudacpp/ee_mumu.mad/SubProcesses/P1_ll_ll/makefile'
  'epochX/cudacpp/gg_tt.mad/SubProcesses/Makefile'
  'epochX/cudacpp/gg_tt.mad/SubProcesses/makefile'
  'epochX/cudacpp/gg_tt.mad/SubProcesses/P1_gg_ttx/Makefile'
  'epochX/cudacpp/gg_tt.mad/SubProcesses/P1_gg_ttx/makefile'
  'epochX/cudacpp/gg_ttg.mad/SubProcesses/Makefile'
  'epochX/cudacpp/gg_ttg.mad/SubProcesses/makefile'
  'epochX/cudacpp/gg_ttg.mad/SubProcesses/P1_gg_ttxg/Makefile'
  'epochX/cudacpp/gg_ttg.mad/SubProcesses/P1_gg_ttxg/makefile'
  'epochX/cudacpp/gg_ttgg.mad/SubProcesses/Makefile'
  'epochX/cudacpp/gg_ttgg.mad/SubProcesses/makefile'
  'epochX/cudacpp/gg_ttgg.mad/SubProcesses/P1_gg_ttxgg/Makefile'
  'epochX/cudacpp/gg_ttgg.mad/SubProcesses/P1_gg_ttxgg/makefile'
  'epochX/cudacpp/gg_ttggg.mad/SubProcesses/Makefile'
  'epochX/cudacpp/gg_ttggg.mad/SubProcesses/makefile'
  'epochX/cudacpp/gg_ttggg.mad/SubProcesses/P1_gg_ttxggg/Makefile'
  'epochX/cudacpp/gg_ttggg.mad/SubProcesses/P1_gg_ttxggg/makefile'

I guess we have to change those name to _fortran _cuda instead

valassi:
[...]

Issue #413, opened 2022-04-01
Should we cleanly separate madevent and cudacpp directories or not?


valassi:
Hi @oliviermattelaer, @roiser, all, this is a part of #400 (makefiles for madevent + cudacpp), but I prefer to split it out as a separate question because it is quite an important decision.

The issue is the following: with the mechanism that I think we are currently targeting,
```
output madevent <dir> --vector_size=16 --me_exporter=standalone_cudacpp
```
the code generator is currently generating the usual madevent, and then the cudacpp ON TOP OF THAT. 

This is not even completely precise. because the standalone cudacpp would create a P1 sudirectory with "Sigma_sm" in the name, while as ME exporter that name disappears. See for instance
https://github.com/valassi/madgraph4gpu/tree/v311/epochX/cudacpp/gg_tt/SubProcesses/P1_Sigma_sm_gg_ttx
https://github.com/valassi/madgraph4gpu/tree/v311/epochX/cudacpp/gg_tt.mad/SubProcesses/P1_gg_ttx

Anyway, the point is, a typical madevent+cudacpp directory, such as
https://github.com/valassi/madgraph4gpu/tree/v311/epochX/cudacpp/gg_tt.mad/
now contains these subdirectories, amongst others:
- Source, with madevent Source
- src, with cudacpp src
- Subprocess, with madevent+cudacpp Subprocess SUPERPOSED
- Subprocess/P1_gg_ttx, with madevent+cudacpp SUPERPOSED (as mentioned above, cudacpp is what would be in SubProcesses/P1_Sigma_sm_gg_ttx)

Specifically for issue #400, this means that there are now in SubProcesses/P1_gg_ttx two make-files:
- "makefile" is that from madevent
- "Makefile" is that from cudacpp
I actually like having two sets of make-files. I will expand on this below.

Anyway, the QUESTION I want to ask is the following:
- OPTION 1, do we want to keep it like this, with source files from madevent/fortran and from cudacpp mixed in the same directries?
- OPTION 2, or do we want to actually have cleanly separate directories for madevent and cudacpp source code?...
(And if we go for option 1. do we even want to go one step further and merge Source and src?)

FOR THE MOMENT, my working assumption is to go with option 1. But we should agree on what seems a most reasonable directory structure, eventually.

[...]

Issue #400, opened 2022-03-10
Prepare Makefile to combine MadEvent + Cudacpp plugin


valassi:
Hi @oliviermattelaer I am having a look at trying to combine MadEvent and the cudacpp plugin.

I am using for instance
```
generate g g > t t~
output madevent CODEGEN_mad_gg_tt --vector_size=16 --me_exporter=standalone_cudacpp
```
This is the infrastructure you created that contains BOTH the fortran code and the cudacpp code.

As a reference, I am using the madevent only
```
generate g g > t t~
output madevent CODEGEN_mad_gg_tt --vector_size=16
``` 

Can you please remind me already how to build this? I imagine I need to do
```
cd Source
make
cd ../SubProcesses/P1_gg_ttx
make
```
but the first step fails with
```
../bin/madevent treatcards run
launch in debug mode
...
  File "/data/avalassi/GPU2020/madgraph4gpuX/epochX/cudacpp/gg_tt.madonly/bin/internal/common_run_interface.py", line 3752, in set_configuration
    config_file = open(config_path)
FileNotFoundError: [Errno 2] No such file or directory: '/data/avalassi/GPU2020/MG5aMC/3.1.1_lo_vectorization/input/mg5_configuration.txt'
/cvmfs/sft.cern.ch/lcg/releases/gcc/10.2.0-c44b3/x86_64-centos7/bin/gfortran -w -fPIC  -ffixed-line-length-132  -c -o setrun.o setrun.f
setrun.f:78: Error: Can't open included file 'run_card.inc'
make: *** [setrun.o] Error 1
```

Thanks
Andrea

valassi:
Thanks Olivier!
I am using the 3.1.1_lo_vectorization branch.
Andrea

oliviermattelaer:
So the weird part is why you do not have the file:
/data/avalassi/GPU2020/MG5aMC/3.1.1_lo_vectorization/input/mg5_configuration.txt
Can you do
```
touch /data/avalassi/GPU2020/MG5aMC/3.1.1_lo_vectorization/input/mg5_configuration.txt
```
This file can technically be empty so this should fix the issue. (in principle it should be created the first time you run the code such that you can edit it without having to worry about commiting the change)

(my question was on the madgraph4gpu side which branch you use)

valassi:
Hm wait... there may be something wrong in my scripts... my bad, sorry!

valassi:
Now it built fine... let m edebug it
[...]

PR #484, opened 2022-06-13
Rename cudacpp Makefile as cudacpp.mk

valassi:
This is a PR to fix the clash between 'makefile' and 'Makefile' on case insensitive systems #480


valassi:
I have closed the following three issues related to this
- #480, the original motication, the clash between Makefile and makefile on case insensitive systems
- #400, the need to integrate cudacpp builds inside the madevent makefile: now the madevent makefile internally calls cudacpp.mk (previously it was already working, but calling Makefile, now at least it is a bit cleaner)
- #413, the question whether we should keep madevent and cudacpp files in the same P1 or in two different directories: eventually we mentioned we might like the second option, but for the moment everything is implemented using the first option (in particular, the need to avoid the makefile/Makefile clash only appears if the same directory is used).

All tests are passing, I will self merge

Issue #476, opened 2022-06-05
Physics results are different in fortran and cudacpp MEs at the ~permille level: different handling of top quark width


valassi:
I am stripping off as a separate issue the fine grained comparison of physics results at the E-4 to E-2 level that I am observing in ggtt* processes (in eemumu I think I get E-8 agreement if not better).

This is a followup of #417, but also #419, where the differences by factors or orders of magnitude are discussed (eg alphas, multichannel) are discussed. I will close those soon. Instead the fine grained comparison remains as a separate issue.

valassi:
I have debugged this in PR #490.

I intially thought this was a QCD related problem, such as alphas related. Instead it is due to the handling of the top quark width:

-    fortran MEs use a top quark width hardcoded to ZERO
-    cudacpp MEs use a non-zero top-quark width

The analysis is still WIP. But we also need to understand how to handle this...

@oliviermattelaer do you have a suggestion? Is there a switch in code generation to treat the top quark as width-0? Thanks

oliviermattelaer:
Excellent, Good catch.

Yes indeed the width has a couple of "dark magic" handling at LO which are set directly within matrix.f so far:
I do remember two of such treatment:
 - set the width of T-channel to zero (new since 2.8.x): the one problematic here
 - special handling of small width (less than 1e-6 times the mass)
and likely in the future it will be a third one 
 - remove onshell resonance within the matrix-element (but that's in the future )

The first one is related to a flag "zerowidth_tchannel" so one can turn it off/on via the command
set zerowidth_tchannel T/F
(before the generate command)

The second is related to the run_card parameter
set small_width_treatment X # where X is the threshold value


For the moment, none of them are critical in term of physics and for our target, so the best is likely to add those to the todo list of things to handle. and for the comparison, generate the code with "set zerowidth_tchannel F".

valassi:
Thanks Olivier, I confirm that the fix works! Now I get agreement between cudacpp and fortran in the cross sections in 10+ digits also for ggtt*... a nice milestone! I will close this as it is fixed by PR #490.

I opened https://github.com/oliviermattelaer/mg5amc_test/issues/23 as a reminder that (with low priority) we should also have zero widths in cudacpp if the experiments use this...

Issue #404, opened 2022-03-11
MadEvent final bridge interface: add multichannel, alphas, color index


valassi:
Hi @oliviermattelaer @roiser this is a dump of what I understand we still miss for the final MadEvent bridge interface (thanks OLivier for the chat this morning!)

The current SMATRIX1 call looks like this
https://github.com/madgraph5/madgraph4gpu/blob/34ab2d4571c61712ddff848d7ba740f1a4c311a8/epoch2/fortran-cuda/pp_ttx/SubProcesses/P1_gg_ttx/auto_dsig1.f#L443
```
      DO IVEC=1, NB_PAGE
        CALL SMATRIX1(P_MULTI(0,1,IVEC),
     &	                         hel_rand(IVEC),
     &				 channel,
     &				 out(IVEC),
C       &				 selected_hel(IVEC),
     &				 jamp2_multi(0,IVEC),
     &				 IVEC
     &				 )
       ENDDO
```

In the above
- p_multi is the input momenta
- out is the output ME (or more precisely, ME with a suppression factor that depends on the channel)
- channel is the input channel index
- ivec is simply an event number: it is needed because there is a loop over events (SMATRIX1 uses one event only), but also because the alphas value is read from a fortran common block for every event : so in a way this is the input alphas or G
- jamp2_multi is the output amplitudes for different colors for every event: this is used to determine one color index that is randomly chosen and associated to the output events (NB: IIUC this does not affect the cross section calculation)
- hel_rand and selected_hel, similarly, are used to randomly determine which helicity is written out for every event (NB: again, IIUC this does not affect cross sections)

Reminder to put it in context: the outputs of the whole generation are
- a cross section, which is calculated here with a multi channel integration, so all of momenta, MEs, channel are relevant
- a set of events written to file, which are not only unweighted after computing their MEs, but also are randomly assigned one specific color index and one specific helicity, and both of these depend on those extra inputs/outputs like jamp2 and hel_rand

The bridge interface is currently the following
https://github.com/madgraph5/madgraph4gpu/blob/34ab2d4571c61712ddff848d7ba740f1a4c311a8/epochX/cudacpp/gg_ttg/SubProcesses/fbridge.inc#L34
```
C
C Execute the matrix-element calculation "sequence" via a Bridge on GPU/CUDA or CUDA/C++.
C - PBRIDGE:  the memory address of the C++ Bridge
C - MOMENTA:  the input 4-momenta Fortran array
[...]

Issue #403, opened 2022-03-11
Randomly pick one helicity for each event written to file


valassi:
Randomly pick one helicity for each event written to file

Will give details elsewhere

valassi:
Full details are given in #404.

This is on the critical path to get unweighted event sets to the experiments where each event has one helicity (but it should not be needed for cross section calculations)

oliviermattelaer:
note that https://github.com/oliviermattelaer/mg5amc_test/issues/24
is changing the fortran handling of color/helicity in order to have a clean api 
This means:
- the fortran/cpp api is now changed to the what we discussed a while ago (input: 2 random number, output: 2 integer)
- New fortran function has been added to the code
  - select_color(rcol, jamp2, channel, proc_id,  icol [out]) 
    - [ ] need to write such function within the CPP plugin (work should have started in that direction before the break)
  - get_helicities(..)
    - no need to convert such function to C++

valassi:
I am working on the implementation of this feature.

I realised that an ESSENTIAL prerequisite is to move the loop over npagV from calculate_wavefunction to sigmakin aka issue #415

oliviermattelaer:
Why? I do not see the issue/ the problem.

valassi:
The issue is actually that I need to invert the helicity loop and ipagV loop. To do the selection of a helicity, I need to store for each event all partial sums of MEs up to each helicity. This is TS in the SMATRIX1 Fortran code, which is an array of NCOMB values, because this part is always computed event by event. In cuda, this should not be needed because in any case sigmaKin is a kernel, so all the variables there are local to one thread and to one event. The problem is C++, because sigmaKin is called for all nevt and internally does a loop over "pages" ipagV (i.e. SIMD vectors), or better it delegates the loop over ipagV to calculate_wavefunctions (CW). That is to say, sigmakin does a loop over helicities, calls CW for one helicity, and then CW internally does the loop over event pages. I could still do the random choice of helicity in this approach, but then I need to store NEVT*NCOMB numbers, the partial sums for each helicity for each event, so that sigmaKin can do the choice at the end.

[...]

Issue #402, opened 2022-03-11
Randomly pick one color for each event written to file (jamp2 handling)


valassi:
Randomly pick one color for each event written to file (jamp2 handling)

Will give details elsewhere

valassi:
Full details are given in #404.

This is on the critical path to get unweighted event sets to the experiments.

oliviermattelaer:
note that https://github.com/oliviermattelaer/mg5amc_test/issues/24
is changing the fortran handling of color/helicity in order to have a clean api 
This means:
- the fortran/cpp api is now changed to the what we discussed a while ago (input: 2 random number, output: 2 integer)
- New fortran function has been added to the code
  - select_color(rcol, jamp2, channel, proc_id,  icol [out]) 
    - [ ] need to write such function within the CPP plugin (work should have started in that direction before the break)
  - get_helicities(..)
    - no need to convert such function to C++

valassi:
This is in progress. I have some agorithm but it gives me different results from Fortran and C++ (the last two columns).

Maybe not ideal to use gg_tt for debugging as it has only two colors...
```
  33  0.66637210E-01  0.66637210E-01          1.00000000000  3  3   1   1
  34  0.36652604E-01  0.36652604E-01          1.00000000000  8  8   1   1
  35  0.41960814E-01  0.41960814E-01          1.00000000000 12 12   1   1
  36  0.64245071E-01  0.64245071E-01          1.00000000000  2  2   1   1
  37  0.67260622E-01  0.67260622E-01          1.00000000000 15 15   1   1
  38  0.72644897E-01  0.72644897E-01          1.00000000000 15 15   1   1
  39  0.48924468E-01  0.48924468E-01          1.00000000000 15 15   1   1
  40  0.71085828E-01  0.71085828E-01          1.00000000000  2  2   1   1
  41  0.71529802E-01  0.71529802E-01          1.00000000000 15 15   1   1
  42  0.70385722E-01  0.70385722E-01          1.00000000000  2  2   1   1
  43  0.31575040E-01  0.31575040E-01          1.00000000000 12 12   1   1
  44  0.69272519E-01  0.69272519E-01          1.00000000000 15 15   2   1
  45  0.71009335E-01  0.71009335E-01          1.00000000000 15 15   1   1
  46  0.68831440E-01  0.68831440E-01          1.00000000000  2  2   1   1
  47  0.69539468E-01  0.69539468E-01          1.00000000000  2  2   1   1
[...]

PR #486, opened 2022-06-14
First attempt to produce cross sections and events using cudacpp MEs (debug common blocks behind SMATRIX1: add RESET_CUMULATIVE_VARIABLE call)

valassi:
Hi @oliviermattelaer @roiser 

This WIP PR is a first attempt to produce cross sections and unweighted events using the madevent sampling plus the cudacpp MEs.

A first milestone has been achieved! If I compute BOTH the Fortran and cudacpp MEs (FBRIDGE_MODE=-1), and then I simply replace the Fortran ME by the cudacpp ME, I do get the correct cross sections and numbers of unweighted events (I have not compared event files yet). This was quite trivial to do , I just needed to add one line copying OUT=OUT2 which replaces the fortran ME by the cudacpp ME.

There is still a problem however: if I ONLY compute the cudacpp ME and I skip the Fortran ME calculation (i.e. if I skip SMATRIX1 calls, this is FBRIDGE_MODE=1), then I get several problems: 
- the cross sections in fortran and cudacpp differ (?!)
- the numbers of unweighted events generated differ in fortran and cudacpp
- the unweighted event generation fails with some processes again with 'Error: failed to reduce to color indices', see https://github.com/oliviermattelaer/mg5amc_test/issues/14

TODO:
- fix the issues above, one should be able to only run the cudacpp ME (is one output argument of SMATRIX1 needed that is not computed by the bridge? or is there more fundsamentally some commob block lurking somewhere with info produced by SMATRIX1?)
- in parallel, start comparing the event files when both cudacpp and fortran MEs are computed: I expect that the color and helicity info should be different, or missing?
- in parallel also, try to increase the precision with which the cross section is disaplayed: in principle, we do get E-3 differences, so it is kind of surprising that I get exactly the same results (#476)... note in any case that I tested using OUT=2*OUT2, and I do get a cross section multiplied by 2 and various differences in unweighted events, so actually the status so far is pretty good!

### DIFF
--- a/epochX/cudacpp/CODEGEN/MG5aMC_patches/patch.P1
+++ b/epochX/cudacpp/CODEGEN/MG5aMC_patches/patch.P1
@@ -206,10 +206,10 @@ index 104747ce9..0f522165f 100644
              ALL_P1(2,I,IVEC)=-ALL_P1(2,I,IVEC)
              ALL_P1(3,I,IVEC)=-ALL_P1(3,I,IVEC)
 diff --git b/epochX/cudacpp/gg_tt.mad/SubProcesses/P1_gg_ttx/driver.f a/epochX/cudacpp/gg_tt.mad/SubProcesses/P1_gg_ttx/driver.f
-index 91e1f5b4e..455d6e33a 100644
+index 91e1f5b4e..1e70e9881 100644
 --- b/epochX/cudacpp/gg_tt.mad/SubProcesses/P1_gg_ttx/driver.f
 +++ a/epochX/cudacpp/gg_tt.mad/SubProcesses/P1_gg_ttx/driver.f
-@@ -71,13 +71,41 @@ c      double precision xsec,xerr
+@@ -71,13 +71,42 @@ c      double precision xsec,xerr
  c      integer ncols,ncolflow(maxamps),ncolalt(maxamps),ic
  c      common/to_colstats/ncols,ncolflow,ncolalt,ic
  
@@ -234,6 +234,7 @@ index 91e1f5b4e..455d6e33a 100644
 +#else
 +      NB_PAGE_LOOP = 32
 +#endif
++      write(*,'(a16,i6)') ' NB_PAGE_LOOP = ', NB_PAGE_LOOP
 +      if( nb_page_loop.gt.nb_page_max .or. nb_page_loop.le.0 ) then
 +        write(*,*) 'ERROR! Invalid nb_page_loop = ', nb_page_loop
 +        STOP
@@ -251,7 +252,7 @@ index 91e1f5b4e..455d6e33a 100644
  c
  c     Read process number
  c
-@@ -132,7 +160,8 @@ c   If CKKW-type matching, read IS Sudakov grid
+@@ -132,7 +161,8 @@ c   If CKKW-type matching, read IS Sudakov grid
            exit
   30       issgridfile='../'//issgridfile
            if(i.eq.5)then
@@ -261,7 +262,7 @@ index 91e1f5b4e..455d6e33a 100644
              stop
            endif
          enddo
-@@ -199,8 +228,33 @@ c      call sample_result(xsec,xerr)
+@@ -199,8 +229,33 @@ c      call sample_result(xsec,xerr)
  c      write(*,*) 'Final xsec: ',xsec
  
        rewind(lun)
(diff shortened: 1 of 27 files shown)


valassi:
I have increased the precision of the cross section printout. Now one can see E-6 differences in ggtt*, while eemumu is essentially the same results... see #476

valassi:
I am debugging a simple eemumu case with 32 events. Very peculiar, in dsample.f the elements are the same, but then the mean is different?

image

valassi:
Hi @oliviermattelaer ... who is kn in dsample.f?

Looks like this is filled when SMATRIX1 is called, but has an invalid value when SMATRIX1 is not called?

Thanks
Andrea

valassi:
Looks like indeed there was a common block hidden in the background...
```
      integer             dim, events, itm, kn, cur_it, invar, configs
      common /sample_common/
     .     tmean, trmean, tsigma, dim, events, itm, kn, cur_it, invar, configs
```

oliviermattelaer:
I will need to double check (tommorow) but this sounds like a event counting number.
(like the number of events that have returned non zero value or the number of event in this iteration so far).
I do not think that this variable is updated/used within the matrix element

valassi:
Thanks Olivier... maybe it is as simple as a missing call to 'RESET CUMULATIVE VARIABLE', this is the printout that is missing (and it is where kn=0 is reset).

valassi:
Hm yes in 5a363859aac4e905132898c06a957b12e23399f9 I added a printout
```
         IF(NTRY(IMIRROR).EQ.(MAXTRIES+1)) THEN
+          write(6,*) 'DEBUG_SMATRIX1 #2'
           CALL RESET_CUMULATIVE_VARIABLE()  ! avoid biais of the initialization
         ENDIF
```

and this was called once
```
    ./cmadevent_cudacpp < /tmp/avalassi/input_ggtt_cpp | grep DEBUG_ | sort | uniq -c
      16416  DEBUG_SMATRIX1 #1
     262656  DEBUG_SMATRIX1 #1a
          1  DEBUG_SMATRIX1 #2
      16416  DEBUG_SMATRIX1 #4
         25  DEBUG_SMATRIX1 #4a
          1  DEBUG_SMATRIX1 #4b
      16416  DEBUG_SMATRIX1 #7
      16416  DEBUG_SMATRIX1 #8
```

So I need to understand where to call that...

oliviermattelaer:
Ok makes sense. good catch.

The call to "RESET_CUMULATIVE_VARIABLE" it is more a security than anything else. 
The idea is that in weird cases, the choices of the helicity can have impact on the multi-channel factor and therefore I want to trash away the event where I did keep all the helicity to avoid any potential bias.

So it should be fine to ignore it especially if we move to 
 - helicity recycling  (planned for vectorised cpu)
 - preprocessing of helicity list (needed for gpu)

valassi:
Thanks Olivier! Listen, for the moment I have a workaround, namely I inject a single RESET_CUMULATIVE_VARIABLE call when I am in cudacpp-only mode, ie no fortran MEs are computed via SMATRIX1, and this seems to work. I now get the same cross sections and numbers of unweighted events whether I use fortran MEs or cudacpp MEs, so this seems to work.

And it is a good milestone :-)

The results are nice for eemumu. For ggtt* I still get the E-4 deviations, I will investigate that too.

valassi:
The commits so far are a self consistent set.

This mainly includes two important fixes in the fortran
- adding OUT=OUT2 to allow using cudacpp MEs in xsec calculations and event unweighting
- adding one RESET_CUMULATIVE_VARIABLE call to get the right cumulative cross sections from the same event-by-event weights

In addition:
- the fortran driver now allows setting FBRIDGE_MODE (compute only fortran or cudacpp MEs, or compute both and compare) at runtime through the input file
- the scripts include a few additional tweaks to compare fortran ad cudacpp xsections and numbers of events

The three main things still to do in 'basic' fortran vs cudacpp comparisons:
    - compare events.lhe fortran vs cudacpp (so far only checked numbers of events)... this comparison will give differences until we fix the random choice of colors and helicities (#402 #403 #404), but all the rest should already be the same
    - fix color errors in three ggttg* processes (https://github.com/oliviermattelaer/mg5amc_test/issues/14)
    - fix E-4 discrepancies fortran vs cudacpp (#476)

I will merge this as a standalone PR.

All tests are passing. I am self merging

PR #489, opened 2022-06-15
Next patch for cudacpp-only MEs: add dummy color choice to fix 'failed to reduce to color index' 

valassi:
This is a workaround for https://github.com/oliviermattelaer/mg5amc_test/issues/14

It is the first step of #402: it implements a dummy color choice (as opposed to a random color choice)

Essentially I have just added JAMP2_MULTI(0,IVEC) = NCOLOR. I have not even initialised JAMP2_MULTI(1:MAXFLOW,IVEC), it seems that addmothers.f manages this by itself somehow


valassi:
This is a followup to th eprevious PR #486.

Amongst the things that remain to be done before a first milestone (xsections ok, and events ~okish except for random color and helicity) there are these two:
- compare events.lhe fortran vs cudacpp (so far only checked numbers of events)... this comparison will give differences until we fix the random choice of colors and helicities (https://github.com/madgraph5/madgraph4gpu/issues/402 https://github.com/madgraph5/madgraph4gpu/issues/403 https://github.com/madgraph5/madgraph4gpu/issues/404), but all the rest should already be the same
- fix E-4 discrepancies fortran vs cudacpp (https://github.com/madgraph5/madgraph4gpu/issues/476)

All tests are passing, I am self merging

PR #490, opened 2022-06-16
Use non-zero t-channel widths in fortran as in cudacpp (recover the same physics precision in fortran and cudacpp to 10+ digits)

valassi:
This is a PR for issue #476, the ~permille differences between fortran and cudacpp MEs for ggtt* processes.

I intially thought this was a QCD related problem, such as alphas related. Instead it is due to the handling of the top quark width:
- fortran MEs use a top quark width hardcoded to ZERO
- cudacpp MEs use a non-zero top-quark width

The analysis is still WIP. But we also need to understand how to handle this...

@oliviermattelaer do you have a suggestion? Is there a switch in code generation to treat the top quark as width-0? Thanks

PR #493, opened 2022-06-19
COMPLETE FIRST MILESTONE: same xsec AND same events.lhe (except color and helicity) in fortran and in cuda or cpp

valassi:
This completes my first desired milestone for comparing fortran and cudacpp MEs: 
- as from the previous PR #490 we now get the same xsec to 10+ digits for all five processes
- in this new simple PR I just check that we also get the same events.lhe files for all five processes (except for random color #402 and random helicity #403)

In practice, I added a simple script that modified the fortran events.lhe to remove the color and helicity info, then I compare this to those I get from cuda and cpp

Next step will be the random color and helicity, see also #404.


valassi:
All tests pass, I self merge

PR #494, opened 2022-06-20
First summary tables for madevent+cudacpp

valassi:
This WIP PR has first summary tables for madevent+cudacpp


valassi:
Notice in the table above that the MEs throughputs insie madevent (eg 6.11e01 for cpp/none) are compatible with those obtained from standalone MEs (eg with bridge and without bridge both 7.16e+01), within 20%. This is true for ggttggg, but less so for less complex processes.

Note also that these numbers (computed in this set of tests) are compatible with those in the tput summary table, https://github.com/madgraph5/madgraph4gpu/blob/bfbfb5316a850c6f5f4a19b07779cd8842bf610f/epochX/cudacpp/tput/summaryTable_bridge.txt
image

valassi:
THis PR is now ready and all tests are passing. 

I plan to use these numbers for the ICHEP talk.

I am self merging.

Issue #505, opened 2022-08-19
support for clang >= 13 (and CXX with more than one word)


valassi:
I have tested routinely clang from cvmfs at cern until clang 12.

Starting with clang13, I have a problem because the cvmfs installations I am using are broken
https://sft.its.cern.ch/jira/browse/SPI-1933

The problem here is that the old clang installations (using gcc toolchains from gcc version sthat are not the default on the O/S) were working with COMPILER_PATH, while this now seems no longer supported (see issue 51351 in llvm/llvm-project).

In the cvmfs installations (eg clang 14), the issue now seems to be fixed using an explicit use of gcc toolchains in the CXX
```
echo $CXX
clang++ --gcc-toolchain=/cvmfs/sft.cern.ch/lcg/releases/gcc/12.1.0-57c96/x86_64-centos7
```

This may work fine for madgraph cpp builds. However, it does not work ok for cuda builds, because I am using -ccbin (is this really needed), and I cannot pass 'clang++ --gcc-toolchain...' to -ccbin. I will disable cuda builds when these versions of clang are used.

Eventually, I hope that COMPILER_PATH is supported again, otherwise I am not sure what we can do.

valassi:
Specifically I also need to fix errors like
```
which: unrecognized option '--gcc-toolchain=/cvmfs/sft.cern.ch/lcg/releases/gcc/12.1.0-57c96/x86_64-centos7'
```

Issue #504, opened 2022-08-19
add an easy way to add cuda -allow-unsupported-compiler


valassi:
On juwels while helping @Jooorgen 
```
ccache /p/software/juwels/stages/2022/software/CUDA/11.5/bin/nvcc  -O3  -lineinfo -I. -I../../src -I../../../../../tools -I/p/software/juwels/stages/2022/software/CUDA/11.5/include/ -DUSE_NVTX -gencode arch=compute_80,code=compute_80  -gencode arch=compute_80,code=sm_80 -use_fast_math -std=c++17  -ccbin /p/software/juwels/stages/2022/software/Clang/13.0.1-GCCcore-11.2.0/bin/clang++ -DMGONGPU_FPTYPE_DOUBLE -Xcompiler -fPIC -c gcheck_sa.cu -o gcheck_sa.o
In file included from <built-in>:1:
In file included from /p/software/juwels/stages/2022/software/CUDA/11.5/include/cuda_runtime.h:83:
/p/software/juwels/stages/2022/software/CUDA/11.5/include/crt/host_config.h:147:2: error: -- unsupported clang version! clang version must be less than 13 and greater than 3.2 . The nvcc flag '-allow-unsupported-compiler' can be used to override this version check; however, using an unsupported host compiler may cause compilation failure or incorrect run time execution. Use at your own risk.
```

This happens all the time, if we want to use the latest clang or gcc. This was clang13 with cuda 11.5.

For the specific case we can test clang11 from 'mdule spider clang' on juwels. BUt in general it's good to have this option too

Issue #503, opened 2022-08-19
add support for cpp-only builds in .mad directories


valassi:
On juwels while helping @Jooorgen
```
[valassi1@jwlogin01 bash] ~/madgraph4gpu/epochX/cudacpp/ee_mumu.mad2/SubProcesses/P1_ll_ll> make 
cudacpp.mk:103: CUDA_HOME is not set or is invalid. Export CUDA_HOME to compile with cuda
cudacpp.mk:187: Using AVX='avx2' because this is faster than avx512vl for clang
ccache /p/software/juwels/stages/2022/software/GCCcore/11.2.0/bin/gfortran -o ./gmadevent_cudacpp myamp.o genps.o unwgt.o setcuts.o get_color.o cuts.o cluster.o reweight.o initcluster.o addmothers.o setscales.o idenparts.o dummy_fct.o driver_cudacpp.o  auto_dsig1_cudacpp.o auto_dsig.o matrix1.o   -L../../lib/ -ldhelas -ldsample -lmodel -lgeneric -lpdf -lcernlib  -lbias  -lstdc++  -Wl,--no-relax   -fopenmp counters.o -L../../lib//. -lmg5amc_common -lmg5amc_ll_ll_cuda -Wl,-rpath,'$ORIGIN/../../lib/'
/p/software/juwels/stages/2022/software/binutils/2.37-GCCcore-11.2.0/bin/ld.gold: error: cannot find -lmg5amc_ll_ll_cuda
driver_cudacpp.o:driver.f:function MAIN__: error: undefined reference to 'fbridgecreate_'
driver_cudacpp.o:driver.f:function MAIN__: error: undefined reference to 'fbridgedelete_'
auto_dsig1_cudacpp.o:auto_dsig1.f:function smatrix1_multi_: error: undefined reference to 'fbridgesequence_'
auto_dsig1_cudacpp.o:auto_dsig1.f:function smatrix1_multi_: error: undefined reference to 'fbridgesequence_'
collect2: error: ld returned 1 exit status
make: *** [makefile:120: gmadevent_cudacpp] Error 1
```

To be checked. I thought this was running smoothly but maybe not

valassi:
Note, some things were wrong in this patch. I had to fix it again. The final patch is in PR #499.

I think the point is that the Fortran Makefile does not know the NVCC variable, an dit would be too complex and unnecesary to add it. So instead I used a different workaround
[...]

PR #506, opened 2022-08-19
A few fixes for Jorgen on juwels

valassi:
A few fixes for Jorgen on juwels
- #503 fix the behaviour of 'export CUDA_HOME=invalid' for cpp only builds on .mad directories
- #504 allow the use of unsupported compilers if ALLOW_UNSUPPORTED_COMPILER_IN_CUDA is set
- #505 disable CUDA builds (and add other fixes) for multi-word CXX env variables (like clang --gcc-toolchain)

### DIFF
--- a/epochX/cudacpp/CODEGEN/MG5aMC_patches/patch.common
+++ b/epochX/cudacpp/CODEGEN/MG5aMC_patches/patch.common
@@ -147,7 +147,7 @@ index 6b581bdfd..32c71b457 100644
              enddo
           endif
 diff --git b/epochX/cudacpp/gg_tt.mad/SubProcesses/makefile a/epochX/cudacpp/gg_tt.mad/SubProcesses/makefile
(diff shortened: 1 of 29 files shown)


valassi:
@Jooorgen I am about to self merge this.

This should allow you to
- If you want to test a version of clang but do not care about gpu builds, set 'export CUDA_HOME=invalid': this will no walso work in .mad directories, previously it was only in .sa directories
- If you have strange combinations of old cuda and new clang (eg cuda 11.5 and clang13) on juwels, do 'export ALLOW_UNSUPPORTED_COMPILER_IN_CUDA=1' before building

Conversely it will still not work for me on clang13/clang14 at cern from cvmfs, but at least now the error messages will be more reasonable (and actually it might work fine for clang14, but only in cpp vuilds, while cuda builds are automatically disabled)

PR #510, opened 2022-08-21
Additional patches for benchmarking

valassi:
Additional patches for benchmarking

So far: add distclean to madevent makefiles


valassi:
I have added a protection against parallel builds of googletest. 

I used flock https://stackoverflow.com/a/32666215

In the bmk project, one downloads the project and immediately builds it. This is why a make -j upfront sees issues with googletest (in my dev directories, ususally it is there and I never touch it)

valassi:
But I should have imagined that mac must be different and has no flock. I will fix it only for linux then

valassi:
Self merging - th eprevious tests succeeded, now I just changed throughputX.sh

PR #513, opened 2022-08-24
Compiler tests for Jorgen and minor fixes for bmk

valassi:
Compiler tests for Jorgen and minor fixes for bmk

This includes
- the first tests for cuda117 (with gcc11.2)
- a test of cuda115 and gcc112 for juwels

For the bmk I further simplified the googletest makefiles


valassi:
self merging

Issue #516, opened 2022-08-25
icpx: warning: overriding '-ffp-contract=fast' option with '-ffp-contract=on' [-Woverriding-t-option]


valassi:
This is for clang14 (also 13?), or for icpx that uses clang14.

I think that it is because I am partly disabling fast math on one file for c++, but cuda always keeps it
``` 
ccache /usr/local/cuda-11.7/bin/nvcc  -O3  -lineinfo -I. -I../../src -I../../../../../tools -I/usr/local/cuda-11.7/include/ -DUSE_NVTX -gencode arch=compute_70,code=compute_70 -gencode arch=compute_70,code=sm_70 -use_fast_math -std=c++17  -ccbin /cvmfs/projects.cern.ch/intelsw/oneAPI/linux/x86_64/2022/compiler/2022.1.0/linux/bin/icpx -allow-unsupported-compiler -DMGONGPU_FPTYPE_DOUBLE -Xcompiler -fno-fast-math -Xcompiler -fPIC -c gCrossSectionKernels.cu -o gCrossSectionKernels.o
clang++: warning: overriding '-ffp-contract=fast' option with '-ffp-contract=on' [-Woverriding-t-option]
clang++: warning: overriding '-ffp-contract=fast' option with '-ffp-contract=on' [-Woverriding-t-option]
clang++: warning: overriding '-ffp-contract=fast' option with '-ffp-contract=on' [-Woverriding-t-option]
```

This is  llvm/llvm-project issues/54625

valassi:
I added a patch to avoid conflicting fast math and no fast math, but this is not enough: even just specifying no fast math triggers the warning?!
```
icpx  -O3  -std=c++17 -I. -I../../src -I../../../../../tools -DUSE_NVTX -Wall -Wshadow -Wextra   -march=skylake-avx512 -mprefer-vector-width=256  -DMGONGPU_FPTYPE_DOUBLE -fno-fast-math -I/usr/local/cuda-11.7/include/ -fPIC -c CrossSectionKernels.cc -o CrossSectionKernels.o
clang++: warning: overriding '-ffp-contract=fast' option with '-ffp-contract=on' [-Woverriding-t-option]
```

I therefore added instead a `-Wno-overriding-t-option`... I am not sure it is a good idea (maybe I should leave the warning?) but I will go for this for the moment? Or I will keep it there disabled for now.

valassi:
[...]

Issue #515, opened 2022-08-25
Fix cuda11.7 with clang14 and gcc12.1


valassi:
I added wrappers for clang14. This allows to go beyond gcc toolchain issues (https://sft.its.cern.ch/jira/browse/SPI-1933), but I still get issues. This may be from cuda117 being not supported on gcc12?
```
ccache /usr/local/cuda-11.7/bin/nvcc  -O3  -lineinfo -I. -I../../src -I../../../../../tools -I/usr/local/cuda-11.7/include/ -DUSE_NVTX -gencode arch=compute_70,code=compute_70 -gencode arch=compute_70,code=sm_70 -use_fast_math -std=c++17  -ccbin /data/avalassi/GPU2020/madgraph4gpuX/tools/compilers/mg-lcg-clang++-14.0.6 -allow-unsupported-compiler -DMGONGPU_FPTYPE_DOUBLE -Xcompiler -fPIC -c gcheck_sa.cu -o gcheck_sa.o
/cvmfs/sft.cern.ch/lcg/releases/gcc/12.1.0-57c96/x86_64-centos7/lib/gcc/x86_64-pc-linux-gnu/12.1.0/../../../../include/c++/12.1.0/bits/stl_map.h:593:15: error: expression contains unexpanded parameter pack '_Args'
if constexpr (__usable_key< __decltype(__a), remove_reference_t< _Args> > ) 
              ^                                                  ~~~~~
1 error generated.

```

valassi:
Trying gcc12.1 (not clang14) I get other issues
```
ccache /usr/local/cuda-11.7/bin/nvcc  -O3  -lineinfo -I. -I../../src -I../../../../../tools -I/usr/local/cuda-11.7/include/ -DUSE_NVTX -gencode arch=compute_70,code=compute_70 -gencode arch=compute_70,code=sm_70 -use_fast_math -std=c++17  -ccbin /cvmfs/sft.cern.ch/lcg/releases/gcc/12.1.0-57c96/x86_64-centos7/bin/g++ -allow-unsupported-compiler -DMGONGPU_FPTYPE_DOUBLE -Xcompiler -fPIC -c gcheck_sa.cu -o gcheck_sa.o
[...]

PR #517, opened 2022-08-25
Additional patches testing new compiler combinations for Jorgen

valassi:
Additional patches testing new compiler combinations for Jorgen

@Jooorgen unfortunately this does not fix all the issues you saw, but it's the best I can do for now... we will need to invent some other combinations, first at CERN and then at HPCs.

This includes
- some clang wrappers solving the issues with gcc toolchain (thanks to @hassec !) ie https://sft.its.cern.ch/jira/browse/SPI-1933
- a few fixes for errors and warnings with newer versions of clang and gcc (an error from gfortran12 and a warning from clang++13 IIRC)

I am moving more and more to cuda 11.7. Note that gcc12.1 is still giving me issues (#515), I stopped at gcc11.2. 

Note also that I have some warnings from newer clang (and icx) because of using/disabling fast math (#516). I have added a patch for disabling the warning, but it is commented out... maybe it is better to keep the warning and think about what it really measn.


valassi:
Self merging

PR #518, opened 2022-08-28
Update scalability plot with latest ggttgg.sa code

valassi:
Update scalability plot with latest ggttgg.sa code.

These are the benchmarking-style plots for simd and multicore. They are produced with manually built code and ad hoc parsing for the moment. Eventually they can be produced by using prebuilt singularity containers


valassi:
The relevant plot is this one.
https://github.com/madgraph5/madgraph4gpu/blob/b6839202dbd5aa88fc2c0174618efc9e44b83936/tools/benchmarking/plots/ggttgg.2022/pmpe-simd.png

It can still be improved, eg plot throughput ratio to 1copy/SIMDnone to see actual numbers (should get up to ~64 at most, with 16 physial cores and a factor 4 from SIMD)
image

valassi:
Voila I also added a throughput ratio plot. You clearly see how we get almost the full theoretical 64x speedup for 16 physical cores with 4 doubles per AVX2 register.
https://github.com/madgraph5/madgraph4gpu/blob/6ff646bfc6fbae790c527ae81e404a7b2916d57e/tools/benchmarking/plots/ggttgg.2022/pmpe-simd-ratios.png
image

valassi:
This is related to #519, I have created an issue to discuss these plots also for sycl and kokkos.

Self merging

PR #521, opened 2022-08-29
Additional patches for benchmarking - new throughputX.sh options, new logs, new summary tables

valassi:
Additional patches for benchmarking - new throughputX.sh options, new logs, new summary tables

I went back to upstream/master logs for easier MRs, but they are available in the history


valassi:
Self merging

PR #522, opened 2022-08-30
fix bug in throughputX.sh (missing fptype in cpp tests!)

valassi:
Fix bug in throughputX.sh (missing fptype in cpp tests!)

The bug was introduced in 13a8faca40f3cb3afbd4c6b276416ac6c8262f76


valassi:
Self merging - no change in code, only in a script

PR #523, opened 2022-08-31
Minor fixes in bmk scripts

valassi:
Minor fixes in bmk scripts

Avoid erros from simdSymSummary on a read only filesystem


valassi:
Self merging. Changes are only in scripts

PR #525, opened 2022-08-31
add BMKEXEARGS to allow BMK containers to change blocks and threads

valassi:
add BMKEXEARGS to allow BMK containers to change blocks and threads


valassi:
just changed a script - self merging

PR #526, opened 2022-09-01
Additional features for benchmarking

valassi:
Additional features for benchmarking

Add BMKEXEARGS

Use six degimal digits and dump to file the pmpe04 tests


valassi:
self merging

Issue #519, opened 2022-08-28
scalability plots with different SIMD and different numbers of cores used


valassi:
Hi all, it would be nice to have some scalability plots with different SIMD and different numbers of cores used

I produced some for ggttgg in cudacpp in MR #518 

![image](https://user-images.githubusercontent.com/3473550/187084056-ae50f51e-27e3-4cad-b4a9-622e447c5122.png)

@jtchilders @nscottnichols @whhopkins do you think it would be possible to create something similar for kokkos and sycl? Here for cudacpp I am doing several copies in parallel as I have no MT (I disabled OMP). For kokkos and sycl the question that we discussed in some meetings is whether/how you can switch on/off simd levels (and if they do anything at all) and also how to modify the numbers of CPU cores that are used.

Thanks!
Andrea

valassi:
The plots above were obtained manually, by building the cudacpp code and then executing it with ad hoc scripts.

In MR #527 I have now added a set of scripts and plots based on containers from the BMK project. This makes it MUCH easier to extract data. 

Note: the fact that avx2 above seens to exploit HT and even overcommit is probably an artefact. The tests are too short, then they do not overlap with one another long enough. I definitelty saw this in other places (eg ggttg are configured there to run too short, and the effect is much more visible). I will evevntually rerun with much longer durations...

valassi:
For cudacpp, this is what I have done here
https://indico.cern.ch/event/1170924/contributions/4954511/
![image](https://user-images.githubusercontent.com/3473550/195983921-cf1d109a-d33a-40a7-a838-67c09f2dfd58.png)

PR #527, opened 2022-09-03
Scripts and plots from BMK containers

valassi:
This is a followup on #519.

I finally managed to create a decent container in the BMK project, and using it makes it MUCH easier to create benchmarking plots for whole node scalability.

This is a first bunch of scripts, and plots on two nodes, a Haswll and a Silver.

There is a problem here because my tests were too short, so the jitter between different copies/processes was too much (they are supposed to overlap, but they ended up not overlapping). So especially for fast tests like 512y, or for processes configured with too few events like ggttg, it seems that overcommit has an effect. This is an artefact, https://its.cern.ch/jira/browse/BMK-1056


valassi:
This contains only scripts and plots. Self merging, no need for the CI

PR #528, opened 2022-09-15
Large (final?) set of json files, tables and plots using BMK containers

valassi:
This is one large (final?) set of json files, tables and plots using BMK containers


valassi:
Self merging

valassi:
Note a few additional comments on the plots here
https://its.cern.ch/jira/browse/BMK-995

roiser:
i get access denied ..

valassi:
Thanks Stefan, repeating here.

# July 2022

For the Madgraph vectorized C++ container, it would be useful to define at least two very well defined benchmarks: no-vectorization and best-vectorization.

In terms of absolute throughputs, typically "best" will give a factor 4 higher throughput than "none" on one of our standard Intel machines (using double precision arithmetics).

The important point however is that HEP-SCORE contributions are all normalized to 1 on a reference machine. In practice, I expect that on our typical reference machine the "1" of "best" actually has 4x the absolute throughput as the "1" of "none".

There are however some cases where we expect absolute throughputs where "best"/"none" is not 4:

-    On Intel Gold (thanks to two FMA units?) we get a factor 8 from AVX512, see https://github.com/madgraph5/madgraph4gpu/issues/173#issuecomment-1044368109
-    On Power9 there is only the equivalent of SSE with 128bit, so we get only a factor two, https://github.com/madgraph5/madgraph4gpu/issues/223#issuecomment-873041568

It would be interesting to plot:

-   the absolute throughput for "best" vs "none": you will see a linear behaviour, but some points off by a factor two too high (Gold) or too low (Power9)
-   similarly, the normalised benchmark for "best" vs "none" (both being 1 at the reference node), with the same jumos by a factor 2 high or low
-   the general plot of hep workloads (y axis) vs hepspec06 (x axis), with one set of points per different benchmark: you will clearly see that "best" does similar jumps by a factor 2 off too high or too low

# September 2022

This is essentially done here
https://github.com/madgraph5/madgraph4gpu/blob/4cceab2789075ce4b05a828ace67f1918e43a026/tools/benchmarking/BMK-COMPARE/all-sa-cpp-d-inl0.png

image

There are four plots, one per process.

Each plot has several curves, one per SIMD build. There are two or three points for each curve on the plot, for the three nodes itscrd70, pmpe04, bmk6130. The 512y and 512z curves only have two points, because pmpe04 stops at avx2 and does not have avx512.

There is also one curve for the "best" SIMD (which is avx2 for pmpe04, 512y for itscrd70 and 512z for bmk6130).

I do not have hepspec06 for these machines, so the x axis is always the ggttgg-d-inl0-none score (which I consider a reasonable 'traditional' measure of the power of the machine, without vectorization).

I have chosen to plot the total score for all cores, rather than the score per core, for th eonly and simple reason that this allows the plots to be separated (otherwise they would all be on top of one another, as these are Intels with similar frequency, and the differences are small).

They are normalized to itscrd70 as reference machine, which means that ALL points of itscrd70 are 1 by default. Note the x axis has a point at ~3.8 and one at ~8: this is because the numbers of cores are 4, 16, 32, so in a ratio of 1, 4, 8, but the central machine pmpe04 is a bit older, so has a smaller score per core.

The two notable aspects here are especially the 'best' score in the brown lines

    1. The pmpe04 'best' score on the y axis is a bit (~10%) lower than the diagonal
    2. The bmk6130 'best' score can be much higher (up to ~40% for ggttgg) than the diagonal
    This is because the diagonal is an extrapolation of itscrd70 'best', which is 512y: pmpe04 stops at avx2, which is ~10% less than 512y, while bmk6130 has two FMA units and profits from 512z (it is not a factor 2, maybe because of the ~30% clock slowdown in an overloaded full AVX512 node?). Also interesting that different processes have different patterns, i.e. essentially only ggttgg shows a benefit of 512z, while for the others it is close to 512y (not clear why).

Final note: the violet curves are somewhat an artefact. Remember that these compare 512z on bmk6130 (where it is ~40% higher than 512y) to 512z on itscrd70 (where it is 20% LOWER than 512y!). This gives an overall 60% increase for ggttgg. In other words, if one was forced to use 512z all the time, then bmk6130 would be 60% more powerful than itscrd70. BUT on itscrd70 one can use 512y and gain back 20%, so for the best SIMD bmk6130 is only 40% better than itscrd70.

Closing this as done

PR #530, opened 2022-09-19
Explicitly made fpvectors initialised with 4 zeros

zeniheisser:


valassi:
Thanks Zenny - this eventually got slightly improved with "{}" rather than "{0,0,0,0}" which only works for SIMD with 4-element vectors
https://github.com/madgraph5/madgraph4gpu/pull/532/commits/b9793b2dec4e66f7e79bef8c09d65e4678b6f7ea
The new patch will go to master via MR #532 (just cleaning up some documentation now...)

Issue #537, opened 2022-09-29
Single precision color algebra


valassi:
As discussed at the hackathon from Olivier's and Zenny's ideas/work - we can move color algebra to single precision because here we get always E-5 precision or better.

In the code: agreed to keep a switch anyway to support both.
I propose `#ifdef MGONGPU_COLORALGEBRA_FLOAT`.

We need a float_sv type, where float_v has neppV twice the neppV for standard fptypes.
I will take care of that.

Zenny will work on the implementation. We need something like
```
#ifdef __CUDACC__
      typedef float float_sv;
#else
      typedef fptype_sv float_sv; // AV FIXME!
#endif      
#ifdef MGONGPU_COLORALGEBRA_FLOAT
      typedef float_sv fptype2_sv;
#else
      typedef fptype_sv fptype2_sv;
#endif
      fptype2_sv fjampR_sv = (fptype2_sv)( cxreal( jamp_sv ) );
      fptype2_sv fjampI_sv = (fptype2_sv)( cximag( jamp_sv ) );
      ... color algebra...;
      deltaMEs += unchanged (sum a float to a double)
```

valassi:
This is complex for vector C++. Suppose you have avx2, so vectors of 4 doubles or 8 floats would be optimal (there is NO INTEREST in floats for color algebra unless you can use wider SIMD vectors). The problem is that the handling of FFVs gives you a "page" of 4 events. I have done some relatively simple gymanstics to save the previous 4 events, and do the color algebra and the ME update only once every two pages. This functionaly works... but it is SLOWER than using double all the way! The problem is that at some point I must merge two 4-vectors into one 8-vector and similarly at some point split one 8-vector into two 4-vectors... I am using explicit loops, as I thought that in any case these operations are a small overhead, but somehow the whole thing is slower.

One avenue to explore is to do these 4/8 conversions using `__builtin_shufflevector` (https://gcc.gnu.org/onlinedocs/gcc/Vector-Extensions.html), but in my gcc11.2 this does not work. Apparently it is supported in gcc12, because it comes from clang, https://gcc.gnu.org/bugzilla/show_bug.cgi?id=88601.

Not sure what i will do. Most likely I will just implement it for CUDA as Zenny had initially done, and reevaluate C++ much later...

valassi:
Hm `__builtin_shuffle` should work... but it will take some time

oliviermattelaer:
Hi Andrea,

[...]

Issue #475, opened 2022-06-05
Check again if it is possible to simplify color algebra (A-iB)(M)(A+iB) as AMA+BMB


valassi:
Just a brief reminder after the discussion at the last meeting.

It was not 100% clear if this change (merged in PR #457) made sense, for instance
- is the matrix M symmetric?

Note in any case that
- This is a complex operation but the code was already assuming the result is real (quite simply the imag part was ignored?)

valassi:
Renamed as "Simplify color algebra (A-iB)(M)(A+iB) as AMA+BMB and in addition use the fact M is symmetric "

From Oliver's idea: we can gain a factor two by using the fact that a_i*M_ij*a_j + a_j*M_ij*a_i = 2*a_i*a_j*M_ij!

Conversely, my initial assumption that computing only AMA+BMB instead of AMA+BMB +iAMB - iBMA gains a factor two is wrong! We were in any case calculating ztemps as MA and MB, so the part I omitted was just two vector*vector multiplications, which computationally are negligible (thanks Olivier!).

There is indeed the possibility to gain a factor two, but this comes from the extra rewriting suggested above for 2*a_i*a_j*M_ij

valassi:
The speedup due to a symmetric M will be in PR #545. Note that initially this was in a commit with also single precision color algebra, I have moved that to #537 which will be in a separate PR.

valassi:
Uff... also this one is much more complex than it seemed.

Taken alone (without the mov eto float), this rewriting of the algorithm does go faster on C++ (a bit, maybe 10%), but it is SLOWER on cuda. And all of this changes depending on the process, and on the grid size. For ggttgg with a full grid, or ggttggg, CUDA is slower. So this must be reverted...

One thing that I plan - but it will take time, probably after ACAT - is to use some constexpr. We should really COUNT the numbers of operations. Maybe it is the "2 *" that is actually one more cycle. Using that once and for all in a constexpr (from cf define another constexpr cf2) should speed this up, and maybe would help even in C++. I do not thinks the compiler can optimize this away. Or maybe it is the defining of new variables (but I tried to work around it, no luck).

So I would say: for the hack2 PR, I will add this to C++, not to cuda. Then later on we reconsider.
And then a hack3 PR where we move also to float and mixed, but here it is C++ that gets more complicated.
No rocke science anywhere, but a lot of places where we should just be slow and careful, nit something to do in a rush

valassi:
[...]

PR #545, opened 2022-10-15
Hackathon developments (2) Simplify color algebra with symmetric color matrix

valassi:
This is essentially just using the fact that the color matrix is symmetric in #475 

This was in a hackathon branch by @zeniheisser and @oliviermattelaer , I have split it in two parts. The second part (using single precision color algebra #537) will be in a later MR


valassi:
I have unlinked #475 because this might not be the last time this is touched. In the end I used only on c++ the fact that the matrix is symmetric (and also a constexpr), while on CUDA this seems slower. In any case this might go to cublas and tensor cores on cuda... For CUDA for the moment however I did split real/imag, this is harmles (and needed for single precision color matrix).

I linked the fix in the throughput script instead.

I am rerunning all tmad tests. I will recreate summary tables and merge... need to wait 4-5 hours at least (see #547).

valassi:
I am reviewing thi sbefore merging. In my hack2 branch
```
git log upstream/master..hack2 
```
gives me only changes relevant to #475.

Indeed what I changed is just
- use the fact that the matrix is real, and split the real and imaginary computations into two different variable sets entirely (both for c++ and cuda)
- use the fact that the matrix is symmetric, and use an upper triangular constexpr matrix to do the color algebra: however this is only for c++, as for cuda this seems to go slower

I will therefore keep #475 open, as a possible investigation on cuda.

Then I will self merge.

valassi:
I added comments in #475, and will keep that open.

Self merging.

Issue #546, opened 2022-10-16
Reassess mad/MEs split in tmad timing measurements (speed up the Bridge?)


valassi:
Reassess mad/MEs split in tmad timing measurements

There is clearly something fishy, that is obvious with ggttggg (but has some small signs in the other processes too): the "mad" part decreases from SIMD/none to higher SIMD exactly like the MEs do. I would tend to exclude that the Fortran is vectorized (also because the compiler flags that determine SIMD/none vs SIMD/avx2 are only in cudacpp I think).

Note also that this is VERY HIGH in c++/none and much lower in Fortran-only...

Most likely, there is some other part (in the bridge??) that is now wrongly attributed to Fortran "mad" and is instead C++ or CUDA. 

It would be nice not only to assign it to the right component in timing, but also to speed this up...

See this for ggttggg
https://github.com/madgraph5/madgraph4gpu/blob/d72071a332b98f06dfd7cc6f625748143e8d4c50/epochX/cudacpp/tmad/summaryTable_ggttggg.txt
```
===========================================================================================================
|            | mad                        | mad               | mad               | sa/brdg   | sa/full   |
-----------------------------------------------------------------------------------------------------------
| ggttggg    | [sec] tot = mad + MEs      | [TOT/sec]         | [MEs/sec]         | [MEs/sec] | [MEs/sec] |
===========================================================================================================
| nevt/grid  |                       8192 |              8192 |              8192 |      8192 |      8192 |
| nevt total |                      90112 |             90112 |             90112 |  256*32*1 |  256*32*1 |
-----------------------------------------------------------------------------------------------------------
| FORTRAN    | 1226.61 =   5.01 + 1221.60 |  7.35e+01 (= 1.0) |  7.38e+01 (= 1.0) |       --- |       --- |
| CPP/none   | 1576.95 = 115.27 + 1461.68 |  5.71e+01 (x 0.8) |  6.16e+01 (x 0.8) |  7.46e+01 |  7.45e+01 |
| CPP/sse4   |  841.07 =  63.89 +  777.19 |  1.07e+02 (x 1.5) |  1.16e+02 (x 1.6) |  1.40e+02 |  1.40e+02 |
| CPP/avx2   |  412.51 =  33.40 +  379.11 |  2.18e+02 (x 3.0) |  2.38e+02 (x 3.2) |  2.89e+02 |  2.89e+02 |
[...]

Issue #540, opened 2022-10-12
Performance regressions between February and July 2022? (or difference jwlogin07 vs jwlogin04?)


valassi:
I am checking the ICHEP paper numbers, I realise that I showed two sets of numbers on the slides, and they have a 20% throughput difference.

See MR #499 for Juwels and look especially at commit https://github.com/madgraph5/madgraph4gpu/commit/e1a3c73b4da38708d061f2209748e20304a4cbba

Note, I have another issue open #442 about the alphas regressions, but this is probably something else.

I will give in the ichep paper only the latest numbers. These are slower (and only exist for double), but at least they are consistent with madevent.

valassi:
I think this is not a performance regression in the code, but just a very large 20% oscillation, either due to load, or to differences (that I cannot understand) between login nodes.

It seems that my faster results two days ago were on jwlogin07, maybe the earlier ones were elsewhere? Anyway today I repeated THE SAME CODE tests on a different jwlogin04 and this is 10% to 20% slower. This is NOT a regression, it is the same code.

https://github.com/madgraph5/madgraph4gpu/pull/499/commits/77863e757e858ef7b36713b2a5e4ada351ace653
```
< +++  REVISION eb30e41 (commit date: 2022-10-12 16:42:20 +0200) +++
< On jwlogin07.juwels [CPU: Intel(R) Xeon(R) Gold 6148 CPU] [GPU: none]:
---
> +++  REVISION 59b311b (commit date: 2022-10-14 18:50:19 +0200) +++
> On jwlogin04.juwels [CPU: Intel(R) Xeon(R) Gold 6148 CPU] [GPU: none]:
15,20c15,20
< | FORTRAN    |    4.08 =   0.43 +    3.65 |   38.33 =   2.54 +   35.80 |  2.14e+03 (= 1.0) |  2.29e+03 (= 1.0) |       --- |       --- |
< | CPP/none   |    4.12 =   0.45 +    3.67 |   39.12 =   2.49 +   36.63 |  2.10e+03 (x 1.0) |  2.24e+03 (x 1.0) |  2.27e+03 |  2.31e+03 |
< | CPP/sse4   |    2.27 =   0.42 +    1.85 |   21.08 =   2.48 +   18.60 |  3.89e+03 (x 1.8) |  4.41e+03 (x 1.9) |  4.45e+03 |  4.57e+03 |
[...]

PR #499, opened 2022-06-24
ICHEP2022 results (including madevent+cudacpp tests on juwels in June 2022)

valassi:
This is a WIP PR for madevent+cudacpp tests on juwels

A few issues to solve eg #498 and #497 but also some nice results already

Issue #470, opened 2022-05-24
Print out how many helicities are used


valassi:
Minor reminder as a followup to #419 : print out how many helicities are used out of comb (in fortran and cudacpp code, and then keep that through tput/tmad scripts)

valassi:
This is going to be merged in MR #548 
See https://github.com/madgraph5/madgraph4gpu/pull/548/commits/0481c260e21e42547a8a9ce62770913e4824fb81

valassi:
This should be extended in #563

PR #548, opened 2022-10-16
Hackathon developments (3) Mixed precision (single precision color algebra, double precision FFV) + ACAT2022 results

valassi:
I now have something that starts to look mergeable

THis is about Mixed precision (single precision color algebra, double precision FFV ie #537

I had many issues on c++ (it was slow), but now it is working ok, not slower not faster, but ok (and its worth trying out eg on ggttggg)

On cuda it was obviously faster immediately.

Never a factor 2 anyway... but decent.

(NB THis includes hack2, which I will merge independently before)


valassi:
I am reviewing this MR which essentially can be merged
```
git log upstream/master..hack3 
```

The changes included are seentially the following
- this MR is MAINLY about implementing mixed precision #537 (single precision color matrix), completing earlier work by @zeniheisser , see the full description of changes in #537 
- there are many more summary tables, because this MR was used to prepare for ACAT and contains all relevant results shown there
- in particular, it is clear that the 'fortran overhead' may include some non-fortran part (because it goes faster when in avx2 mode, #546), and also that it would be useful to reduce the number of events per iteration (#555) 
- note that juwels results are computed on a very specific login node to avoid regressions, #540 (but this was already in MR #499)
- I have added here (because I wanted to make a slide, which then I did not do in ACAT but only in QCD@LHC) the printout of the number of good helicities #470 

I think this should be all for this MR. There is probably some extra stuff that can/should be cleaned up, but I think that it can be merged as is.

Will self merge.

Issue #544, opened 2022-10-15
(post-hack cleanup) rename 3.1.1_lo_vectorization as PROD


valassi:
One of many upcoiming items of post hackathon cleanup

We now moved to a different upstream branch thanks to Olivier and Zenny. The name 3.1.1_lo_vectorization makes no sense anymore in the CODEGEN stuff. Actually this should have been changed when moving from bazaar to git, with git we can use a commit hash and it does not matter which branch this is in. Probably still makes sense to keep this as a directory (in the unlikely case we want to exercise two upstream branches at the same time, as we did with 270gpu and 311veclo), so rename this as DEFAULT for instance or PRODUCTION.

valassi:
For the record, the move to the new commit will be in MR #532

valassi:
I have made some minor changes that I will merge in #556 
- the directory is called PROD
- like commit.GIT there is now a branch.GIT that contains the branch name

Note that the branch name is not really needed... one could checkout a commit, or reset any branch to a particular commit. But I keep it like this for the moment

PR #556, opened 2022-12-06
Rename 3.1.1_lo_vectorization as PROD in CODEGEN infrastructure

valassi:
Minor changes implementing #544


valassi:
All ok self merging

PR #557, opened 2022-12-06
Prepare CODEGEN to allow both PROD and DEV

valassi:
This is a followup to #544 

Prepare CODEGEN to allow both PROD and DEV


valassi:
Self merged without checks - no changes in the code

PR #559, opened 2022-12-07
First set of lhe patches (towards random color/helicity): upgrade upstream MG5aMC to vecMLM and port my vector.inc patches upstream

valassi:
This is a WIP MR to start the work on integrating the choice of random color #402 and helicity #403 in the LHE file

The very first step is upgrading the upstream MG5aMC that I use for code generation from nuvecMLM to vecMLM. This is because vecMLM, unlike nuvecMLM, includes a change by @oliviermattelaer in the fortran, to prepare it for this purpose.
See https://github.com/oliviermattelaer/mg5amc_test/issues/24

The current status is that I have a working version with vecMLM, but no random color/helicity yet. 

I had to modify the madX.sh script that compares LHE files in fortran and cudacpp. It seems that previosuly I was writing "0" as helicity in cudacpp, while now I am writing "2" (probably some default value?).

I start thinking that I will make this in several MRs, this first one being only the upgrade to vecMLM. I can take this also as an opportunity to clean up a bit the code generation. Still to do in particular
- try to remove all template files from CODEGEN, and move small differences to patch.* files (I did this already for unwgt.f and addmothers.f)
- check if I can further minimise differences to upstream
- if possible, package a MR for the upstream mg5amcnlo rather than keep my patches...
- regenerate all processes
- rerun all tests and check performance

About the last point, in particular, I am a bit puzzled as the fortran version now seems to go a factor 3-4 faster in the ME calculation?... Maybe I need to modify the way the counters are used, as some functions for choosing helicity/color have moved elsewhere (and should still be included...)


valassi:
All tests succeeded - I self merge

Issue #534, opened 2022-09-28
Add pp_tttt as an example of a process with many P1 subdirectories which need different nwf values


valassi:
example of process with two matrix-element in the same directory:
define p = j / g
define j = p
generate p p > w+ w+  j j

valassi:
To make it clearer

- pp to something creates two subdirectories: qq to something and gg to something

- qq to W something has nprocesses>1 within one qq subdirectory (because not all qq pairs are equivalent in MEs, but they are still integrated at the same time)

valassi:
A first example of a process with two P1 subdirectories is pp to 4 tops. There are issues there, MR #560 is in WIP

valassi:
As correctly pointed out by @zeniheisser the issue here is that some parts of mgOnGpuConfig are P1-specific, so there should really be a separate mgOnGpuConfig in each P1 subdirectory. Or, some parts of these should be moved back to the CPPProcess class where @oliviermattelaer had essentially added them initially. This is essentially what the cleanup issue #266 should all be about... (also vaguely related to #356)

PR #560, opened 2022-12-08
add pp_tttt to codegen and fix its builds (two P1 subdirectories which need different nwf values)

valassi:
Hi @zeniheisser this is a WIP MR with the changes we discussed. I am including the commit log below.
Maybe you can have a look at the build warnings in gg to tttt?
Thanks!
Andrea

--

Add pp_tttt to CODEGEN/generateAndCompare.sh and CODEGEN/checkFormatting.sh

Code is generated ok for both sa and mad. They include two P1 subdirectories each.

In sa, P1_Sigma_sm_gg_ttxttx builds ok but issues many warnings like 
```
CPPProcess.cc: In function ‘void 
mg5amcCpu::calculate_wavefunctions(int, const fptype*, const fptype*, mgOnGpu::fptype*, int)’: CPPProcess.cc:334:34: warning: array subscript 13 is above array bounds of ‘mgOnGpu::fptype* [13]’ {aka double* [13]’} [-Warray-bounds]
  334 |       FFV1_1<W_ACCESS, CD_ACCESS>( w_fp[2], w_fp[6], COUPs[1], cIPD[0], cIPD[1], w_fp[13] );
      |       ~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
Then check.exe segfaults at runtime (not surprisingly)
This should be fixed in the cudacpp plugin.

In sa, P1_Sigma_sm_uux_ttxttx builds ok and check.exe succeeds at runtime.

In mad, P1_gg_ttxttx builds ok with the same warnings above, and (not surprisingly) segfaults at runtime.

In mad, P1_Sigma_sm_uux_ttxttx does not build. There are some errors like 
```
ccache /cvmfs/sft.cern.ch/lcg/releases/gcc/11.2.0-ad950/x86_64-centos7/bin/gfortran -w -fPIC -O3 -ffast-math -fbounds-check -ffixed-line-length-132 -w -cpp  -c auto_dsig1.f -I../../Source/ -fopenmp -o auto_dsig1.o auto_dsig1.f:338:47:

  338 |       DATA D1,U1,S1,C1/NB_PAGE_MAX*1D0,NB_PAGE*1D0,NB_PAGE*1D0,NB_PAGE*1D0/
      |                                               1
Error: Symbol ‘nb_page’ must be a PARAMETER in DATA statement at (1)
auto_dsig1.f:339:51:

  339 |       DATA CX2,SX2,UX2,DX2/NB_PAGE_MAX*1D0,NB_PAGE*1D0,NB_PAGE*1D0,NB_PAGE
      |                                                   1
Error: Symbol ‘nb_page’ must be a PARAMETER in DATA statement at (1)
auto_dsig1.f:274:49:

  274 |       DOUBLE PRECISION D1(NB_PAGE_MAX),U1(NB_PAGE),S1(NB_PAGE),C1(NB_PAGE)
      |                                                 1
Error: Symbol ‘nb_page’ at (1) has no IMPLICIT type
```
This should be fixed in the fortran code generation and/or patching.


valassi:
PS The commands to create the .sa and .mad respectivel;y are
```
./CODEGEN/generateAndCompare.sh pp_tttt
./CODEGEN/generateAndCompare.sh pp_tttt --mad
```

zeniheisser:
Found bug source: parameter nwf (number of wave functions) is set in shared header file src/mgOnGpuConfig.h. Since this header is identical for all subprocesses, it leads to to the array w_fp having incorrect allocation if nwf should differ between subprocesses

Issue does not appear when generating "u u~ > t t~ t t~" and "g g > t t~ t t~" independently

@valassi expects issue to be in the cudacpp plugin

valassi:
Copying @roiser for info

valassi:
I am having a look at this. Using the latest mg5amcnlo upstream, I get no build errors from nb_page anymore (there is no nb_page anymre, only vecsize_used or vecsize_memmax, maybe this helped).

However I still get the build warnings, and the runtime segfaults

valassi:
I have rebased this on the latest upstream/master. I confirm that pptt cde builds ok, while pptttt does not. Both have several P1 subdirectories, but only pptttt shows the issue described by @zeniheisser , namely that nwf should be different in the various P1 an dmust be moved from common code to each P1

valassi:
Note the connected issues
- #534, adding an example with two P1 dirs using different nwf: I link it to this WIP, it can be closed when thi sis done
- #636, about testing BOTH P1s in tput/tmad: low priority, for the moment it is enough to add here in #560 the test of one single P1 (the one that fails the build...)

Also will add comments in #272 that I am about to close

valassi:
Marking this as related to #644 that described this issue more generally.

I just rebased to the latest master

valassi:
I think that this MR is now almost ready to merge, including a fix for #644: I moved nwf from mgOnGpuConfig.h to the calculate_wavefunction deeply hardcode in CppProcess.cc. I was not able to do it in CppProcess.h, because the correct result for nwf is only available after a call to `self.helas_call_writer.get_matrix_element_calls`. Anyway, it looks ok now to me.

I am running tests on all processes and then I will merge.

(NB I am not including pp_tttt to the repository - yet?)

valassi:
All tests have passed, I am self merging. This closes #644

Issue #561, opened 2022-12-10
Large (x4) speedup in Fortran MEs with Olivier's random helicity/color vecMLM code: disable OMP multithreading


valassi:
Hi @oliviermattelaer @roiser @hageboeck @zeniheisser I have an interesting one.

I am making progress in the integration of random color and helicity. For the moment I have essentially completed the rebasing of my madgraph4gpu cudacpp patches on top of Olivier's latest upstream code (ie moving from nuvecMLM to vecMLM branches). I have two ongoing MRs on upstream mg5amcnlo and two MRs on madgraph4gpu, I will give the details.

I am rerunning my usual set of tests. The interesting, puzzling finding is the following: the Fortran ME calculation is now a factor 4 faster than it used to be. I do not think that Fortran is now magically vectorizing SIMD (should be checked with objsim), I would rather imagine that the algorithm for Fortran ME calculation has changed. It may also be that I am doing the "bookeeping" wrong, now that random color/helicity have moved elsewhere, but again I do not think this is the issue, as the OVERALL time taken by Fortran seems a factor 4 faster.

Note that
- the cross section produced by Fortran with the same random seeds has now changed
- the cudacpp implementation also produces this new cross section, which agrees with the new Fortran

This is in itself good news as speedups are always good, but it kind of significantly reduces the interest of C++ vectorization...

I think that we need to understand this quite well before we give the code to the experiments? It would especially be important to do some physics validation of the old Fortran vs the new Fortran.

Another think that it would be useful to test is whether changing the vector size has any effect.

Thanks for any feedback! Andrea

valassi:
See for instance
```
[avalassi@itscrd70 gcc11.2/cvmfs] /data/avalassi/GPU2020/madgraph4gpuX/epochX/cudacpp> git diff f8117c408 f4329daeb tmad/logs_ggttggg_mad/log_ggttggg_mad_d_inl0_hrd0.txt 
...
85c85
< 8192 ! Number of events in a single Fortran iteration (nb_page_loop)
---
> 8192 ! Number of events in a single Fortran iteration (VECSIZE_USED)
94c94
<  [XSECTION] nb_page_loop = 8192
---
>  [XSECTION] VECSIZE_USED = 8192
98,102c98,102
<  [XSECTION] Cross section = 2.158e-07 [2.1583423593436162E-007] fbridge_mode=0
<  [UNWEIGHT] Wrote 97 events (found 1199 events)
[...]

PR #562, opened 2022-12-10
Second set of lhe patches (towards random color/helicity): streamline VECSIZE_USED in upstream MG5aMC and rerun performance tests

valassi:
This is a second set of patches towards random color/helicity and towards a backport of my patches upstream. It is a followup to #559, that I already merged.

It is based on the second set of upstream patches that I opened upstream, https://github.com/mg5amcnlo/mg5amcnlo/pull/24.

It is still in WIP because
- the upstream MR is still WIP, as it has some cosmetic things I want to fix
- I am running performance tests on madgraph4gpu with these latest changes

Concerning performance tests, n particular, I observed a bizarre feature: Fortran MEs are now a factor 4 faster, see #561. I am strongly suspecting that this is due to some regression in how I integrated cudacpp, which resurrected issue #419 about LIMHEL. In practice, I suspect that in Fortran I am running four times fewer helicities than in cudacpp, because there are different LIMHEL thresholds? To be checked.

Anyway, I keep this in WIP. I will rerun tput perf tests, and I will investigate the performace changes. Note that this performance regression (Fortran speedup) was already in #559 that I merged.

When these issues are fixed, I will work on a third "|lhe" patch to actually integrate random color/helicity.


valassi:
I am running again all tests, but this is essentially ready to be mrged.

I have also added the printout of helicities #563 from both fortran and cudacpp

valassi:
This is now ready to be merged. I will wait for the CI to succeed and then self merge.

valassi:
Al lchecks passed, self merging

Issue #569, opened 2022-12-14
Different helicity numbering in fortran and cudacpp?


valassi:
I am making progress in the random choice of helicity #403.

However the chosen helicities have different indices, Example in ggtt
```
   1  0.64067963E-01  0.64067963E-01          1.00000000000 12 10
   2  0.58379673E-01  0.58379673E-01          1.00000000000 12 10
   3  0.70810768E-01  0.70810768E-01          1.00000000000 15 15
   4  0.67192668E-01  0.67192668E-01          1.00000000000  2  0
   5  0.71590585E-01  0.71590585E-01          1.00000000000 15 15
   6  0.72862110E-01  0.72862110E-01          1.00000000000 15 15
   7  0.14271254E-01  0.14271254E-01          1.00000000000  2  0
   8  0.63986754E-01  0.63986754E-01          1.00000000000 15 15
   9  0.46316382E-01  0.46316382E-01          1.00000000000 12 10
  10  0.35372741E-01  0.35372741E-01          1.00000000000 15 15
  11  0.73958407E-01  0.73958407E-01          1.00000000000 15 15
  12  0.70691203E-01  0.70691203E-01          1.00000000000  3  3
  13  0.70805000E-01  0.70805000E-01          1.00000000000 15 15
  14  0.30801404E-01  0.30801404E-01          1.00000000000  5  5
  15  0.64111868E-01  0.64111868E-01          1.00000000000 15 15
  16  0.74312047E-01  0.74312047E-01          1.00000000000  2  0
  17  0.60961835E-01  0.60961835E-01          1.00000000000  2  0
  18  0.67698020E-01  0.67698020E-01          1.00000000000  2  0
  19  0.49748773E-01  0.49748773E-01          1.00000000000 15 15
  20  0.71951996E-01  0.71951996E-01          1.00000000000  5  5
  21  0.52116331E-01  0.52116331E-01          1.00000000000 12 10
  22  0.69245648E-01  0.69245648E-01          1.00000000000  2  0
  23  0.64808141E-01  0.64808141E-01          1.00000000000  2  0
  24  0.66861231E-01  0.66861231E-01          1.00000000000 14 12
  25  0.70041112E-01  0.70041112E-01          1.00000000000 15 15
  26  0.61135249E-01  0.61135249E-01          1.00000000000 15 15
  27  0.66574932E-01  0.66574932E-01          1.00000000000  2  0
[...]

Issue #415, opened 2022-04-02
Move the loop over npagV from calculate_wavefunction to sigmakin?


valassi:
While debugging issues in the madevent+cudacpp integration (silly issue: I need to add debug printouts for momenta and MEs), I realise that the C++ loops on events and helicities may currently be suboptimal.

One of the very first changes I did to implement C++ vectorization was to invert the order of the helicity and event loops, making the event loop the innermost loop. This is absolutely essential for the events on an event page (neppV). However, it is not necessary, and may be slower, for the loop over event pages (npagV).

In other words:
- currently: sigmakin loops over helicity and calls calculate_wavefunction (cw) for a give helicity; internally, cw loops over npagV, while there is no loop over neppV because this is done with vector instructions
- another possibility would be: sigmakin loops over npagV and helicities, and calls cw for a given ipagV and helicity; internally, cw must do the calculation for all neppV in the given ipagV, which is still done with vector instructions

The advantage of the new option may be that the momenta for a given event page ipagV are reused immediately for different helicities, improving dat alocality. The current option instead forces all momenta to be reloaded for every new helicity. I am not sure this makes a big difference, probably the difference would be more noticeable for less complex processes. But it may be useful to test.

Low priority...

valassi:
I guess I was talking about this bit of code
```
    // PART 1 - HELICITY LOOP: CALCULATE WAVEFUNCTIONS
    // (in both CUDA and C++, using precomputed good helicities)
    // FIXME: assume process.nprocesses == 1 for the moment (eventually: need a loop over processes here?)
    for( int ighel = 0; ighel < cNGoodHel; ighel++ )
    {
      const int ihel = cGoodHel[ighel];
#ifdef __CUDACC__
      calculate_wavefunctions( ihel, allmomenta, allcouplings, allMEs );
#else
      calculate_wavefunctions( ihel, allmomenta, allcouplings, allMEs, nevt );
#endif
      //if ( ighel == 0 ) break; // TEST sectors/requests (issue #16)
    }
```

I have looked at this again in the context of PR #434 for alphas, and streamlining a bit the memoryaccessfunctions. In that context, I was indeed thinking that it would be useful to remove the nevt parameter above, i.e. have
```
      calculate_wavefunctions( ihel, momenta, couplings, MEs );
```
for both cuda and cpp. The difference would be that this piece of code that is presently inside calcuate_wavefunction would need to pass to sigmakin:
```
#ifdef __CUDACC__
[...]

PR #570, opened 2022-12-14
Event by event random choice of one helicity

valassi:
Hi @oliviermattelaer @roiser I have finally ~completed the patch for the random choice of helicity #403 

This MR does NOT include the random choice of color, which I have not yet done (and may be a bit more complex).

The MR is essentially complete with code regeneration of all processes. I am now rerunning the full tests to check performance and see functionalities.


valassi:
A few comments
- One of the most tedious parts was to add the new function arguments to all function calls. This is done not only for helicity #403 but also for color #402. There are random numbers in for helicity and color, and seleted helicity and color out.
- Correspondingly I also had to add buffers for helicity and color.
- As discussed in #415, the implementation for C++ required that I invert the order of the loop over event pages (SIMD vectors) and over helicities.
- The fact that we have recently added support for 'mixed precision' mode was an additional complication in the C++ implementation. Some simplifications may be appropriate in the future.
- I have now reenabled in madX.sh the event by event comparison of selected helicity in the LHE file for fortran and cudacpp. The "dummyHelicity" script is no longer needed.
- One additional complication #569 was that out of the box the order of helicities was different on fortran and cudacpp, so the helicity index was initially wrong, this is now fixed

While writing this I see that some tests failed in the CI, so there is more work to be done...

valassi:
Peculiar, it is some tests that fail, I can reproduce this interactively
- eemumu fails
- ggtt (where I developed the code) succeeds
- ggttg also succeeds
- ggttgg fails

The failures are that the events with given momenta give MEs which are not those in the refernce files?....

valassi:
Ah voila, as I thought I had forgotten some hardcoded parameters in the codegen...
```
epochX/cudacpp/gg_ttgg.mad/SubProcesses/P1_gg_ttxgg/fcheck_sa.f /tmp/avalassi/jgigIx_fcheck_sa.f e43625dcf872b3faa292e8b5df57ecf103fdd8f0 100644 epochX/cudacpp/gg_ttgg.mad/SubProcesses/P1_gg_ttxgg/fcheck_sa.f c0bbf580efa6892ab11c9a340b82f15338a2ce7c 100644
7c7
<       PARAMETER(NEVTMAX=2048*256, NEXTERNAL=6, NP4=4)
---
>       PARAMETER(NEVTMAX=2048*256, NEXTERNAL=4, NP4=4)
```

valassi:
OK I fixed the bug and regenerated all processes. It looks better in ggttgg interactively

valassi:
This MR is now complete. The CI tests are succeeding, I will self merge

Issue #575, opened 2022-12-16
Retry OMP multithreading in cudacpp


valassi:
With the changes for the random choice of helicity (#403, MR #570 and especially #415), the OMP multithreading loop has moved inside cudacpp. It is now in a place where maybe it could work better out of the box.

Note in fact that also Fortran OMP is now quite good (see #561), so I would expect something similar in cudacpp.

While doing the code move I disabled (commented out) the OMP pragmas. They should be reenabled and tested..

https://github.com/madgraph5/madgraph4gpu/blob/3780502a369c9583aa86cb878d50a9d7b9d491aa/epochX/cudacpp/gg_tt.mad/SubProcesses/P1_gg_ttx/CPPProcess.cc#L878
```
#ifdef _OPENMP
    // (NB gcc9 or higher, or clang, is required)
    // - default(none): no variables are shared by default
    // - shared: as the name says
    // - private: give each thread its own copy, without initialising
    // - firstprivate: give each thread its own copy, and initialise with value from outside
#pragma omp parallel for default( none ) shared( allmomenta, allcouplings, allMEs, channelId, allNumerators, allDenominators )
#endif // _OPENMP
    */
    for( int ipagV2 = 0; ipagV2 < npagV2; ++ipagV2 )
    {

```

valassi:
The idea is essentially the following:
- a C++ "grid" with VECSIZE_USED events is sent in parallel
- this includes ideally NPROC (or NPROC times a factor) SIMD vectors, and each SIMD vector or group thereof goes to one thread
- each  vector within  a thread uses SIMD

For instance on a 4-core machine with AVX2
- in double mode,  launch grids of 16 events: 4 threads process each one SIMD vector of 4 events
- in float mode, launch grids of 32 events: 4 threads process each one SIMD vector of 8 events
- in mixed mode, launch grids of 32 events: 4 threads process each two SIMD vectors of 4 events (Feynman) or one SIMD vector of 8 events (color matrix)

valassi:
In particular this hould be tested against pmpe04 or another node with 30+ cores. See the previous suboptimal results in #196

valassi:
Rather than open a new issue, I add a few ideas here.

OMP is one solution for MT in cudacpp. But custom multithreading is another possibility. What I am thinking of is the following
- Custom multithreading may be more efficient and tunable, if we put constraints on the allowed numbers of threads given the cores available and the NUMA domains too.
[...]

Issue #568, opened 2022-12-14
Remove (or simplify) ifdefs for MULTICHANNEL


valassi:
I want to simplify a bit the code structure, in particular remove the ifdefs for multichannel.

Essentially, calculate_wavefunctions has two different signatures if multichannel is enabled or disabled. This makes CPPProcess.cc even heavier in places
```
#ifdef MGONGPU_SUPPORTS_MULTICHANNEL
        constexpr unsigned int channelId = 0; // disable single-diagram channel enhancement
        calculate_wavefunctions( ihel, allmomenta, allcouplings, allMEs, channelId, allNumerators, allDenominators, ievt00 );
#else
        calculate_wavefunctions( ihel, allmomenta, allcouplings, allMEs, ievt00 );
#endif
```

I essentially NEVER use or test the option with MGONGPU_SUPPORTS_MULTICHANNEL not defined. I even realised now (I had completely forgotten!) that in the .sa mode I build with it undefined (see #473), but I no longer build it ir test it since a while.

I find it much easier/better to do the following:
- keep always the ful API with channelid, numerators, denominators
- internally, if MGONGPU_SUPPORTS_MULTICHANNEL is not enabled, ignore those parameters (similar to what I do now for channelid=0)

One of the main differences is that numerators and denominator arrrays will not be allocated at all.

I will give it a try

valassi:
I am getting a slightly different idea about this.
1. What I really dislike is that sigmaKin et al have different interfaces depening on the ifdef. So I would like to keep always the API where multichannel is possible (whether if thst is used or not)
2. One of the main advantages of removing (or ifdeffing away) some parts of the code in non multichannel mode is speed. In CPPProcess.cc the various numerator/denominator checks against channelId should be skipped altogether for instance.
3. I started to think about this because in no multichannel mode I do not even write out coloramps - so it is silly to keep the ifdefs if not all the code is there.

What I think would be cleaner is the following
1. Keep always the same API whether MGONGPU_SUPPORTS_MULTICHANNEL is supported or not. However pass null pointers for numerators and denominators and pass 0 for channelId if the ifdef is false. Optionally add a sanity check in all functions that channelId is 0 (if it is not zero you should throw because you are unable to respect the role of the function).
[...]

PR #577, opened 2022-12-19
Reenable OPENMP multithreading in cudacpp

valassi:
This is about #575, it reenables OMP MT in cudacpp


valassi:
The CI was failing, I needed to fix the patch for the differente API without MULTICHANNEL #568 

Now done

valassi:
This failed on MAC with clang. I was unable to fix that, see #578. So I just disabled it.

En passant, I also fixed the OMP hacks I had added for madevent to e able to disable it in #561

valassi:
And I fixed it again because Apple clang does not start with clang...

valassi:
Ok the CI finally succeeded including clang/MAC (with omp disabled #578)

I wil self merge

PR #586, opened 2023-02-17
Move CI to Alma9.

hageboeck:
In order to move to Alma9, devtoolset-10 had to be disabled, since Alma9 is already at gcc-11.
Cuda was updated to v12.0.

Both concern the GPU part only.


valassi (approved):
All good! Thanks a lot!

valassi:
Thanks @hageboeck ! I am merging this.

Issue #588, opened 2023-02-22
Alma9 builds of AVX=none use sse4.2


valassi:
I recently moved from centos7 (with gcc11.2 from cvmfs) to Alma9 (with a native gcc11.3 compiler). I have run my usual tests and I realised that the AVX=none builds actually use SSE4.2 vectorization.

I just ran a fast check and it is clear that "x86-64_v2" ie SSE4.2 is enabled by default
```
[avalassi@itscrd90 gcc11/usr] ~> echo | g++ -v -E -
Using built-in specs.
COLLECT_GCC=g++
OFFLOAD_TARGET_NAMES=nvptx-none
OFFLOAD_TARGET_DEFAULT=1
Target: x86_64-redhat-linux
Configured with: ../configure --enable-bootstrap --enable-host-pie --enable-host-bind-now --enable-languages=c,c++,fortran,lto --prefix=/usr --mandir=/usr/share/man --infodir=/usr/share/info --with-bugurl=http://bugs.almalinux.org/ --enable-shared --enable-threads=posix --enable-checking=release --enable-multilib --with-system-zlib --enable-__cxa_atexit --disable-libunwind-exceptions --enable-gnu-unique-object --enable-linker-build-id --with-gcc-major-version-only --with-linker-hash-style=gnu --enable-plugin --enable-initfini-array --without-isl --enable-offload-targets=nvptx-none --without-cuda-driver --enable-gnu-indirect-function --enable-cet --with-tune=generic --with-arch_64=x86-64-v2 --with-arch_32=x86-64 --build=x86_64-redhat-linux --with-build-config=bootstrap-lto --enable-link-serialization=1
Thread model: posix
Supported LTO compression algorithms: zlib zstd
gcc version 11.3.1 20220421 (Red Hat 11.3.1-2) (GCC) 
COLLECT_GCC_OPTIONS='-v' '-E' '-shared-libgcc' '-mtune=generic' '-march=x86-64-v2'
 /usr/libexec/gcc/x86_64-redhat-linux/11/cc1 -E -quiet -v - -mtune=generic -march=x86-64-v2 -dumpbase -
ignoring nonexistent directory "/usr/lib/gcc/x86_64-redhat-linux/11/include-fixed"
ignoring nonexistent directory "/usr/lib/gcc/x86_64-redhat-linux/11/../../../../x86_64-redhat-linux/include"
#include "..." search starts here:
#include <...> search starts here:
[...]

Issue #584, opened 2023-02-04
Upgrade build and performance infrastructure to cuda12 and el8 - and add ARM


valassi:
In the context of the BMK project I would like to build the madgraph container on ARM. See https://its.cern.ch/jira/browse/BMK-1160. I realied that for ARM cuda is only available as cuda12 starting on el8 or el9.

It is probably a good occasion to
- upgrade my x86 build node from el7 to el8 or el9 (this will also solve various crashes and driver issues)
- upgrade my performance summaries from cuda11.7 to cuda 12
- set up an ARM node with an Nvidia GPU, for both madgraph4gpu and BMK, see https://its.cern.ch/jira/browse/BMK-1161
- all of the above will also simplify compatibilities with more recent gcc, clang, icpx etc, possibly nvhpc

Note: the main reason why I initially resisted going from el7 to el8 was a problem in timing due to heavy system calls. This was in 2020, I hope it must have been fixed by now! But this should be cross checked, see #116

valassi:
This will be closed in MR #589 that addresses all isues for cuda12 and x86 nodes.

I have moved the ARM node to a separate issue #590.

PR #589, opened 2023-02-22
Move to cuda12 and to Alma9/gcc11.3

valassi:
This MR implements most of the infrastructure changes described in #584. I have upgraded my infrastructure in various steps, collecting logs each time 
1. First on centos7 (itscrd70), with gcc11.2 from cvmfs, I moved from cuda11.7 to cuda12.0, and I ran tests and collected logs
2. Then, with gcc11.2 from cvmfs and cuda12, I moved from centos7 (itscrd70) to alma8 (itscrd80), and I ran tests and collected logs
3. Finally, with cuda12, I moved from alma8 (itscrd80) to alma9 (itscrd90), where I also use a more recent native compiler gcc11.3, and I ran tests and collected logs

During step3, I realised that gcc11.3 on alma9 breaks AVX=none builds (#588), because it assumes SSE4.2 (x86-64-v2) by default, so I fixed that too. 

I also fixed another small issue that is probably related to gcc11.3 (an explicit "-I." is needed for ompnumthreads.cc).

And I improved the debugging in CudaRuntime.h (because during the setup of the alma9 node on puppet I was getting a strange situation where nvidia-smi was ok, but the cuda applications were failing with error 999, so now CudaRuntime.h is printing out that error code).

All code is ready, I will merge this when all logs on alma9 are done.


valassi:
I have rerun all tests on itscrd90 and committed alma9 logs.

I will self merge.

valassi:
All tests succeed, self merge

PR #591, opened 2023-02-23
Add support and logs for clang14

valassi:
Add support and logs for clang14

There are many interesting results in performance and in precision in the logs.

The only real changes were
- added clang14 wrapper scripts for cvmfs
- increased tolerance for xsec from 2E-4 to 4E-4 (ggttggg/f gives 2.8E-4 difference)


valassi:
All ok, self merging

Issue #592, opened 2023-02-23
Fix cuda12.0 with clang16 (in icx2023)


valassi:
I have moved from cuda11.7 to cuda12.0 (and from centos7 to alma9, which comes with builtin gcc11.3).

I tried builds of
- cuda12.0 with gcc113 native, no issues
- cuda12.0 with clang14 from cvmfs, no issues
- cuda12.0 with icx2023, which includes clang16 and gcc12.1, this gives some warnings below
- I will later try also gcc12.1

This issue replaces #515 about cuda11.7 with clang14 and gcc12.1, which I will close (I moved to cuda12.0)

Warnings in cuda12.0 plus icx2023 (clang16, gcc12.1)
```
ccache /usr/local/cuda-12.0/bin/nvcc  -O3  -lineinfo -I. -I../../src -I../../../../../tools -I/usr/local/cuda-12.0/include/ -DUSE_NVTX -gencode arch=compute_70,code=compute_70 -gencode arch=compute_70,code=sm_70 -use_fast_math -std=c++17  -ccbin /cvmfs/projects.cern.ch/intelsw/oneAPI/linux/x86_64/2023/compiler/2023.0.0/linux/bin/icpx -allow-unsupported-compiler -DMGONGPU_FPTYPE_DOUBLE -DMGONGPU_FPTYPE2_DOUBLE -Xcompiler -fPIC -c gcheck_sa.cu -o gcheck_sa.o
/usr/local/cuda-12.0/include/thrust/detail/type_traits.h:139:107: warning: builtin __has_trivial_constructor is deprecated; use __is_trivially_constructible instead [-Wdeprecated-builtins]
template< class T> struct has_trivial_constructor : public integral_constant< bool, is_pod< T> ::value || __has_trivial_constructor(T)>  { 
                                                                                                          ^
/usr/local/cuda-12.0/include/thrust/detail/type_traits.h:155:112: warning: builtin __has_trivial_copy is deprecated; use __is_trivially_copyable instead [-Wdeprecated-builtins]
template< class T> struct has_trivial_copy_constructor : public integral_constant< bool, is_pod< T> ::value || __has_trivial_copy(T)>  { 
                                                                                                               ^
/usr/local/cuda-12.0/include/thrust/detail/type_traits/has_trivial_assign.h:33:131: warning: builtin __has_trivial_assign is deprecated; use __is_trivially_assignable instead [-Wdeprecated-builtins]
[...]

PR #593, opened 2023-02-23
Add support and logs for icpx 2023

valassi:
Add support and logs for icpx 2023

This is the setup
/cvmfs/projects.cern.ch/intelsw/oneAPI/linux/x86_64/2023/setvars.sh

This is wtill WIP, I need to add the logs


valassi:
For the moment I have fixed the warning in #592 
https://github.com/madgraph5/madgraph4gpu/pull/593/commits/8ca4de967cc090815a3bc4275a117c8c2609d72a

valassi:
Note: there are still warnings about ffp instead 
```
icpx: warning: overriding '-ffp-contract=fast' option with '-ffp-contract=on' [-Woverriding-t-option]
```
This is in #516, which remains open

valassi:
I have committed all logs.

All in all icx2023 has a very similar performance to clang14.
This is not very surprising as it is based on clang - but in the past I had the impression icx was even better than clang, which is no longer the case.

valassi:
All is ok, I self merge

PR #596, opened 2023-02-24
Fix OpenMP link options for clang and Intel icx - and fix helicity selection with OMP

### LINKED ISSUE #578: Fix OMP link errors in clang and icpx for cudacpp.mk
Fix OMP link errors in clang and icpx for cudacpp.mk

I have reenable OMP in cudacpp. It works in gcc, but link fails in icpx and clang.

Note that the fortran madevent links ok on both clang and icpx instead with openmp, it is only cudacpp that gives issues.

Note that there is a choice of linkers: gcc, nvcc or gfortran are used in different targets, for instance I think the madeent uses gfortran/FC.

I have fixed icpx without cuda, but not with cuda. As for clang I am not sure (I fixed icpx with static-intel, so not a clang solution)

valassi:
Fix OpenMP link options for clang and Intel icx - fixes #578


valassi:
The tests were initially failing because for Apple clang it still fails. I have disabled OMP in Apple clang - tant pis.

En passant instead I also did a few tests of madevent with OMP multithreading. I found out there was a rather serious bug in helicity selection (some variables were shared but should have been private, or better moved inside the loop), see #597. This is now also fixed

valassi:
All checks have passed, I self merge

PR #617, opened 2023-03-30
WIP: Added PEP files to separate /tools/ subdirectory

zeniheisser:
Added the two libraries I've been working on for the reweighting project to a distinct subdirectory of .../tools/, called /tools/PEP/. Currently just includes the LHE parser (PEP) and the reweighting lib (PEPPER), both WIP


zeniheisser:
@valassi  yes, there was a preamble in the previous one I forgot to remove which should not have been there. If you want to take a look, this is the rq to look at!

hageboeck:
> Hi @zeniheisser thanks. Do I understand correctly that this PR #617 replaces #612?
> 
> By the way, the GPU CI fails, but it has nothing to do with this specific PR. I think that it is an infrastructure issue (@hageboeck maybe you can have a look please? thanks)

The cuda path changed, so it doesn't find it. I will add `/usr/local/cuda/` to PATH now, reboot in 1 minute

hageboeck:
> The cuda path changed, so it doesn't find it. I will add `/usr/local/cuda/` to PATH now, reboot in 1 minute

The CI machine now correctly sets its own PATH when you log in, but in your branch, `/usr/local/cuda-12` is still hardcoded in the CI workflow file.
You have to rebase for the change to take effect.

PR #623, opened 2023-03-31
[WIP] Added performance profiling scripts to tools/profiling

Jooorgen:
Added the performance profiling scripts I have been working on, also implemented changes to the C/C++ CI (Now goes through epochX processes as well) and added CI for SYCL. Also added a script with generating a container for running this infrastructure.

Minor changes to SYCL makefiles also included in this PR to accommodate the profiling infrastructure, namely including a way to inject correct GCC toolchain for compilation.


Jooorgen:
> Revert the changes to the `sycl.mk` makefiles and sync your branch once PR #637 goes through. The added options and directories in this PR #623 should not be hardcoded into the plugin generated directories and would get removed during the next code regeneration. If the GCC toolchain option is needed during compilation, it should be added to the `SYCLFLAGS` environment variable, although it shouldn't be necessary if oneAPI DPC++ was configured properly during installation. Also, device filtering (`ONEAPI_DEVICE_SELECTOR=???:???`) can be used along with other environment variables to get the proper device for the default `--device_id 0`. Note, there is currently a bug in the latest oneAPI DPC++ when compiling the full app, so I have been [using a build from March 10th 2023](https://github.com/intel/llvm/tree/fea99cc9ad676afeb7fa3404f7db598e42804468). Let's connect over email to go over the details of configuring the DPC++ build in the container and setting the appropriate environment variables for the different CIs.

Makes total sense. For ease of use I used oneAPI available from CVMFS, but moving the CI to containers I should build everything in the container for more control instead of using CVMFS. Thanks for the feedback!

PR #619, opened 2023-03-31
WIP: Upgrade to Olivier's latest gpucpp branch upstream ("split_nonidentical_grouping")

valassi:
Upgrade to Olivier's latest gpucpp branch upstream

The only change is that in eemumu.mad the P1_ll_ll subdirectory is now P1_epem_mupmum

Thisis related to #272, see https://github.com/madgraph5/madgraph4gpu/issues/272#issuecomment-1480745508 . Th elast change by @oliviermattelaer now prevents directories with nprocesses>1. See the discussion at the meeting earlier this week https://indico.cern.ch/event/1263518/

PR #624, opened 2023-04-03
Fixes for SUSY code generation of gg_tt

### LINKED ISSUE #622: codegen for susy process fails with "ValueError : not enough values to unpack (expected 2, got 1)"
The codegen for a susy process fails with "ValueError : not enough values to unpack (expected 2, got 1)"

This is an issue that I just discussed with @oliviermattelaer . IIUC this had also been reported by @nscottnichols in a similar setup. Copying also @zeniheisser for info.

We had a look with Olivier. This comes from the codegen for HRDCOD=1. The issue is that the original code includes not only "parameter = value" but also "if ( condition ) parameter = value".

A first quick fix in model_handling.py could be
```
         parset_lines = {}
         for line in parset.split('\n'):
+            if line.startswith('if'): continue
             par, parval = line.split(' = ')
```
However then I get an assert in 
```
        assert( len(pardef_lines) == len(parset_lines) ) # AV sanity check (same number of parameters)
```

I must look at the unmodified code and see what it gives... with HRDCOD=1 we need constexpr stuff, so the counting of parameters must be handled with care...

valassi:
This completes the first "susy" patch: now susy_gg_tt can be generated correctly (but it does not build).
In practice, the main (only?) issue it addresses is that it closes #622.

Further patches (susy2 and possibly more) will attempt to fix these builds.

NB: At this stage, CODEGEN is still using the upstream mg5amcnlo without "split_nonidentical_grouping" (PR #619 and #272)

Copying @oliviermattelaer and @nscottnichols because this may be something similar to issues they were having. (NB: I have many more issues coming...)


valassi:
All tests passe, I will self merge.

Copying also @zeniheisser for info

Issue #614, opened 2023-03-29
Support for non-SM UFO models: HRDCOD=1 build fails in model SMEFTsim_topU3l_MwScheme_UFO


zeniheisser:
Generic issue: processes in non-SM models are not always supported (see also #615 )

Speceific issue: Processes in eg model SMEFTsim_topU3l_MwScheme_UFO can be exported through the plugin, but fail at compile time both due to failed assertions and to non-fitting parameter names, see code snippet for full error message.

```
[zwetters@itscrd80 P1_Sigma_SMEFTsim_topU3l_MwScheme_UFO_gg_ttx]$ make HRDCOD=1
makefile:77: CUDA_HOME was not set: using "/usr/local/cuda-11.8/"
OMPFLAGS=-fopenmp
AVX=512y
FPTYPE=d
HELINL=0
HRDCOD=1
RNDGEN=hasCurand
Building in BUILDDIR=. for tag=512y_d_inl0_hrd1_hasCurand (USEBUILDDIR is not set)
make -C ../../src  -f cudacpp_src.mk
make[1]: Entering directory '/data/zwetters/modded5amc/PROC_PLUGIN_SMEFTsim_topU3l_MwScheme_UFO_3/src'
AVX=512y
g++  -O3  -std=c++17 -I.  -fPIC -Wall -Wshadow -Wextra -ffast-math  -fopenmp -march=skylake-avx512 -mprefer-vector-width=256  -DMGONGPU_FPTYPE_DOUBLE -DMGONGPU_FPTYPE2_DOUBLE -DMGONGPU_HARDCODE_PARAM -c Parameters_SMEFTsim_topU3l_MwScheme_UFO.cc -o Parameters_SMEFTsim_topU3l_MwScheme_UFO.o
In file included from /opt/rh/gcc-toolset-11/root/usr/include/c++/11/cassert:44,
                 from Parameters_SMEFTsim_topU3l_MwScheme_UFO.h:78,
                 from Parameters_SMEFTsim_topU3l_MwScheme_UFO.cc:8:
Parameters_SMEFTsim_topU3l_MwScheme_UFO.h:381:56:   in ‘constexpr’ expansion of ‘Parameters_SMEFTsim_topU3l_MwScheme_UFO::constexpr_pow(2.0e+0, 2.5e-1)’
[...]

PR #625, opened 2023-04-03
fix susy gg_tt builds and add it to the repo

valassi:
This is a followup to #624, where I fixed the code generation of SUSY gg_tt. Here I am also trying to fix the builds, but there are issues, some of which will remain "unfixable" for a while (I will post details: trigonometric functions are not constexpr; and FFV function calls should not include "-COUPS" as these are pointers not values).

Note also, the main reason why I started going into susy processes with PR #624 and this new PR here is that I wanted to check "nprocesses>1" in #272, before PR #619. This PR is currently also including some SM processes to that effect (I will probably split them away).


valassi:
I have cleaned this up, splitting other stuff like the SM gq_ttq to PR #626

valassi:
For the record also keep in mind this comment by Olivier
https://github.com/madgraph5/madgraph4gpu/issues/272#issuecomment-1495526182
"also it seems that some matrix element are missing (the symmetric ones) in the susy case, this need to be fix."

valassi:
I have just merged the latest master into this PR (which contains a new sa generated code dir, so there are no conflicts).

Note that the code was generated with old code, so this cannot be directly trasnposed to codegen

valassi:
I have updated this PR to our latest code bases and updated the PR. A few findings

- Code generation for susy gg->tt succeeds

- In HRDCOD=0 mode, the build fails (but I need to investigate... in any case there is a hardcoded "use HRDCOD=1 instead" to remove)

- In HRDCOD=1 moe, nvcc builds fail because sin/cos/atan are not constexpr (see #360 

- However in HRDCOD=1 mode, no-CUDA builds with C++ only do succeed! See https://github.com/madgraph5/madgraph4gpu/pull/625/commits/415e27bcefd62f31bde199618311ee0d17342a12 and https://github.com/madgraph5/madgraph4gpu/pull/625/commits/d5ded87a291fe7b6779d8f7cb9e1d0b6c30ec0af. This might mean that it might be possible to hardcode the constexpr implementation from gcc into the nvcc part of our code if we wanted this?

In any case, the real goal here (also to allow @zeniheisser 's reweighting) is to allow HRDCOD=0 builds.

valassi:
A further update for HRDCOD=0.
- I have "temporarely" removed the error "you should use HRDCOD=1", see https://github.com/madgraph5/madgraph4gpu/pull/625/commits/717a5deaffd3f6c9fda58d57629fa817cf8b91fe: this needs a cleanup, make it a warning? only for cuda?
- I have fixed the first of two issues, the visibility of mdl_I51x11, see https://github.com/madgraph5/madgraph4gpu/pull/625/commits/df8b87a9dd01eb23a010c43cf5f21097af0f3f5a. I had to use a getInstance() accessor to a static class. This goes through in c++, but it gives a warning in cuda, which I suspect is the reason why MEs are all 0 in cuda at runtime,
- I have fixed the second of two issues, by defining an "indices" variable that was commented out, see https://github.com/madgraph5/madgraph4gpu/pull/625/commits/4f3b83a0a3cea27a9784075eae17a48b27c2d368. This is ok here, but maybe should remain commented out in other codes where this is not needed (or an unused variable warning must be silenced?). This neeeds some cleanup before the backport to codegen.

Anyway, all in all, this is now the matrix of where we are with this:
- C++ builds and tests are now fully ok both with HRDCOD=0 and HRDCOD=1. This is very good, so I guess this PR can evolve from WIP and be merged at some point.
- CUDA builds fail with HRDCOD=1 because of sin/cos/atan not constexpr (#627 )
- CUDA builds succeed with HRDCOD=0, but MEs are all 0 at runtime. I guess that a better structure of parameter classes is needed here, to avoid that getInstance which host/device code does not like? Or this mdl_I51x11 must be also copied from the host to the device? I should cross check again where the alphas running happens in cuda, and if it is on the host then it is just the coupling that must be transferred to the device...

valassi:
Question for @oliviermattelaer : the code generation here risks to be tricky, because some parameters are involved in alphas running and others are not. I have not started to look at this yet, maybe it is obvious and easy, but in case do you have advice/suggestions on how to identify variables such as `mdl_I51x11`? This enters the calculatiion here
```
    // Model couplings dependent on aS
    //constexpr cxsmpl<double> GC_6 = -G; // now computed event-by-event (running alphas #373)
    //constexpr cxsmpl<double> GC_51 = -( mdl_complexi * G * mdl_I51x11 ); // now computed event-by-event (running alphas #373)
...
#ifdef MGONGPU_HARDCODE_PARAM
      using namespace Parameters_MSSM_SLHA2;
#else
      const double mdl_I51x11 = Parameters_MSSM_SLHA2::getInstance()->mdl_I51x11; // fix HRDCOD=0 susy builds
#endif
      // NB: hardcode cxtype cI(0,1) instead of cxtype (or hardcoded cxsmpl) mdl_complexi (which exists in Parameters_MSSM_SLHA2) because:
      // (1) mdl_complexi is always (0,1); (2) mdl_complexi is undefined in device code; (3) need cxsmpl conversion to cxtype in code below
      const cxtype cI( 0., 1. );
      DependentCouplings_sv out;
      // Begin non-SM (e.g. EFT) implementation - special handling of vectors of floats (#439)
#if not( defined MGONGPU_CPPSIMD && defined MGONGPU_FPTYPE_FLOAT )
      {
        const fptype_sv& G = G_sv;
        // Model parameters dependent on aS
        //const fptype_sv mdl_sqrt__aS = constexpr_sqrt( aS );
        //const fptype_sv G = 2. * mdl_sqrt__aS * constexpr_sqrt( M_PI );
        const fptype_sv mdl_G__exp__2 = ( ( G ) * ( G ) );
        // Model couplings dependent on aS
        out.GC_6 = -G;
        out.GC_51 = -( cI * G * mdl_I51x11 );
      }
#else
...
```
Thanks Andrea

valassi:
Rephrasing. I think that this model parameter should end up in tIPD, so that it can get copied from host to device an dbe available in `computeDependentCouplings_fromG`, which is called _on the device_ inside `G2COUP`. These are the parameters that presently end up in tIPD:
```
[avalassi@itscrd90 gcc12.1/cvmfs] /data/avalassi/GPU2023/madgraph4gpuBis/epochX/cudacpp> grep 'const fptype tIPD' *.sa/Sub*/P*/CPP*
ee_mumu.sa/SubProcesses/P1_Sigma_sm_epem_mupmum/CPPProcess.cc:    const fptype tIPD[2] = { (fptype)m_pars->mdl_MZ, (fptype)m_pars->mdl_WZ };
gg_ttggg.sa/SubProcesses/P1_Sigma_sm_gg_ttxggg/CPPProcess.cc:    const fptype tIPD[2] = { (fptype)m_pars->mdl_MT, (fptype)m_pars->mdl_WT };
gg_ttgg.sa/SubProcesses/P1_Sigma_sm_gg_ttxgg/CPPProcess.cc:    const fptype tIPD[2] = { (fptype)m_pars->mdl_MT, (fptype)m_pars->mdl_WT };
gg_ttg.sa/SubProcesses/P1_Sigma_sm_gg_ttxg/CPPProcess.cc:    const fptype tIPD[2] = { (fptype)m_pars->mdl_MT, (fptype)m_pars->mdl_WT };
gg_tt.sa/SubProcesses/P1_Sigma_sm_gg_ttx/CPPProcess.cc:    const fptype tIPD[2] = { (fptype)m_pars->mdl_MT, (fptype)m_pars->mdl_WT };
gq_ttq.sa/SubProcesses/P1_Sigma_sm_gu_ttxu/CPPProcess.cc:    const fptype tIPD[2] = { (fptype)m_pars->mdl_MT, (fptype)m_pars->mdl_WT };
gq_ttq.sa/SubProcesses/P1_Sigma_sm_gux_ttxux/CPPProcess.cc:    const fptype tIPD[2] = { (fptype)m_pars->mdl_MT, (fptype)m_pars->mdl_WT };
heft_gg_h.sa/SubProcesses/P1_Sigma_heft_gg_h/CPPProcess.cc:    //const fptype tIPD[0] = { ... }; // nparam=0
susy_gg_tt.sa/SubProcesses/P1_Sigma_MSSM_SLHA2_gg_ttx/CPPProcess.cc:    const fptype tIPD[2] = { (fptype)m_pars->mdl_MT, (fptype)m_pars->mdl_WT };
``` 

In summary, todo:
- first, develop the actual generated code for this susy process with `mdl_I51x11` as part of `tIPD`
- second, understand what needs to be done in code generation so that this parameter ends up in tIPD...

valassi:
> In summary, todo:
> 
>     * first, develop the actual generated code for this susy process with `mdl_I51x11` as part of `tIPD`
> 
>     * second, understand what needs to be done in code generation so that this parameter ends up in tIPD...

Ok I have indeed fixed CUDA HRDCOD=0 builds by adding `mdl_I51x11` as part of `tIPD`.

The only pending problem are CUDA HRDCOD=1 builds because of missing constexpr sin/cos/atan (https://github.com/madgraph5/madgraph4gpu/issues/627)

This is probably ready to be backported to CODEGEN and eventually merged

valassi:
Replying to the original description of this PR:

> This is a followup to #624, where I fixed the code generation of SUSY gg_tt. Here I am also trying to fix the builds, but there are issues, some of which will remain "unfixable" for a while (I will post details: trigonometric functions are not constexpr; and FFV function calls should not include "-COUPS" as these are pointers not values).

Two issues are mentioned above:
-  trigonometric functions are not constexpr, this is still the case and a blocker for HRDCOD=1 builds but only in cuda (#627), while it is not a problem in c++ (at least in gcc 12.1)
- the function calls which includ `-COUPS` have been fixed (#628)

> Note also, the main reason why I started going into susy processes with PR #624 and this new PR here is that I wanted to check "nprocesses>1" in #272, before PR #619. This PR is currently also including some SM processes to that effect (I will probably split them away).

The `nprocesses>1` is now in the repo from gq_ttq, but the SM version was enough, there was no need to use the susy version of that.

valassi:
I have made progress both on CODEGEN and on constexpr trig functions. Still need some work, but should all be doable.

valassi:
The fix for HRDCOD=1 builds in SUSY is now complete
- I implemented the relevant trig functions as constexpr using Taylor series
- I added __device__ to the relevant constexpr variables
This closes #627.

I only need to complete the CODEGEN backport (almost done, just some details missing).

valassi:
I have finally completed the backport to code generation.

I need to make sure that this also works for other processes now (it does not, the CI is failing)

valassi:
I added a few fixes for ee_mumu, now relaunching the CI - and doing more manual tests

valassi:
I have added a few fixes for FPTYPE=f, now relaunching the CI tests

valassi:
The CI was failing for Mac because it does not know M_PIl from cmath.
I thought of std::numbers::pi but this needs c++20 and we are still using c++17.
I could use my constexpr_atan(1)*4... but a literal is better? So I went for literals (the same as in cmath).
Relaunching all tests in the CI.

valassi:
And I had forgotten another M_PIl that was failing on Mac. Relaunching the CI again.

valassi:
And... testing a fix for yet another issue on Mac

valassi:
Hi @oliviermattelaer this is now ready for review - can you please review it?
(It is also ~almost ready for merging, I want to run my long manual tests first, it will not be ready before the weekend).

Essentially: **SUSY gg to tt now builds and runs ok, both in HRDCOD=1 and HRDCOD=0**.
I also tested FPTYPE=f and Mac, through the CI. (cc @roiser)

The main issue was CODEGEN. I had to find a way to identify which independent BSM parameters are used in the computation of the depetdent couplings from G. This is done. It is quite ugly and certainly improvable, but it works. One problem I had is that I find out the relevant paraeters in a piece of code that creates Param.h, but by that time CPPProcess.cc has already been created... so I had to do some gymnastics back and forth.

About HRDCOD=1, I solved the missing constexpr by defining them myself, through Taylor series approximations. Very fast code, and heavily tested now.

One unfortunate issue is that all this work still does not fix SMEFT (cc @zeniheisser). I tested this, but there are still quite separate issues #614 and #616. That will need some more work, in a separate PR (WIP #632).

Anyway, having at least one BSM process with SUSY is better than having none at all I think.

I will run my tests in the meantime, let me know if you have comments. Thanks.

valassi:
Hi @oliviermattelaer I confirm that this is ready to be merged.

NB This fixes gg_tt and gq_ttq in SUSY (so not even SUSY processes itself), and partly gg_gogo. But many other issues remain, see #820. I would still suggest to merge this one already, and then investigate the rest later on.

valassi:
Note: this is a standalone PR but is also included in #824, which adds additional fixes for SUSY.

Rephrasing:
- after this PR #625 some SUSY issues are fixed, but there are many pending
- after PR #824, all SUSY issues I identified are fixed

epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/model_handling.py
```diff
@@ -716,9 +721,13 @@ def write_hardcoded_parameters(self, params):
             res = '// (none)\n'
             return res
         pardef = pardef.replace('std::complex<','cxsmpl<') # custom simplex complex class (with constexpr arithmetics)
-        parset = parset.replace('std::complex<','cxsmpl<') # custom simplex complex class (with constexpr arithmetics)
+        parset = parset.replace('std::complex<','cxsmpl<') # NB this is really needed twice! (if there are two std::complex on the same line)
```
oliviermattelaer, line 724: Are you sure?
https://www.w3schools.com/python/ref_string_replace.asp
said that "ALL occurrences" will be replaced. Or am I wrong that parset is of type string?

epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/model_handling.py
```diff
@@ -716,9 +721,13 @@ def write_hardcoded_parameters(self, params):
...
         parset = parset.replace('sqrt(','constexpr_sqrt(') # constexpr sqrt (based on iterative Newton-Raphson approximation)
-        parset = parset.replace('pow(','constexpr_pow(') # constexpr sqrt (based on iterative Newton-Raphson approximation)
+        parset = parset.replace('pow(','constexpr_pow(') # constexpr pow
+        parset = parset.replace('atan(','constexpr_atan(') # constexpr atan for BSM #627
+        parset = parset.replace('sin(','constexpr_sin(') # constexpr sin for BSM #627
+        parset = parset.replace('cos(','constexpr_cos(') # constexpr cos for BSM #627
+        parset = parset.replace('tan(','constexpr_tan(').replace('aconstexpr_tan(','atan(') # constexpr tan for BSM #627
         parset = parset.replace('(','( ')
         parset = parset.replace(')',' )')
         parset = parset.replace('+',' + ')
```
oliviermattelaer, line 733: Stating the obvious here, this start to be a smelling code.
I guess that next time we should 
1) move to re replacement with substitution (will be already a great improvment)
2) potentially change the expression upstream where such rule are likely existing --so likely just need to overwrite how such mathematical expression are converted from the token representation to those expression.

But I agree -> for later

epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/model_handling.py
```diff
@@ -736,7 +745,12 @@ def write_hardcoded_parameters(self, params):
             type, pars = line.rstrip(';').split(' ') # strip trailing ';'
             for par in pars.split(','):
                 ###print(len(pardef_lines), par) # for debugging
-                pardef_lines[par] = ( 'constexpr ' + type + ' ' + par )
+                if par in majorana_widths:
+                    pardef_lines[par] = ( 'constexpr ' + type + ' ' + par + "_abs" )
+                elif par in deviceparams:
```
oliviermattelaer, line 750: Can you comment on the reason of handling majorana width in a special way? I'm just puzzled by it

epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/model_handling.py
```diff
@@ -888,12 +926,30 @@ def super_generate_parameters_class_files(self):
             replace_dict['dcoupsetdcoup2'] = '      // (none)'
             replace_dict['dcoupoutdcoup2'] = ''
         # Require HRDCOD=1 in EFT and special handling in EFT for fptype=float using SIMD
+        replace_dict['bsmdefine'] = '#define MGONGPUCPP_NBSMINDEPPARAM_GT_0 1' if len( bsmparam_indep_real_used ) > 0 else '#undef MGONGPUCPP_NBSMINDEPPARAM_GT_0'
         if self.model_name[:2] == 'sm' :
```
oliviermattelaer, line 930: This condition is really dangerous since some model can be 
sm_with_zprime

epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/model_handling.py
```diff
@@ -888,12 +926,30 @@ def super_generate_parameters_class_files(self):
...
+    // BSM parameters that do not depend on alphaS but are needed in the computation of alphaS-dependent couplings;
+    static constexpr int nBsmIndepParam = %s;
+    %sdouble mdl_bsmIndepParam[nBsmIndepParam];''' % ( len( bsmparam_indep_real_used ), '' if len( bsmparam_indep_real_used ) > 0 else '//' )
+            replace_dict['bsmip1'] = '''\n
+    // BSM parameters that do not depend on alphaS but are needed in the computation of alphaS-dependent couplings;
+    constexpr int nBsmIndepParam = %s;
+    %s__device__ constexpr double mdl_bsmIndepParam[nBsmIndepParam]%s;''' % ( len( bsmparam_indep_real_used ), '' if len( bsmparam_indep_real_used ) > 0 else '//', ' = { %s }' % ', '.join( bsmparam_indep_real_used ) if len( bsmparam_indep_real_used ) > 0 else '' )
+            replace_dict['eftwarn0'] = '\n//#warning Support for non-SM physics processes (e.g. SUSY or EFT) is still limited for HRDCOD=0 builds (#439 and PR #625)'
+            replace_dict['eftwarn1'] = '\n//#warning Support for non-SM physics processes (e.g. SUSY or EFT) is still limited for HRDCOD=1 builds (#439 and PR #625)'
+            if len( bsmparam_indep_real_used ) == 0:
```
oliviermattelaer, line 949: Is those warning still making sense?

epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/model_handling.py
```diff
@@ -1144,6 +1200,14 @@ def get_process_function_definitions(self, write=True):
             replace_dict['cipd2tipd'] = '//memcpy( cIPD, tIPD, %i * sizeof( fptype ) ); // nparam=0'%len(params)
             replace_dict['cipddump'] = ''
             replace_dict['cipdhrdcod'] = '//__device__ const fptype* cIPD = nullptr; // unused as nparam=0'
+        if self.model_name[:2] == 'sm' :
```
oliviermattelaer, line 1203: same comment here ... 
to add details, it is likely better to trigger this depending on the presence/absence of parameter that can not be handle like the SM...

epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/model_handling.py
```diff
@@ -789,14 +803,31 @@ def super_write_set_parameters_onlyfixMajorana(self, hardcoded): # FIXME! split
...
+        # because they are used in the event by event calculation of alphaS-dependent couplings
+        # WARNING! This is only implemented and has only been tested so far for real parameters (complex parameters need twice the storage)
+        bsmparam_indep_real_used = []
+        if self.model_name[:2] != 'sm' :
+            for param in self.params_indep:
+                if param.type == 'real':
+                    for coup in self.coups_dep.values():                
+                        if param.name in coup.expr:
+                            bsmparam_indep_real_used.append( param.name )
+            bsmparam_indep_real_used = set( bsmparam_indep_real_used )
```
oliviermattelaer, line 829: please made the code to crash if param.type != 'real'
such that we now at generation time that we have to fix the issue...

oliviermattelaer (commented):
Here I'm struggling to understand two points (sorry for that)
 - what is problematic/special with the width of majorana particle? (i.e. do we really need to separate those from the other widths?)
 - What is the issue with BSM coupling? They should also split in running/non running coupling (like the SM does)?  

Typically, I try to avoid to single out SM compare to BSM model (my typical strategy is saying that SM is a BSM model like any other), but this is not a require strategy.
However, using the model name to check for SM model is likely a bad idea since model can be loop_sm and still be SM ... (and the opposite is True too, as written in one comment)

I will check some of the issue linked to see if I can figure it out on my own but comment on this already (maybe I should have done the opposite, sorry in case)

epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/model_handling.py
valassi: Thanks Olivier, yes I agree... there are many many many things that should be done in the code upstream, rather than in the plugin. But lets keep it for later.

(It also depends if we want to keep cudacpp as a plugin, or move it to replace the existing export_cpp.py. I imagine I have a slight preference for keeping the plugin: in that case, "cleaning up" would mean hardcoding all of export_cpp in the plugin, and fixing it immediately instead of overloading methods as it is done here? Anyway, for later...)

epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/model_handling.py
valassi: Ah... interesting. What do you suggest instead? Spell out the full SM model name and compare to that string? On the other hand sm_nobmass requires no special handling.

Maybe I should check what happens if I do this "extra BSM handling" also for all SM processes. I guess that it will add some unused and unnecessary boilerplate code, but will not change the functionality.

Suggestions?

epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/model_handling.py
valassi: Well... SUSY now is fixed (touch wood...). SMEFT is not fixed, I hope it can be fixed soon, but will need some more work. The comment is commented out anyway... but just a reminder that there is some stuff not yet functional

epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/model_handling.py
valassi: Hm ok yes. I see what you mean, a bit what  was saying above
- do some extra "BSM-like" checks all the time, also in SM processes, and see what happens
- if some extra non-SM Parameters are identified, then do some special handling...

I need to try this out in practice.

This is a big change. I would try this out only directly on the second more comprehensive PR #824... or, even merge #824 and then do this in a later step...? Any preferences?

valassi:
> Here I'm struggling to understand two points (sorry for that)

No need to be sorry!

>     * what is problematic/special with the width of majorana particle? (i.e. do we really need to separate those from the other widths?)

What is special is that there is a need for some code that does "if mass<0 then width=-width", i.e. if the mass is less than 0 then also the width is less than 0.

What is problematic is that, before various fixes, there were some code generation errors. I saw this in #622 for cudacpp, while Walter and Nathan also sa it in #694 for the SYCL plugin.

Now, why I chose to split one python method into two methods "without Majorana fixes" and "only apply Majorana fixes", unfortunately I do not remember. Maybe I could have kept those in the same method, I am not sure. The good thing is that the patch works and fixed some issues.

On top of this, in addition to #622 codegen error "ValueError : not enough values to unpack", later on I got some other errors specific to this code. One was for instance that, in the HRDCOD=1 setup, you cannot right "if m<0 then w=-w", because this is not a constexpr assignment. So I rewrote this as "w = ( m<0 ? -absw : + absw )" or something similar.

Finally, note that the "if mass<0 then width=-width" machinery is something that I found in the original code, I did not write it myself (it comes from UFO? from code you wrote? no idea...). I just made sure that the logic works ok now in all codegen and builds.

Does this answer your question?

>     * What is the issue with BSM coupling? They should also split in running/non running coupling (like the SM does)?

Listen, here I also have a hard time myself. I have done enough reverse engineering to fix the issues, but I must say that I still do not understand the logic. "Experimentally", I just see that BSM models introduce some parameters that are handled by very different parts of the python code, at different moments in time.

We should probably sit down and take a look at some point...?

If you want to get a better idea, anyway, I suggest you look at what appears in a generated SUSY process. For instance compare susy gg to stop pairs with SM gg to top pairs? This is from the more recent PR #824,
https://github.com/valassi/madgraph4gpu/blob/susy/epochX/cudacpp/susy_gg_t1t1.sa/src/Parameters_MSSM_SLHA2.h
https://github.com/valassi/madgraph4gpu/blob/susy/epochX/cudacpp/susy_gg_t1t1.sa/src/Parameters_MSSM_SLHA2.cc
https://github.com/valassi/madgraph4gpu/blob/susy/epochX/cudacpp/susy_gg_t1t1.sa/SubProcesses/P1_Sigma_MSSM_SLHA2_gg_t1t1x/CPPProcess.cc
vs
https://github.com/valassi/madgraph4gpu/blob/susy/epochX/cudacpp/gg_tt.sa/src/Parameters_sm.h
https://github.com/valassi/madgraph4gpu/blob/susy/epochX/cudacpp/gg_tt.sa/src/Parameters_sm.cc
https://github.com/valassi/madgraph4gpu/blob/susy/epochX/cudacpp/gg_tt.sa/SubProcesses/P1_Sigma_sm_gg_ttx/CPPProcess.cc

What is especially interesting are the Parameters.h and the CPPProcess.cc files. (The Parameters.cc is less interesting).

I think that in the end there are three main differences? maybe I am missing some
- the majorana stuff already discussed, something here is unavoidable
- the fact that SIMD vectors of floats need some special handling (issue #439... this I fixed almost two years ago, May 2022, so I do not remember the details), something here is also unavoidable I guess
- the fact that there are some parameters `double mdl_bsmIndepParam[nBsmIndepParam] = { mdl_I52x33.real(), mdl_I52x33.imag(), mdl_I51x33.real(), mdl_I51x33.imag(), mdl_I75x33.real(), mdl_I75x33.imag(), mdl_I74x33.real(), mdl_I74x33.imag(), mdl_I52x36.real(), mdl_I52x36.imag(), mdl_I51x36.real(), mdl_I51x36.imag() };` that I copy to constant memory like we do for cIPC SM parameters, but which do not go through the same code... a legitimate question is, "could we not handle those like the cIPC?", the answer maybe is yes, but somehow the code that manages the cIPC does not manage them...? again, I got lost in the overall logic of the python, which function controls which parameters

Along the same lines, one other thing that so far I was unable to understand is the order in which these parameters are discovered, and the collaboration bwteen the UFOconverter, the HelasWriter and the other classes. In particular, I have this ugly comment, which shows that I do not completely master the order of generation of Parameters.h and CPPProcess.cc
```
  // AV Jan 2024 (PR #625): this ugly #define was the only way I found to avoid creating arrays[nBsm] in CPPProcess.cc if nBsm is 0
  // The problem is that nBsm is determined when generating Parameters.h, which happens after CPPProcess.cc has already been generated
  // For simplicity, keep this code hardcoded also for SM processes (a nullptr is needed as in the case nBsm == 0)
#ifdef MGONGPUCPP_NBSMINDEPPARAM_GT_0
```

Anyway, as above: it works now! So, it can be kept as a cleanup for later... unless you see obvious ways to improve this fast.

> 
> Typically, I try to avoid to single out SM compare to BSM model (my typical strategy is saying that SM is a BSM model like any other), but this is not a require strategy. However, using the model name to check for SM model is likely a bad idea since model can be loop_sm and still be SM ... (and the opposite is True too, as written in one comment)

Ok yes I agree, as written in the comment above. I can try to apply the "special handling" all the time, and see if it still produces SM code which makes sense.

> I will check some of the issue linked to see if I can figure it out on my own but comment on this already (maybe I should have done the opposite, sorry in case)

Voila I commented on everything, hope this helps. I am sure you have many comments on my comments...

Thanks!

oliviermattelaer:
Ok thanks a lot,

This is helpfull, Maybe the only change that need to be done (but not required to be done in this PR) is to run all the BSM trick for the SM model as well and check that SM is still working correctly.

We can also sit together on Friday to see if I can design a quick and  nice way to fix your issue with those coupling. But, I guess that the answer will not be quick...

One reason of the complexity is that the structure of the code is likely:
1) read the UFO model -> convert it to a MG5 object
2) potential optimization of the model by MG5
3) generate the diagrams
4) generate the helas call -> write the process file
5) retrieve from step 3 or 4 the list of couplings/lorentz structure needed for the particular process and create code to compute only such list. (Note that all parameters are handle separetly and likely do not have such optimization ...)
 
So I guess the confusion might partly comes from the difference between coupling and parameter (and the difference is in a way arbitrary) and  the fact that some parameter can/should be real (coupling should always be treated as complex --even if they might sometimes/often be real or pure imaginary--).

But in itself you can consider this as an approve :-)

Cheers and thanks,

Olivier

valassi:
> Ok thanks a lot,
> 
> This is helpfull, Maybe the only change that need to be done (but not required to be done in this PR) is to run all the BSM trick for the SM model as well and check that SM is still working correctly.
...
> 
> But in itself you can consider this as an approve :-)
> 
> Cheers and thanks,
> 
> Olivier

Hi Olivier, thanks!
LIsten, I would do the following:
- consider this PR #625 as approved and merge it
- then work more on PR #824 and try to include the BSM tricks also in all SM and see what it gives
- after I merge #625, it becomes easier for you to review the additional changes that are already in PR #824 (and any new changes I will add)

Does this sound ok?

> We can also sit together on Friday to see if I can design a quick and nice way to fix your issue with those coupling. But, I guess that the answer will not be quick...

Thanks, it would help to sit down together. But today I am just too busy, maybe next week? 

> 
> One reason of the complexity is that the structure of the code is likely:
> 
>     1. read the UFO model -> convert it to a MG5 object
> 
>     2. potential optimization of the model by MG5
> 
>     3. generate the diagrams
> 
>     4. generate the helas call -> write the process file
> 
>     5. retrieve from step 3 or 4 the list of couplings/lorentz structure needed for the particular process and create code to compute only such list. (Note that all parameters are handle separetly and likely do not have such optimization ...)
> 
> 
> So I guess the confusion might partly comes from the difference between coupling and parameter (and the difference is in a way arbitrary) and the fact that some parameter can/should be real (coupling should always be treated as complex --even if they might sometimes/often be real or pure imaginary--).

I will add these comments to #824 so that we have them there for further discussion.

valassi:
> > Ok thanks a lot,
> > This is helpfull, Maybe the only change that need to be done (but not required to be done in this PR) is to run all the BSM trick for the SM model as well and check that SM is still working correctly.
> > ...
> > But in itself you can consider this as an approve :-)
> > Cheers and thanks,
> > Olivier
> 
> Hi Olivier, thanks! Listen, I would do the following:
> 
>     * consider this PR #625  as approved and merge it

I will do the above: consider this as approved, and merge it.

I will move the pending comments and todo to #824.

PR #626, opened 2023-04-03
Add SM gq to ttq (and gq to ttllq) - example of a process with DSIG1 and DSIG2

valassi:
This is a WIP PR adding SM gq to ttq and gq to ttllq.

The first is a deminstrator for the need of split_nonidentical_grouping upstream #619
- there is DSIG2 not interfaced to our functionality in cudacpp
- the build fails
This will need to be reassessed after #619 
- note that all tput/tmad scripts need to be changed to use two P1 subdirectories

The second is another test I had made before I understood that gq to ttq was enough for #272 and #619. It is a demonstrator of another issue that I need to file (FFV should not include "-COUP")

Note that both of these processes I had initially included as SUSY rather than SM (inspired from the cards of @nscottnichols in https://github.com/nscottnichols/madgraph4gpu/blob/f9689784390e2975a055711b3a8d79f7b7688860/epochX/sycl/CODEGEN/generateAndCompare2.sh#L58). But in the end a SM process is enough

Copying @oliviermattelaer and @zeniheisser for info

I will add even more issues for SUSY


valassi:
Notes from trying to reverse engineer DSIG2.

The 2 in DSIG2 is `proc_id` from https://github.com/mg5amcnlo/mg5amcnlo/blob/60fae94f7aafbe8b3a2e2a0cff322bb65fa8d3df/madgraph/iolibs/template_files/auto_dsig_v4.inc#L1

That proc_id comes IIUC from `ProcessExporterFortranME.write_auto_dsig_file` in https://github.com/mg5amcnlo/mg5amcnlo/blob/60fae94f7aafbe8b3a2e2a0cff322bb65fa8d3df/madgraph/iolibs/export_v4.py#L4857

In particular proc_id is `ime+1` from https://github.com/mg5amcnlo/mg5amcnlo/blob/60fae94f7aafbe8b3a2e2a0cff322bb65fa8d3df/madgraph/iolibs/export_v4.py#L6175

where `ime` is `for ime, matrix_element in enumerate(matrix_elements):` in https://github.com/mg5amcnlo/mg5amcnlo/blob/60fae94f7aafbe8b3a2e2a0cff322bb65fa8d3df/madgraph/iolibs/export_v4.py#L6107

and higher up `matrix_elements = subproc_group.get('matrix_elements')` from `ProcessExporterFortranMEGroup.generate_subprocess_directory(self, subproc_group, ...)`

valassi:
I have tested that Olivier's "split_nonidentical_grouping" #619 fixes the issues with DSIG2 in gq_ttq.mad (I generated it before the patch, with DSIG2 and build errors, and after the patch, with two P1 directpries and no build errors).

I will now also remove all other 'nprocesses' comments for #272

valassi:
After merging #619 yesterday, I will now self-merge this one, which is complete.

This adds gq_ttq, which is one example of a process where, BEFORE MR 619, there was only one P1 with DSIG1 and DSIG2. The patches I merge here include code before and code before MR 619. Conversely, AFTER MR 619, there are two P1 each with a single DSIG1. This is (I think) what the "nprocesses>1" of #272 was about. I have also cleaned up all comments about "nprocesses>1". See also #343" that is related.

Note, I have modified tput and tmad scripts to handle also gqttq (for the moment, using only oine of the two P1 directories). ne problem is that the fortran and cpp xsecs differ: this however will be handled in a separate issue #630.

Self-merging.

PR #637, opened 2023-04-11
[SYCL] Add vector type support plus other changes

nscottnichols:
@valassi @jtchilders This PR should only effect the SYCL branch, so can one of you merge once verified?


valassi:
Thanks @nscottnichols indeed this only touches epochX/sycl files - I approved it to run (strange, I thought it would run anyway and would only need approval to merge? @hageboeck any comment?)

I will merge it once the tests pass (which should normally be the case as this PR does not affect them)

valassi:
OK all tests passed, I am merging - thanks @nscottnichols 

cc @oliviermattelaer @roiser @jtchilders

Issue #613, opened 2023-03-29
generated code must be relocatable (must build outside the madgraph4gpu project)


valassi:
This is an essential issue for the release, which I gave for granted (in point 3 of the workplan #576) but probably I did not write down explicitly: the generated code must be relocatable. ie it muct build outside the madgraph4gpu project

Presently, instead, there are several things that we take from the top of the madgraph4gpu project. Two notable examples
- the common random numbers
- the googletest infrastructure

I guess in particular we could
- move the common random header inside each and every process directory
- move any pending test headers inside each and every process directory (already partially done at the time of #268 for MadgraphTest.h)
- develop a way to allow a SINGLE external googletest directory (as is the case now), possibly involving symlinks... I do NOT want to have every single build redownload and reconfigure google test

Thanks @zeniheisser for reporting the issues related to this!

valassi:
Hi @zeniheisser this is related to the iussue we just discussed about common random numbers....

valassi:
The first point in this issue (CommonRandomNumbers.h) is fixed by MRs #657 and #662

The issue remains open as we still need to fix googletest

valassi:
I started work on a WIP MR #666 for googletest.

There are a couple of decisions to take there, for builds outside madgraph4gpu and for new processes

1. The tests in runTests include minor process-independent tests but especially a process-specific test that depends on the existence of a reference file. The reference currently exists for the processes we have in the cudacpp repo (and a selected few more), but clearly this will NOT be produced for every new process. Currently the test fails if the reference file does not exists. So things to decide are for instance, what to do for new processes?
- Should the test app keep failing if the ref is not found, or should it issue a warning? This is probably irrelevant as somehow we will not run the test routinely, see below. I would keep the error: a test with no ref is useless anyway, we will probably not run it or not even build it!
- My feeling is that we will only routinely use and run runTests,exe inside madgraph4gpu, but not outside. If at some point we add a new process to madgraph4gpu as something to focus on, we will probably create a reference file for it: so inside madgraph4gpu we should alway build the test, and have an error if the ref does not exist. Outside madgraph4gpu, conversely, we should probably not build the tests if there is no ref file, and we may even decide not to build the tests at all unless for special cases?

2. What should we be doing with googletest outside madgraph4gpu?
[...]

PR #657, opened 2023-05-23
added CommonRandomNumbers.h to code generation

zeniheisser:
Minor fix to the plugin, so that all necessary headers are exported at code generation

only missing part to be able to run straight out of the madgraph interface now is googletests


roiser:
Shall we remove the header in the "old" place in our github repo (/tools)?

valassi:
Hi @zeniheisser thanks for the patch!

There were (to my taste) a few missing issues, I have now included them in a followup MR #662, in particular
- I regenerated all processes SA and MAD in the repo with the patch in epochX/cudacpp
- I removed the dependency on the tools directory from cudacpp.mk (this is one of the main points of removing CommonRandomNumbers.h: the dependency is no longer needed, then it is better to remove it for clarity, also so that there is no confusion where the headers are taken from)
- As CommonRandomNumbers.h is now in the plugin, my script checked formatting, and so I reformatted it using clang-format (in addition, I also added the license and copyright)

valassi:
PS This is related to #613 (generated code should be relocatable)

PR #662, opened 2023-05-23
Complete the patch for CommonRandomNumbers.h

valassi:
Complete the patch for CommonRandomNumbers.h

This is a followup to MR #657, completing a few missing issues


valassi:
All tests passed, I am self merging

PS This is related to #613 (generated code should be relocatable)

PR #664, opened 2023-05-24
Proc gen

zeniheisser:


valassi:
As discussed, I rolled back the removal of tools/CommonRandomNumbers.h in #665, mainly because this is needed by sycl. This is actually also needed by epoch1 and epoch2, which we are using in the CI tests (in fact, the CI tests were failing here).

PR #666, opened 2023-05-24
WIP: make googletest relocatable

valassi:
WIP: make googletest relocatable


valassi:
The CI tests were initially failing as 'make install' gave a different result than on my node. I then learnt this is no longer supported. So I replaced it by a simpler version. The CI tests are now succeeding.

I will self merge

Issue #667, opened 2023-05-25
Number of particles with add_process


roiser:
I am creating a process

```
generate p p > t t~
add process p p > t t~ j
add process p p > t t~ j j
```

which produces code which would not compile. Potentially the problem stems from the fact that the different processes have various numbers of particles in the final state (2, 3 and 4)

I believe the main problem is the variable  ```const int nparf``` in ```src/mgOnGpuConfig.h``` which is set to the same number for all processes (2 in that case). 

The compilation error observed then shows up in "a" CPPProcess.cc e.g. for variable ```static constexpr short tHel[ncomb][mgOnGpu::npar]``` 

Please note that I believe also the ncomb actually is wrong. We should double check that this is also set correctly.

valassi:
Thanks Stefan I am starting to prepare a patch

valassi:
Ok the patch is essentially ready. I am running tests, will see tomorrow. This is in #668

valassi:
I have rerun tests, but there are still somepeculiar issues I want to have a quick look at
- Performance degrades slightly, with a higher number of GPU registers?!
- Tmad tests fail with cross section mismatches? (this may be an issue on the node I am using)

valassi:
About the two issues above:
- I fixed the tmad bug causing errors on itscrd90 (python vs python3)
- I opened #669 and understood the performance regression: the cause is the simultanenous move from int to size_t, which I will undo

valassi:
Closed by merging #668

Issue #644, opened 2023-04-26
Process-specific paramaters are stored in common /src/ directory


zeniheisser:
This is a generalisation of an issue previously discussed with @valassi and @roiser. The long and short of it is that some parameters which should be stored in the P-directories are stored in the src-directory. Here are two examples.

``
generate p p > t t~ g g
``

This process has two subprocesses, `g g > t t~ g g` and `u u~ > t t~ g g` (the latter one being the combined file for the four different quark cases). Both subprocesses compile, but `g g > t t~ g g` seg faults at runtime due to needing to store more internal wavefunctions in memory than are allocated. Shown in the warning raised at compile time:
``
CPPProcess.cc:1723:36: warning: array subscript 25 is above array bounds of ‘mgOnGpu::fptype* [17]’ {aka ‘double* [17]’} [-Warray-bounds]
 1723 |       VVV1P0_1<W_ACCESS, CD_ACCESS>( w_fp[0], w_fp[7], COUPs[0], 0., 0., w_fp[25] );
      |       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
CPPProcess.cc:160:13: note: while referencing ‘w_fp’
  160 |     fptype* w_fp[nwf];
      |             ^~~~
``

Another case is when running multiprocesses, e.g.
``
generate g g > t t~
add process g g > t t~ g
``
which will generate code with multiple subprocess directories, i.e. directories P1... and P2.... In this case, `g g > t t~ g` fails already at compile time, 
``
gCPPProcess.cu(639): error: too many initializer values

gCPPProcess.cu(640): error: too many initializer values

gCPPProcess.cu(641): error: too many initializer values

gCPPProcess.cu(642): error: too many initializer values

gCPPProcess.cu(643): error: too many initializer values

gCPPProcess.cu(644): error: too many initializer values

gCPPProcess.cu(645): error: too many initializer values

gCPPProcess.cu(646): error: too many initializer values

gCPPProcess.cu(647): error: too many initializer values

gCPPProcess.cu(648): error: too many initializer values

gCPPProcess.cu(649): error: too many initializer values

gCPPProcess.cu(650): error: too many initializer values

gCPPProcess.cu(651): error: too many initializer values

gCPPProcess.cu(652): error: too many initializer values

gCPPProcess.cu(653): error: too many initializer values

[...]

PR #668, opened 2023-05-25
move more process-dependent parameters (npar, ncomb...) from mgOnGpuConfig.h to CPPProcess.h

valassi:
This is a WIP MR for #667 submitted by @roiser (cc @oliviermattelaer)

I am moving more process-dependent parameters from mgOnGpuConfig.h to CPPProcess.h.

This is very similar to the previous nwf issue #644 submitted by @zeniheisser . As in that case, I have realtively easily (well, a bit more complex here) found a way to shape the generated code. But I will have trouble generating in from codegen.

PS As an example I am using a process with P1 and P2, doing both gg_tt and gg_ttg


valassi:
Ok this is essentially complete I think. Code generation was (for reasons that I do not understand) easier than for nwf. I checked that pp_tttt generates ok, just in case.

I am rerunning tests. Tomorrow if all ok this can be merged. It will close #667

valassi:
I have rerun tests, but there are still somepeculiar issues I want to have a quick look at
- Performance degrades slightly, with a higher number of GPU registers?!
- Tmad tests fail with cross section mismatches? (this may be an issue on the node I am using)

valassi:
About the two issues above:
- I fixed the tmad bug causing errors on itscrd90 (python vs python3)
- I opened #669 and understood the performance regression: the cause is the simultanenous move from int to size_t, which I will undo

valassi:
I fixed #669 by going back to int also in CODEGEN, and regenerated all processes

I am rerunning all tests - will only merge when these are ok (note I went back to Alma8 for consistency)

valassi:
All tests passed - this can now be soon merged

valassi:
CI tests passed - self merging

Issue #656, opened 2023-05-23
Port post-generation patches (patchMad.sh) upstream to mg5amcnlo or to the plugin


valassi:
Presently code generation in cudacpp goes through the patchMad.sh script, currently:
https://github.com/madgraph5/madgraph4gpu/blob/3a4850184f1de483bbab1bff5a6a322590692291/epochX/cudacpp/CODEGEN/patchMad.sh

This applies post-generation patches to the code generated using the python madgraph machinery. For a release, these patches must be instead moved upstream to mg5amcnlo (or to the plugin itself).

The main ingredients of this are the following (I may be forgetting a few):
- add a few files to fortran code such as timer.h ompnumthreads.cc (these may best go to the plugin?)
- apply patches patch.P1 and patch.common (these have already been reduced significantly by the vector.inc MRs https://github.com/mg5amcnlo/mg5amcnlo/pull/23, https://github.com/mg5amcnlo/mg5amcnlo/pull/24, https://github.com/mg5amcnlo/mg5amcnlo/pull/54: the rest should go partly to the plugin an dpartly maybe upstream, eg a general mechanism to patch driver.f must be implemented, maybe also for sycl)
- fix a bug in coupl_write.inc, probably this should go upstream (avoid writing n times the coupling if the couplking is a pointer not a value)
- also very important, generate coloramps.h from coloramps.inc

Assigning this to me, cc @roiser and @oliviermattelaer

valassi:
Note, this is part of point 2 in #576

valassi:
I have started a WIP patch for this in MR #675

valassi:
Note: the coloramps.h patch has now been removed from patchMad.sh (see #713, fixing #655)

valassi:
The strategy for integrating patchMad.sh patches upstream has changed:
- this is no longer about getting rid of patchMad.sh and moving the patches to the common LO templates (as in my WIP MR #675, which has now been closed)
- instead, we keep patchMad.sh but we move it to the plugin itself so that it can be called from standard MG5AMC commands (work with @roiser and @oliviermattelaer on MR #747)

This will be closed when we merge #747

PR #675, opened 2023-06-02
WIP: remove patchMad.sh

valassi:
This is a WIP MR for removing patchMad.sh #656 

The initial proof of concept so far is ok: I copied the modified Fortran makefile (which is the most heavily modified piece of code!) to the cudacpp templates, and I copy it directly from there, overriding the previosuly existing non-modified makefile

I must now extend this to all the rest

I also want to implement a sanity check that the original umodified makefile is what we started from. Otherwise one must patch the modified template (in exactly the same way as patch.common was done, but this must be done in the plugin templates, not in an external patchMad)


valassi:
I have just rebased after merging #676

valassi:
This is now superseded by MR #747 "Call the patchMad.sh script from within the plugin's finalize"

In other words, the strategy for integrating patchMad.sh patches upstream has changed:
- this is no longer about getting rid of patchMad.sh and moving the patches to the common LO templates (this MR #675)
- instead, we keep patchMad.sh but we move it to the plugin itself so that it can be called from standard MG5AMC commands

This WIP MR #675 is now obsolete and superseded and it can be closed

PR #620, opened 2023-03-31
WIP: towards a workflow

roiser:
This is a tentatively first set of patches towards running the whole workflow. So far this PR contains two sets of patches to 

- fixes for missing includes (symmetry.f) which were for now not compiled 
- create the input files for madevent with the additional parameters for the application execution 

more to be added in this PR ...

PR #676, opened 2023-06-03
Add gg_tt01g.mad to the repo, including Stefan's P1/P2 fixes

valassi:
Looking at @roiser 's MR #620, this includes also a few fixes for processes with P1 and P2 directories.

I am merging them here, and applying them to gg_tt01g.mad, which this time I am adding permanently to the repo.


valassi:
CI tests are ok, I am self merging

Note, there are now 7 processes sa and 7 processes mad in the repo

valassi:
Hi @roiser, thanks. Yes I agree at some point we should add and try the "real" processes of ATLAS and CMS. Here my point was just to have in the repo a simpler and more manageable provcess, which however has some of the complications of ATLAS/CMS. In particular non of pur previous gg_tt... gg_ttggg alone had separate P1 and P2 directories (which have their own nwf, npar etc... issues). So I added this for the moment and I will try to regenerate it each time (I should also add it to build/runtime tests....)

Issue #679, opened 2023-06-04
Remove C++ code dependencies on CUDA headers, except where curand is needed


valassi:
Part of the cleanup of builds. Remove C++ code dependencies on CUDA headers, except where curand is needed

Related to #674

valassi:
Closed by MR #681

Issue #674, opened 2023-06-02
madeventfortran vs madeventcpp vs madeventcuda


valassi:
Hi @roiser (and @oliviermattelaer) this is one part of #658, with WIP by Stefan in MR #620

There are rather different parts in the #658 "remove non standard features in madevent". What I want to discuss here is: **how does the launch machinery choose whether to build/run fortran, cpp or cuda MEs**.

Stefan has some good points in MR #620.

For instance
https://github.com/madgraph5/madgraph4gpu/pull/620/commits/237620ccbc7af5250b25e00d646e27ab41642709
```
diff --git a/epochX/cudacpp/gg_tt.mad/SubProcesses/makefile b/epochX/cudacpp/gg_tt.mad/SubProcesses/makefile
... 
+.PHONY: madeventfortran madeventcuda madeventcpp
+
+madeventfortran: $(PROG)
+
+madeventcpp: $(CUDACPP_BUILDDIR)/c$(PROG)_cudacpp
+       rm -f $(CUDACPP_BUILDDIR)/$(PROG)
+       cd $(CUDACPP_BUILDDIR); ln -s c$(PROG)_cudacpp madevent
+
+madeventcuda: $(CUDACPP_BUILDDIR)/c$(PROG)_cudacpp
+       rm -f $(CUDACPP_BUILDDIR)/$(PROG)
+       cd $(CUDACPP_BUILDDIR); ln -s g$(PROG)_cudacpp madevent
+
```
(Two suggestions above: fix in the lst dependency should be g; keep a copy of madevent fortran anyway?)

And then
https://github.com/madgraph5/madgraph4gpu/pull/620/commits/15b5974f403bd9affd858d7cad739268039c63b6
https://github.com/madgraph5/madgraph4gpu/pull/620/commits/eb1bcb1d45032952e8e1d336f9fecd282f8df36b
```
git diff upstream/master epochX/cudacpp/gg_tt.mad/bin/internal/madevent_interface.py
______________________________________________________________________________
git diff /data/avalassi/GPU2023/madgraph4gpuX/epochX/cudacpp/gg_tt.mad/bin/internal/madevent_interface.py
...
<                     self.compile(['madevent'], cwd=Pdir)
<                     
---
> 
>                     exec_mode = 0
>                     if 'exec_mode' in self.run_card:
>                         exec_mode = self.run_card['exec_mode']
> 
>                     if exec_mode == '0':
[...]

PR #681, opened 2023-06-04
improve cuda/c++ separation in cudacpp builds (reduce cpp dependency on nvtx/curand)

valassi:
This is a WIP MR to improve cuda/c++ separation in cudacpp builds. It addresses #679.

Part of the cleanup necessary to eventually build cleanly during launch, see #674


valassi:
This bit is complete and can essentially be merged. I encapsulated as much as possible the dependecies of the c++ builds on NVTX and on curand.

Eventually it would be better to have two cleanly separated cuda and c++ builds (eg also with runTest split in two?) but this will be in #674 and similar issues

valassi:
ci tests ok, self merging

PR #688, opened 2023-06-09
Towards a worflow (extending Stefan's work on runcards)

valassi:
Hi @roiser this is a WIP MR that starts from your #620, fixes some issues and extends runcard features. (And is rebased on the laest master)

Amongst other things it addresses parts of #687.

PR #692, opened 2023-06-12
Include Stefan's addition of atlas and cms to the generate script

valassi:
Hi @roiser this is a small MR that includes (and modifies, partly undoing) everything else in your MR #620 that is not going into #688. After this, your #620 can be considered merged (and I will close it).

Essentially here
- I keep your addition of atlas and cms processes (WITHOUT FIXING THEM: there are issues)
- I remove the ppttj* processes (I think we do not need them, and the syntax is wrong for one as add process should be added)
- I remove the top level gitignore (as I added them to each process in #688)


valassi:
I have closed WIP MR #620 because all its commits wre either in #688 or in this #692,

Issue #696, opened 2023-06-13
Add support for HRDCOD=1 builds of loop_no_b_mass SM processes like pp_ttW


valassi:
Add support for HRDCOD=1 builds of no_b_mass SM processes like pp_ttW

Unfortunately no_b_mass pp_ttW does not only fail to build HRDCOD=0 (#695), it also fails to build HRDCOD=1. This is how it fails
```
[avalassi@itscrd80 gcc11.2/cvmfs] /data/avalassi/GPU2023/madgraph4gpuX/epochX/cudacpp/pp_ttW.sa/SubProcesses/P0_Sigma_loop_sm_no_b_mass_dux_ttxwm> make HRDCOD=1
OMPFLAGS=-fopenmp
AVX=512y
FPTYPE=d
HELINL=0
HRDCOD=1
RNDGEN=hasCurand
Building in BUILDDIR=. for tag=512y_d_inl0_hrd1_hasCurand (USEBUILDDIR is not set)
make -C ../../src  -f cudacpp_src.mk
make[1]: Entering directory '/data/avalassi/GPU2023/madgraph4gpuX/epochX/cudacpp/pp_ttW.sa/src'
AVX=512y
ccache /cvmfs/sft.cern.ch/lcg/releases/gcc/11.2.0-ad950/x86_64-centos8/bin/g++  -O3  -std=c++17 -I.  -fPIC -Wall -Wshadow -Wextra -ffast-math  -fopenmp -march=skylake-avx512 -mprefer-vector-width=256  -DMGONGPU_FPTYPE_DOUBLE -DMGONGPU_FPTYPE2_DOUBLE -DMGONGPU_HARDCODE_PARAM -c Parameters_loop_sm_no_b_mass.cc -o Parameters_loop_sm_no_b_mass.o
In file included from Parameters_loop_sm_no_b_mass.cc:15:
Parameters_loop_sm_no_b_mass.h: In function ‘const Parameters_loop_sm_no_b_mass_dependentCouplings::DependentCouplings_sv Parameters_loop_sm_no_b_mass_dependentCouplings::computeDependentCouplings_fromG(const fptype_sv&)’:
Parameters_loop_sm_no_b_mass.h:291:46: error: ‘COND’ was not declared in this scope
[...]

Issue #695, opened 2023-06-13
Add support for HRDCOD=0 builds of loop_no_b_mass SM processes


valassi:
Add support for HRDCOD=0 builds of no_b_mass SM processes

Presently these processes fail the build like non-SM processes. The problem, as usual, is the handling of parameter dependencies.

For instance in ttW (which I just added as suggested by Francesco Giuli in ATLAS):
```
[avalassi@itscrd80 gcc11.2/cvmfs] /data/avalassi/GPU2023/madgraph4gpuX/epochX/cudacpp/pp_ttW.sa/SubProcesses/P0_Sigma_loop_sm_no_b_mass_dux_ttxwm> make 
...
Building in BUILDDIR=. for tag=512y_d_inl0_hrd0_hasCurand (USEBUILDDIR is not set)
make -C ../../src  -f cudacpp_src.mk
make[1]: Entering directory '/data/avalassi/GPU2023/madgraph4gpuX/epochX/cudacpp/pp_ttW.sa/src'
AVX=512y
ccache /cvmfs/sft.cern.ch/lcg/releases/gcc/11.2.0-ad950/x86_64-centos8/bin/g++  -O3  -std=c++17 -I.  -fPIC -Wall -Wshadow -Wextra -ffast-math  -fopenmp -march=skylake-avx512 -mprefer-vector-width=256  -DMGONGPU_FPTYPE_DOUBLE -DMGONGPU_FPTYPE2_DOUBLE -c Parameters_loop_sm_no_b_mass.cc -o Parameters_loop_sm_no_b_mass.o
In file included from Parameters_loop_sm_no_b_mass.cc:15:
Parameters_loop_sm_no_b_mass.h:26:2: error: #error This non-SM physics process only supports MGONGPU_HARDCODE_PARAM builds (#439): please run "make HRDCOD=1"
   26 | #error This non-SM physics process only supports MGONGPU_HARDCODE_PARAM builds (#439): please run "make HRDCOD=1"
      |  ^~~~~
```

valassi:
The process is 
```
    pp_ttW)
      cmd="import model loop_sm-no_b_mass
      define p = p b b~
      define j = p
      define w = w+ w- # W case only
      generate p p > t t~ w @0
      add process p p > t t~ w j @1"
      ;;
```

valassi:
Note, this affects ALL no_b_mass SM processes - including other ATLAS processes like #673

valassi:
[...]

PR #697, opened 2023-06-13
Add ttV to code generation

valassi:
Add ttV to code generation (thanks to Francesco Giuli in ATLAS for the suggestion!)

Code generation succeeds (here also inclusing a minor formatting fix), but builds fail #695 and #696


valassi:
CI passed. Self merging

Issue #630, opened 2023-04-05
xsec from fortran and cpp differ in gg_uu tmad tests (bug in getGoodHel implementation)


valassi:
This is a followup to Olivier's "split_nonidentical_grouping" patch that I included in #619, and to the upcoming #626 about the addition of gq_ttq.

After Olivier's patch, gq_ttq now builds successfully (with two separate P1 directories, each with one DSIG1). The check.exe and gcheck.exe also complete successfully. Previously the build was failing (there was a single P1 directory, with DSIG1 and DSIG2 in Fortran, but the cudacpp presently only handles a single process.

However, in the tests of cmadevent/gmadevent, the cross section from fortran and cpp differ in gqttq tmad tests.

I open this as an issue to fix later on, but I will go on with merging #626 as-is.

Details here https://github.com/madgraph5/madgraph4gpu/pull/626/commits/2a8b34e61f91c8ff8efadd43529337bdca1a6fba

cc: @oliviermattelaer @roiser @zeniheisser 

I guess this might well be a blocker for SM pp to tt+jets? If anyone wants to help, that's welcome :-)
I will focus on other things like the vecsize rewriting...

oliviermattelaer:
So I'm checking the following code

```
#generate p p > t t~
define p = p / g
add process p g > t t~ j
output PROC_sm_3 --vector_size=XXXX
launch
set ickkw 1
set mc_grouped_subproc F
```
(so pure fortran code, with "old" grouping)

with 
- XXXX = 1 : 93.82 (no vectorization)
- XXXX = 2 : 94.37 
- XXXX = 8 : 95.12
- XXXX = 100: 95.39

oliviermattelaer:
one issue found is ipsel but does not seems to solve the real issue.

oliviermattelaer:
So more investigation but no real progress here.
One channel identify with the small change is 
G2 
correct cross-section is 0.2730e+2
and vectorised code returns 0.2764 e+2

The difference is clearly in the rewgt function. 
I thought it was link to the q2bck but it does not seem to be the case.

Need to check more what's going on, so far it just seems has if the scale returned are not the same.
(need to find a way to check that)

oliviermattelaer:
Ok I found the issue (pff that was sooo long for me to identify it)
The issue is that the renormalization scale used for the black box was not set per event but by batch
```
 for ivec=           1 asref=  0.11453761077792829
 for ivec=           2 asref=  0.11453761077792829
[...]

PR #705, opened 2023-06-14
Fix cudacpp getGoodHel to define a good helicity as ME>0 (fix mismatch in cross sections for fortran and cpp in gg_uu)

valassi:
WIP: add gu_ttu (with different cross sections for fortran and cpp)

This is a MR to investigate #630. The issue does not only affect gq_ttq where q is anyone of 8 quarks/antiquarks, it also affects simply the u quark.


valassi:
This MR is now ready for merging.

The MR contains a lot of investigations about comparisons of fortran and cpp cross sections in tmad tests for several processes, including gq_ttq but especially focusing on gg_uu. Many of these investigations will result in opening new issues shortly.

Of the main issues that was identified, specifically for gguu, following up on work by @zeniheisser, is that the old getGoodHel implementation was wrong. The goal was to define a good helicity as an helicity provided a non-zero ME contribution, but the implementation was unable to detect some helicities contributing a relative contribution lower than 1E-16 (or whatever the FP precision is). Therefore gguu in cudacpp was using 6 instead of 8 good helicities, resulting in different cross sections from fortran. This is now fixed and the contribution is compared to 0 (eventually, maybe the same LIMHEL will be used in fortran and cudacpp).

The main benefit of this MR is therefore the fix for getgoodhel (this closes #630). In the investigation I had added gguu.mad to the repo, but I am now removing it to avoid having too many processes there.

Note that gq_ttq cross sections are still wrong, this is now moved to a new issue #748.

valassi:
The CI tests have passed, I am self merging this one.

Issue #683, opened 2023-06-05
Prepare test scripts for "launch" (and, prior to that, document "launch" internals)


valassi:
Prepare test scripts for "launch" (and, prior to that, document "launch" internals)

This is a preliminary step to testing "launch" in #659. First we should prepare test scripts for that (and even before, personally I need to understand - reverse engineer - which files are called in which order).

Also related to Stefan's WIP MR #620

valassi:
Note, I have extended/replace MR #620 with WIP #688, I will try to start from there

valassi:
Advancing in steps. If in ggttmad (as of MR #688) I try generate events, I get a crash
```
[avalassi@itscrd80 gcc11.2/cvmfs] /data/avalassi/GPU2023/madgraph4gpuX/epochX/cudacpp/gg_tt.mad> ./bin/generate_events 
...
Error when reading /data/avalassi/GPU2023/madgraph4gpuX/epochX/cudacpp/gg_tt.mad/SubProcesses/P1_gg_ttx/G1/results.dat
Command "generate_events " interrupted with error:
Exception : Reported error: End code 134.0 
         Full associated log: 
        (ompnumthreadsNotSetMeansOneThread) DEBUG: OMP_NUM_THREADS is not set: will use only 1 thread
        (ompnumthreadsNotSetMeansOneThread) omp_get_max_threads() = 1
        ERROR! Card file '../../Cards/param_card.dat' does not exist and environment variable 'MG5AMC_CARD_PATH' is not set
        terminate called after throwing an instance of 'char const*'
```

valassi:
Try this
```
cd ..; rm -rf gg_tt.mad; git checkout gg_tt.mad; cd gg_tt.mad; git status
MG5AMC_CARD_PATH=$(pwd)/Cards ./bin/generate_events 
```
This is better, but
```
Command "generate_events " interrupted with error:
FileNotFoundError : [Errno 2] No such file or directory: '/data/avalassi/GPU2023/madgraph4gpuX/epochX/cudacpp/gg_tt.mad/HTML/run_01'
Please report this bug on https://bugs.launchpad.net/mg5amcnlo
More information is found in '/data/avalassi/GPU2023/madgraph4gpuX/epochX/cudacpp/gg_tt.mad/run_01_tag_1_debug.log'.
Please attach this file to your report.
```

valassi:
I have added HTML/.keepme. Now the same commands above succeed! But I have no idea what they do.

[...]

Issue #655, opened 2023-05-19
Problems in ggttgg lhe files after the latest upstream changes (replacing channel by iconfig in select_color)


valassi:
Hi @oliviermattelaer (cc @roiser), I have rerun the usual tests after the upgrade to the latest mg5amcnlo upstream.

Most tests succeed, but in ggttgg I get differences in the lhe files:
```
In tmad/logs_ggttgg_mad/log_ggttgg_mad_d_inl0_hrd0.txt:
+ERROR! events.lhe.cpp.1 and events.lhe.ref.1 differ!
+diff /data/avalassi/GPU2023/madgraph4gpuX/epochX/cudacpp/gg_ttgg.mad/SubProcesses/P1_gg_ttxgg/events.lhe.cpp.1 /data/avalassi/GPU2023/madgraph4gpuX/epo$
+3,4c3,4
+<          21   -1    0    0  503  502  0.00000000000E+00  0.00000000000E+00  0.13289043826E+04  0.13289043826E+04  0.00000000000E+00 0. -1.
+<          21   -1    0    0  504  503 -0.00000000000E+00 -0.00000000000E+00 -0.81223316322E+02  0.81223316322E+02  0.00000000000E+00 0. -1.
+---
+>          21   -1    0    0  505  502  0.00000000000E+00  0.00000000000E+00  0.13289043826E+04  0.13289043826E+04  0.00000000000E+00 0. -1.
+>          21   -1    0    0  502  503 -0.00000000000E+00 -0.00000000000E+00 -0.81223316322E+02  0.81223316322E+02  0.00000000000E+00 0. -1.
+6,8c6,8
+<          -6    1    1    2    0  505  0.39403209480E+02 -0.10079469096E+02  0.28578226692E+03  0.33653337532E+03  0.17300000000E+03 0. -1.
+<          21    1    1    2  504  501 -0.19269775075E+03  0.33434234480E+02  0.26595208036E+03  0.33012237159E+03  0.00000000000E+00 0. -1.
+<          21    1    1    2  505  502  0.20498361398E+02  0.29398294961E+02  0.12436578484E+03  0.12942677855E+03  0.00000000000E+00 0. -1.
+---
+>          -6    1    1    2    0  504  0.39403209480E+02 -0.10079469096E+02  0.28578226692E+03  0.33653337532E+03  0.17300000000E+03 0. -1.
[...]

PR #709, opened 2023-06-16
first systematic 'launch'-like tests (and move to the latest select_color upstream)

valassi:
This is a WIP MR with some first systematic 'launch' tests (./bin/generate_events actually) using the lauX.sh script in #683.

There are several issues (icolamp bugs) and a lot to be analysed and tuned - I will open separate issues for those.


valassi:
In the latest commits I have rerun the testst after fixing #710 (using @oliviermattelaer select_color patch ... which however reintroduces #655 that will need to be fixed).

I will now work on tuning the time it takes (the tests are very long) #711

PR #714, opened 2023-06-20
A few minor additions to doc and scripts for gridpacks

valassi:
A few minor additions to doc and scripts for gridpacks


valassi:
Self merging

Issue #721, opened 2023-07-13
code generation patches succeed on itscrd80 and fail on itscrd90?


valassi:
This is a strange issue that affects my environment: code generation patches succeed on itscrd80 and fail on itscrd90?

```
[avalassi@itscrd90 gcc11/usr] /data/avalassi/GPU2023/madgraph4gpuX/epochX/cudacpp> ./CODEGEN/generateAndCompare.sh --mad gg_ttgg
...
INFO: load configuration from /data/avalassi/GPU2023/madgraph4gpuX/epochX/cudacpp/gg_ttgg.mad/Cards/me5_configuration.txt  
INFO: load configuration from /data/avalassi/GPU2023/madgraph4gpuX/epochX/cudacpp/gg_ttgg.mad/Cards/me5_configuration.txt  
Using default text editor "vi". Set another one in ./input/mg5_configuration.txt
Using default eps viewer "evince". Set another one in ./input/mg5_configuration.txt
No valid web browser found. Please set in ./input/mg5_configuration.txt
treatcards param
quit
INFO:  
DEBUG: cd /data/avalassi/GPU2023/madgraph4gpuX/epochX/cudacpp/gg_ttgg.mad; patch -p4 -i /data/avalassi/GPU2023/madgraph4gpuX/epochX/cudacpp/CODEGEN/MG5aMC_patches/PROD/patch.common
patching file Source/genps.inc
patching file Source/makefile
patching file Source/vector.inc
patching file SubProcesses/makefile
patching file Source/make_opts
Hunk #1 FAILED at 1.
1 out of 1 hunk FAILED -- saving rejects to file Source/make_opts.rej
patching file bin/internal/banner.py
patching file bin/internal/gen_ximprove.py
patching file bin/internal/madevent_interface.py
DEBUG: cd /data/avalassi/GPU2023/madgraph4gpuX/epochX/cudacpp/gg_ttgg.mad/SubProcesses/P1_gg_ttxgg; patch -p6 -i /data/avalassi/GPU2023/madgraph4gpuX/epochX/cudacpp/CODEGEN/MG5aMC_patches/PROD/patch.P1
patching file auto_dsig1.f
patching file driver.f
Hunk #1 succeeded at 74 with fuzz 1.
patching file matrix1.f
Hunk #2 succeeded at 191 (offset 48 lines).
Hunk #3 succeeded at 269 (offset 48 lines).
Hunk #4 succeeded at 297 (offset 48 lines).
[...]

PR #722, opened 2023-07-13
Fix code generation on itscrd80 (f2py3 instead of f2py3.9)

valassi:
This is a simple MR closing #721


valassi:
The ci succceeed, I am self merging

cc: @Jooorgen

Issue #701, opened 2023-06-13
Floating point exceptions in CPP launch of pp_ttW  (FPE in vxxxxx function in SIMD mode)


valassi:
Floating point exceptions in CPP launch of pp_ttW 

This is in my branch nobm
```
[avalassi@itscrd80 gcc11.2/cvmfs] /data/avalassi/GPU2023/madgraph4gpuX/epochX/cudacpp> HRDCOD=1 tlau/lauX.sh -CPP nobm_pp_ttW.mad/
Execute lauX.sh for process nobm_pp_ttW.mad/ in directory /data/avalassi/GPU2023/madgraph4gpuX/epochX/cudacpp
...
Generating 10000 events with run name run_01
survey  run_01 
INFO: compile directory 
compile Source Directory
Using random number seed offset = 21
INFO: Running Survey 
Creating Jobs
Working on SubProcesses
INFO:     P1_gu_ttxwpd  
INFO: Building madevent in madevent_interface.py with 'CPP' matrix elements 
INFO:     P1_gd_ttxwmu  
Note: The following floating-point exceptions are signalling: IEEE_INVALID_FLAG IEEE_DIVIDE_BY_ZERO IEEE_UNDERFLOW_FLAG IEEE_DENORMAL
Note: The following floating-point exceptions are signalling: IEEE_INVALID_FLAG IEEE_DIVIDE_BY_ZERO IEEE_UNDERFLOW_FLAG IEEE_DENORMAL
...
```

hageboeck:
I investigated this a little for
`epochX/cudacpp/nobm_pp_ttW.mad/SubProcesses/P1_gd_ttxwmu`

The first floating-point exception is a devision by zero here:
```c++
      const fptype_v pzpt = pvec3 / ( pp * pt ) * sqh * hel;
```
Caused by `pt = {0, 0, 0, 0}`.

It's a different process, but here's the relevant code:
https://github.com/madgraph5/madgraph4gpu/blob/603ff90199e1e169e0b45a2e48d971e5ac46dac4/epochX/cudacpp/gg_ttgg.mad/src/HelAmps_sm.h#L496-L529

The fp exception is in line 516, caused by line 497, where
```
const fptype_sv pt = fpsqrt( ( pvec1 * pvec1 ) + ( pvec2 * pvec2 ) );
pvec1 = @0x7fffffbd98e0: {0, 0, 0, 0}
[...]

PR #706, opened 2023-06-14
fix FPEs and debug nobm_pp_ttW for ATLAS

valassi:
WIP: add sm-no_b_mass processes for ATLAS

These are affected by #695 #696 #701

(PS 25 Nov 2023: the first two issues do not apply to nobm_pp_ttW, only to loop-nobm models; the third issue has been fixed)

Issue #414, opened 2022-04-02
Clean up interaction of Subprocess and src Makefile (export, override etc)


valassi:
Low priority reminder.

Currently the cudacpp plugin drives builds from the Subprocess/P1 directory (by the way, what will happen with pp, with several subprocesses?). A 'make' in Subprocess/P1 triggers a 'make' in src too. The 'override' command is used extensively in both Makefiles (needed? I added this to be insensitive to extetnal env variables I think). Some default values if variables are not set (eg for AVX) are set in Subprocess Makefile, then 'export' is used to propagate them to src too.

This normally works fine, but I realised that for instance 'make AVX=' fails, because AVX is then set with override in Subprocess, 'export AVX' is called too, but then actually src builds still using 'AVX=' from the command line. This is because of MAKEOVERRIDES, see https://www.gnu.org/software/make/manual/html_node/Options_002fRecursion.html. In practice:
```
[avalassi@itscrd70 gcc10.2/cvmfs] /data/avalassi/GPU2020/madgraph4gpuTer/epochX/cudacpp/gg_tt/SubProcesses/P1_Sigma_sm_gg_ttx> make AVX=
OMPFLAGS=
AVX=512y
FPTYPE=d
HELINL=0
HRDCOD=0
RNDGEN=hasCurand
Building in BUILDDIR=. for tag=512y_d_inl0_hrd0_hasCurand (USEBUILDDIR is not set)
make -C ../../src 
AVX=
make[1]: Entering directory `/data/avalassi/GPU2020/madgraph4gpuTer/epochX/cudacpp/gg_tt/src'
Makefile:93: *** Unknown AVX='': only 'none', 'sse4', 'avx2', '512y' and '512z' are supported.  Stop.
make[1]: Leaving directory `/data/avalassi/GPU2020/madgraph4gpuTer/epochX/cudacpp/gg_tt/src'
make: *** [../../lib/libmg5amc_common.so] Error 2
```

One workaround is to reset MAKEOVERRIDES. I have done this in a WIP PR, but it may have bad side effects
https://github.com/madgraph5/madgraph4gpu/pull/401/commits/5a223bf7f1b402e2065fed509755db7d5d9d2569
Another option might be to use 'make -e' but this is even worse as IIUC it would take precedence over all overrides.

I guess that for the moment I will just stay as it was (without resetting MAKEOVERRIDES). It is not a problem if 'make AVX=' fails, this is not meant to be used anyway.

[...]

PR #723, opened 2023-07-17
Fixes in xxxxx for IEEE_DIVIDE_BY_ZERO FPE; separate cpu/gpu namespaces and fix runtest segfault

valassi:
This is WIP MR with comprehensive fixes in xxxxxx funxtions for FPE floating point exceptions

It is motivated by and is meant to fix bug #701


valassi:
I have merged into this MR #723 the contents of my namespace branch of WIP #728, which I will close.  That was a MR fixing https://github.com/madgraph5/madgraph4gpu/issues/725, ie the segfault in runtest: the fix is now implemented here.

Note: this patch essentially extends the use of the two separate namespaces for CPU and GPU builds (as discussed in https://github.com/madgraph5/madgraph4gpu/issues/318). While I am still not convinced that it is a good idea to mix both types of builds in a single executable (which is what was causing the segfault https://github.com/madgraph5/madgraph4gpu/issues/725), at least now this is a bit safer. I am not sure if all issues are fixed, but ceratnly it is much better. A few details
- Now fptype_sv and cxtype_sv, and also cxtype, are defined in either a CPU or GPU namespace, where they have different definitions
- In https://github.com/madgraph5/madgraph4gpu/issues/725 I was reporting not only a segfault, but also a strange mixup where some tests seemed to mix data arrays with different strutures, which is indeed probably a side effect of having different vector sizes mixed
- I have no idea why we have not observed these errors before, and why they only show up in debug builds...
- I have also moved to separate GPU/CPU namespaces the Parameters class. For this I had to also build the Parameters class in CUDA in src.
- The CUDA build in src is done by simply exporting NVCC. This may be a good ides to simplify makefiles eventually in src also for AVX etc.

valassi:
This is finally complete - as good as it gets - and passes the CI tests. I am self merging.

I will then document it a posteriori.

valassi:
This is some documentation for this MR #723.

This is addressing two rather large/complex issues. It is fixing 8 github issues in total.

1) Part 1 is floating point exceptions FPEs (branch fpe)
- It was meant to address all four FPEs in #701 for nobm_pp_ttW. However, in the end it only fixed one of the four, the IEEE_DIVIDE_BY_ZERO in SIMD code (essentially, using a fake denominator instead of 0 in SIMD vectors, and using the `volatile` keyword to outsmart the optimizer). There are three other FPEs pending, to be followed up in #733 and in WIP MR #706. I have closed #701 as IEEE_DIVIDE_BY_ZERO is fixed.
- The FPEs appear when C++ is linked to Fortran. To reproduce FPEs in C++ code only, I have added feenableExcept in testxxx, and I have maded other enhancements in the tests.
- One early attempt to fix IEEE_DIVIDE_BY_ZERO in SIMD code was disabling auto-vectorization for the ixx/oxx/vxx functions. However this resulted in poor performance for these functions. I have fixed this using the `volatile` keyword instead (after another attempt which was failing on clang, see below). This closes #727: now the performance is ok.
- Most of the debugging for IEEE_DIVIDE_BY_ZERO in SIMD code was done using gcc. My early attempt to disable auto vectorization was compiler dependent, so I made a note of testing clang too. It turns out that my second attempt, to recover performance, using fake denominators>0, was ok on gcc but failing on clang. This I fixed using `volatile`. This closes #724: all tests pass on gcc, clang and icx, with good performance.
- En passant while working on clang and icx, I fixed a minor build warning that was pending since a long time. This closes #516.
- The feenableExcept that I used to debug FPEs on linux is missing on MacOS, causing build errors. I have therefore removed feenableExcept from MacOS (FPEs are not tested in c++ code on MacOS). This closes #730.

2) Part 2 is about separating cpu/gpu namespaces and fixing debug builds (branch namespace, initially in MR #728 which I then coalesced with this one)
- The issue is that, to debug FPE issues, I started using debug builds. (Actually these were not always ideal, as some subtle issues only came from -O3 optimized builds anyway...). 
- At some point I realised that the debug build of runTest was crashing. I understood that this was because runTest was mixing functions that were defined in different ways for CPU and GPU, or with different global settings. The solution (in line with #318 about heterogeneous applications) was to improve the separation of the two namespaces for CPU and GPU entities. This fixes the runTest crash and closes #725.
- En passant, I identified and fixed a minor issue in the way compiler flags are passed to CUDA for debug builds (-Xcompiler was missing). This closes #729.
- Separating CPU and GPU namespaces eventually implied that Parameters.cc must be built twice, once for the CPU and once for the GPU. This means that I had to reintroduce CUDA builds in src (these had disappeared, we were only having SubProcesses builds for CUDA). I implemented this using `export NVCC` (which is related to makefile cleanup in #362 and #414... it is probably a good idea to use `export` much more). In doing this, however, I introduced a bug as I forgot to also export CUFLAGS. This was then fixed, closing #731.

I think this should be more or less all for this MR.

Next steps on this line of work would be
- merge the new master into Jorgen's HIP branch, as it will be affected quite significantly
- continue the investigation of FPEs in #733 (the ATLAS process nobm_pp_ttW is unusable)

cc @roiser @oliviermattelaer @hageboeck @zeniheisser @Jooorgen

Issue #736, opened 2023-07-25
Floating point exceptions in testxxx using icx2023.2 (FPEs from sqrt in ixxx/oxxx/vxxx)


valassi:
Floating point exceptions in testxxx using icx2023.2
```
[avalassi@itscrd90 icx2023/cvmfs] /data/avalassi/GPU2023/madgraph4gpuX/epochX/cudacpp/gg_tt.mad/SubProcesses/P1_gg_ttx> make cleanall; CUDA_HOME=none make -f cudacpp.mk -j FPTYPE=f AVX=none 
...
[avalassi@itscrd90 icx2023/cvmfs] /data/avalassi/GPU2023/madgraph4gpuX/epochX/cudacpp/gg_tt.mad/SubProcesses/P1_gg_ttx> ./runTest.exe 
Running main() from /data/avalassi/GPU2023/madgraph4gpuX/test/googletest/googletest/src/gtest_main.cc
[==========] Running 3 tests from 3 test suites.
[----------] Global test environment set-up.
[----------] 1 test from SIGMA_SM_GG_TTX_CPU_XXX
[ RUN      ] SIGMA_SM_GG_TTX_CPU_XXX.testxxx
Floating Point Exception (CPU neppV=1): 'unknown' ievt=-1
```

valassi:
It is clear from debugging and fixing more and more that these are FPEs from sqrts in ixxx/oxxx/vxxx.

So I would say these are one part of #733... even if there the FPEs that I debugged (with gcc) were in VVV functions.

The general rule seems to be:
- whenever there is a division and SIMD, use a volatile
- whenever there is a sqrt with SIMD, use a volatile

Issue #735, opened 2023-07-25
RAMBO is unstable in single precision


valassi:
RAMBO is unstable in single precision

I just realised that for gg_tt in single precision the compareTest test gives deviations larger than the 3% tolerance for some specific builds (debug mode with gcc)
```
[avalassi@itscrd90 gcc11/usr] /data/avalassi/GPU2023/madgraph4gpuX/epochX/cudacpp/gg_tt.mad/SubProcesses/P1_gg_ttx> make cleanall; make -f cudacpp.mk -j debug FPTYPE=f 
...
[avalassi@itscrd90 gcc11/usr] /data/avalassi/GPU2023/madgraph4gpuX/epochX/cudacpp/gg_tt.mad/SubProcesses/P1_gg_ttx> ./runTest.exe 
...
[----------] 1 test from SIGMA_SM_GG_TTX_GPU/MadgraphTest
[ RUN      ] SIGMA_SM_GG_TTX_GPU/MadgraphTest.CompareMomentaAndME/0
INFO: Opening reference file ../../test/ref/dump_CPUTest.Sigma_sm_gg_ttx.txt
MadgraphTest.h:298: Failure
Value of: momentumErrors.str().empty()
  Actual: false
Expected: true

particle 2      component 3
         madGraph:   1.079908162355423e-01
         reference:  1.041966527700424e-01
         rel delta:  3.641348704695702e-02 exceeds tolerance of 2.999999932944775e-02
Google Test trace:
MadgraphTest.h:280: In comparing event 46 from iteration 0
   0  7.500000000000000e+02  0.000000000000000e+00  0.000000000000000e+00  7.500000000000000e+02
ref0  7.500000000000000e+02  0.000000000000000e+00  0.000000000000000e+00  7.500000000000000e+02

   1  7.500000000000000e+02  0.000000000000000e+00  0.000000000000000e+00 -7.500000000000000e+02
ref1  7.500000000000000e+02  0.000000000000000e+00  0.000000000000000e+00 -7.500000000000000e+02

   2  7.500137329101562e+02 -7.050960083007812e+02  2.555634918212891e+02  1.079908162355423e-01
ref2  7.500000000000000e+02 -7.051112670898438e+02  2.555740966796875e+02  1.041966527700424e-01

   3  7.499924926757812e+02  7.051048583984375e+02 -2.555704498291016e+02 -1.049910709261894e-01
[...]

Issue #734, opened 2023-07-24
Test builds with newer compilers (gcc13, clang16, icpx2023.2...)


valassi:
Test builds with newer compilers (gcc13, clang16, icpx2023.2...)

Just a reminder, as some of these have recently been advertised.

valassi:
Note, for the cern cvmfs installations

- icpx2023.2 seems to be working ok (it is based internally on clang17)

- clang16 with cuda120 is giving this build error (NB it is unsupported there)
```
ccache /usr/local/cuda-12.0/bin/nvcc   -Xcompiler -O3 -lineinfo -I. -I../../src -I/usr/local/cuda-12.0/include/ -DUSE_NVTX -gencode arch=compute_70,code=compute_70 -gencode arch=compute_70,code=sm_70 -use_fast_math -std=c++17  -ccbin /data/avalassi/GPU2023/madgraph4gpuX/tools/compilers/mg-lcg-clang++-16.0.3 -allow-unsupported-compiler -DMGONGPU_FPTYPE_DOUBLE -DMGONGPU_FPTYPE2_DOUBLE -Xcompiler -fPIC -c -x cu Parameters_sm.cc -o Parameters_sm_cu.o
/cvmfs/sft.cern.ch/lcg/releases/gcc/13.1.0-b3d18/x86_64-el9/lib/gcc/x86_64-pc-linux-gnu/13.1.0/../../../../include/c++/13.1.0/type_traits(1543): error: identifier "__remove_cv" is undefined
```

- clang16 with CUDA_HOME=none gives this build error
```
ccache /data/avalassi/GPU2023/madgraph4gpuX/tools/compilers/mg-lcg-clang++-16.0.3 -o fcheck.exe ./fcheck_sa.o -fopenmp ./fsampler.o -L../../lib -lmg5amc_common -Wl,-rpath,'$ORIGIN/../../lib'  -lgfortran -L../../lib -lmg5amc_gg_ttx_cpp ./CommonRandomNumberKernel.o ./RamboSamplingKernels.o
/cvmfs/sft.cern.ch/lcg/releases/binutils/2.40-acaab/x86_64-el9/bin/ld: ./fcheck_sa.o: relocation R_X86_64_32 against `.rodata' can not be used when making a PIE object; recompile with -fPIE
/cvmfs/sft.cern.ch/lcg/releases/binutils/2.40-acaab/x86_64-el9/bin/ld: failed to set dynamic section sizes: bad value
[...]

Issue #733, opened 2023-07-21
Three floating point exceptions in CPP launch of nobm_pp_ttW (FPE in COUP values in VVV1P0_1)


valassi:
THREE floating point exceptions in CPP launch of nobm_pp_ttW:
```
  IEEE_INVALID_FLAG IEEE_UNDERFLOW_FLAG IEEE_DENORMAL
```

This is a followup of #701, where I had found that there were FOUR floating point exceptions in CPP launch of pp_ttW:
```
Note: The following floating-point exceptions are signalling: IEEE_INVALID_FLAG IEEE_DIVIDE_BY_ZERO IEEE_UNDERFLOW_FLAG IEEE_DENORMAL
```

In MR #723 I am merging a comprehensive set of fixes, which I hoped would fix #701 completely. Unfortunately, they ONLY fix the division by zero, i.e. one of the four FPEs. The other three are still pending...

valassi:
I enabled FPEs in check.exe and this was enough to detect the FPEs.

They happen in VVV1P0_1 - which maybe only exists in pp_ttW and is why I had not seen this before?

```
[avalassi@itscrd80 gcc11.2/cvmfs] /data/avalassi/GPU2023/madgraph4gpuX/epochX/cudacpp/nobm_pp_ttW.mad/SubProcesses/P1_gd_ttxwmu> HRDCOD=1 make -j
...
[avalassi@itscrd80 gcc11.2/cvmfs] /data/avalassi/GPU2023/madgraph4gpuX/epochX/cudacpp/nobm_pp_ttW.mad/SubProcesses/P1_gd_ttxwmu> gdb --args ./check.exe -p 1 8 1
...
(gdb) run
Starting program: /data/avalassi/GPU2023/madgraph4gpuX/epochX/cudacpp/nobm_pp_ttW.mad/SubProcesses/P1_gd_ttxwmu/check.exe -p 1 8 1
warning: File "/cvmfs/sft.cern.ch/lcg/releases/gcc/11.2.0-ad950/x86_64-centos8/lib64/libstdc++.so.6.0.29-gdb.py" auto-loading has been declined by your `auto-load safe-path' set to "$debugdir:$datadir/auto-load".
To enable execution of this file add
        add-auto-load-safe-path /cvmfs/sft.cern.ch/lcg/releases/gcc/11.2.0-ad950/x86_64-centos8/lib64/libstdc++.so.6.0.29-gdb.py
[...]

PR #737, opened 2023-07-25
Several fixes for icx2023.2 (including fixes for sqrt FPEs in ixx/oxx/vxx)

valassi:
This MR includes a few minor fixes for new platforms (#734), in particular icx2023.2 and clang16
- increase tolerance for rambo momenta in comparison tests #735 (this also affects gcc)
- workaround for FPE in testxxx in icx 2023.2 #736
- add wrappers for clang16 compilation


valassi:
Renaming this as "Several fixes for icx2023.2 (including fixes for sqrt FPEs in ixx/oxx/vxx)"

It is clear that this is related to #733. These are new, differnt FPEs, which are not division by zero but rather sqrt of negative numbers...

valassi:
This has become another very complex FPE fixing campaign... and not only in ixx/oxx/vxx. The icx optimizer, based now on clang17, is doing even more strange stuff. In some cases I have data which is supposed to be exactly 0 and I am unable to take a sqrt of that, so I do it ONLY if it is GREATER than 0.

Rerunning all tests tonight. Hopefully it looks better.

valassi:
I have rerun all tests overnight - now all FPEs are fixed in icpx.

I have also made minor changes in the gtest handling of compiler-specific build directories (#125 and #738) for allwoing backward compatibility to epoch1/epoch2, which were previously failing the CI

This is now ready, I will self-merge. I will document the full contents a posteriori

valassi:
I have merged this MR #737 with another set of comprehensive FPE fixes. Here's some documentation
- This started off as a very minor port of the code to the latest icx2023.2 from cvmfs (see #734). I added a few wrapper scripts also for clang16.
- I increased the tolerance for RAMBO momenta comparisons from 3E-2 to 4E-2 in single precision. This closes #735.
- What took me again a long time was, again, new FPEs. In MR #723 (related to #701 #724 #727) I had implemented many fixes for floating point exceptions in SIMD divisions by zero, in some cases using the volatile keyword, which had been triggered in part by the clang (and icx 2023.1?) optimizers. In moving from icx2023.1 to icx2023.2 (which implies internally a move from clang16 to clang17), the optimizer has now triggered new floating point exceptions around sqrt operations, possibly around sqrt of negative values. As in my previous patch, most changes were in the ixx/oxx/vxx hardcoded functions, and again I fixed these with a generous use of `volatile`. However there were also FPEs in the test infrastructure itself, and I also added sqrt functions with volatile arguments to the fptype/vector headers. In some cases, I had a hard time understanding why there was a problem at all. Anyway, these are all fixed. This closes #736.
- While doing these tests, at some point I got an FPE deep down in libm, called by the googletest framework. This was when using ggogletest built with gcc, against madgraph4gpu in icx. For lack of better ideas, I tried to do a clean compiler-dependent rebuild of googletest with separate directories (as I had suggested already in #125). This fixed the issue, and I kept this in production. I even implemented a backward-compatible mechanism for epoch1/epoch2 in the CI, just in case. This fixes #738.

It should also be noted that another whole set of FPEs is still pending in #733. These are invalid (sqrt?), underflow and overflow FPEs. Initial investigations in WIP MR #706, in any case, suggest that this is due to some bug in coupling propagation in non-SM processes, rather than to the ixx/oxx/vxx functions. To be followed up... that is high priority as it is a blocker for ATLAS pp->ttW.

Volia that's all for the documentation of this MR.

cc @roiser @oliviermattelaer @hageboeck @Jooorgen @zeniheisser

PR #747, opened 2023-08-11
Call the patchMad.sh script from within the plugin's finalize

roiser:
first version of calling the patch script
patch script and patch files moved into the plugin directory


valassi:
I am doing further integration tests and patches, to make sure that calling this from generateAndCompare.sh also still works.

@roiser I am confused however because, unless I did something different from you, the patchMad.sh scripts is not called from finalize (or better, it is called but it does nothing because it fails and prints an error)
```
[avalassi@itscrd80 gcc11.2/cvmfs] /data/avalassi/GPU2023/MG5aMC/ghav-mg5amcnlo> more pippo 
set stdout_level DEBUG
set zerowidth_tchannel F
generate g g > t t~
output madevent CODEGEN_mad_gg_tt --hel_recycling=False --vector_size=32 --me_exporter=standalone_cudacpp

[avalassi@itscrd80 gcc11.2/cvmfs] /data/avalassi/GPU2023/MG5aMC/ghav-mg5amcnlo> ./bin/mg5_aMC pippo
...
INFO: Use Fortran compiler gfortran 
INFO: Use c++ compiler g++ 
INFO: Generate web pages 
DEBUG:  'Entering PLUGIN_ProcessExporter.finalize', self.in_madevent_mode =  Entering PLUGIN_ProcessExporter.finalize True [output.py at line 204] 
Usage: /data/avalassi/GPU2023/MG5aMC/ghav-mg5amcnlo/PLUGIN/CUDACPP_SA_OUTPUT/patchMad.sh <process.[madonly|mad]> <vecsize> <patch_dir> [--nopatch|--upstream]
Output to directory /data/avalassi/GPU2023/MG5aMC/ghav-mg5amcnlo/CODEGEN_mad_gg_tt done.
Type "launch" to generate events from this process, or see
/data/avalassi/GPU2023/MG5aMC/ghav-mg5amcnlo/CODEGEN_mad_gg_tt/README
Run "open index.html" to see more information about this process.
quit
```

In particular, `Usage: /data/avalassi/GPU2023/MG5aMC/ghav-mg5amcnlo/PLUGIN/CUDACPP_SA_OUTPUT/patchMad.sh <process.[madonly|mad]> <vecsize> <patch_dir> [--nopatch|--upstream]` is an error printout. It means that the script failed.

I will add an error check if possible. And also try to fix this

valassi:
I have fixed the two issues I mentioned before
- I have added an error check, now generation fails if patchMad.sh fails
- I have fixed all isssues so that now generation succeeds with patchMad.sh in finalize (and at least in my setup it gives the right thing)

To do still
- allow the patchMad to be used with any vecsize as input (now it assumes it was generated with 16384, so the hardcoded 32 in output.py will fail there)
- resurrect a "no-patch" mode in generateAndCompare: this is needed to create new patch.P1 and patch.common if needed
- final checks that the same code is generated in mg5amc and in my generateAndCompare script (modulo irrelevant details)
- regenerate all processes

valassi:
I will wait for the tests running tonight to succeed, and will merge tomorrow.

This will close issue #656. Note that the strategy for integrating patchMad.sh patches upstream has changed:
- this is no longer about getting rid of patchMad.sh and moving the patches to the common LO templates (as in my WIP MR #675, which has now been closed)
- instead, we keep patchMad.sh but we move it to the plugin itself so that it can be called from standard MG5AMC commands (this MR #747)

valassi:
The CI tests passsed - I am self merging

PR #751, opened 2023-08-15
Minor fixes for param_card.dat path

valassi:
This is the final MR from the AUG2023 branch including the work from last week's hackathon.

Specifically, this contains @hageboeck 's fixes for param_card.dat path and their integration in all other processes (plus some comments I added on initProc)


valassi:
The CI tests succeeded, I am self merging

Issue #661, opened 2023-05-23
upload plugins upstream to mg5amcnlo


valassi:
upload plugins upstream to mg5amcnlo

this is around points 2 and 3 in #576... it is maybe a bit of a new idea, as discussed one month ago

we should make sure mg5amcnlo knows where to find the cudacpp and sycl plugins (probably from two repositires in mg5amcnlo)

valassi:
Hi @oliviermattelaer (and @roiser @hageboeck @zeniheisser) this will come at some point.

Any suggestions?
- name of the repository? https://github.com/mg5amcnlo/cudacpp_plugin ?
- how should this look like inside the mg5amcnlo on disk? presently it is <top>/PLUGIN/CUDACPP_SA_OUTPUT, but definitely the SA does no longer make much sense... how about PLUGIN/CUDACPP_OUTPUT? (I imagine that then I need to modify this line https://github.com/madgraph5/madgraph4gpu/blob/f24bfbd85f1226fd61a3270e5e0005ea5d7773a1/epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/__init__.py#L31 ?)
- I guess that eventually maybe we need some hook to automatically download it? we can always tell people to do it manually to start with... but if you do think of an automated mechanism, maybe we need a naming convention? or Olivier you were just thinking of adding a database in mg5amcnlo, "if a strange plugin is required and you do not have it, look in this database if the plugin is declared to exist in a github repo...'?)

In practice, Olivier can I create the new repo?

Thanks
Andrea

oliviermattelaer:
Hi Andrea,

So my point of view is that as long as we use PLUGIN and that we do not want to ship them with MG5aMC.
Then the solution is to use the "install XXXXX" feature of MG5aMC.

On the madgraph repo itself, the only think that we need to do is to add XXXXX to the list of supported tools.
The rest of the handling (where to get the code/...) is done inside another code/repo (HEPToolsInstaller).

I personally care more about the name that we use for XXXXX that the name of the directory that we put in the PLUGIN directory and/or the name of the repo.
My suggestion might be to use two name for the XXXXXX, one suggestion would be
LO_SIMD
LO_CUDA
Then both can install the same plugin directory which can be CUDACPP or CudaSimd

@valassi Yes you can create the repo.

Cheers,

Olivier

valassi:
This is now almost completed.
[...]

PR #752, opened 2023-08-16
Include mg5amcnlo in madgraph4gpu as a git submodule

valassi:
Hi @oliviermattelaer @hageboeck @roiser @zeniheisser @Jooorgen 
I have completed the inclusion of mg5amcnlo in madgraph4gpu as a git submodule.

This MR is ready to be merged... but please let me know if you have comments or would have liked some things differently.

This completes our "step 1" agreed last week
- Step 1, mg5amcnlo is a submodule of madgraph4gpu, while cudacpp is still a directory in the madgraph4gpu repo
- Step 2, cudacpp is a separate repo, included as a submodule in mg5amcnlo, which is itself a submodule of madgraph4gpu

For info, what I personally did was the following
- by default, mg5amcnlo will be in detached HEAD state with a given commit (no branch)
- I went and added a remote for my fork, and then checked out the valassi/gpucpp branch from my fork as a local "valassi_gpucpp" branch which contains the commit. Then updating the module must be done with the merge option, `git submodule update --remote --merge`.
See section "Working on a Submodule" in https://git-scm.com/book/en/v2/Git-Tools-Submodules

I am also progressing on step2, i.e. the move of the codegen plugin to a separate repo (the only tricky bit is I am keeping the commit history for that and will try to change the links to existing github issues to point to madgraph4gpu - and then making a few tests to understand how to continue porting codegen changes in existing MRs like HIP). It should be ready in one day or two I think.

Thanks
Andrea


valassi:
Hi Stefan thanks. Ah ok now I understand what you mean. We had a misunderstanding between two cases.

1. You are right, if you checkout a different branch, and this branch has a different commit than gpucpp, then you risk committing the wrong commit hash. But if your branch has excatly the same content as gpucpp, then the commit is the same, and this should not change, can you check?

2. What I meant was excatly, that your branch has the same content as the commit hash. In that case, what I had understood from you was that we would risk changing the branch name in .gitmodules (not the commit hash in the submodule). In my understanding, this should NOT happen. Can you check?

valassi:
This is logically part oif #661, even if it is a slightly different issue. Step 2 will be a proper part of issue #661.

roiser:
> Hi Stefan thanks. Ah ok now I understand what you mean. We had a misunderstanding between two cases.
> 
>     1. You are right, if you checkout a different branch, and this branch has a different commit than gpucpp, then you risk committing the wrong commit hash. But if your branch has excatly the same content as gpucpp, then the commit is the same, and this should not change, can you check?

yes you are right, if the two branches "upstream" and "origin" are at the same stage the commit hash is also the same, so there won't be a problem

The only problem left I see is if they diverge then we should understand if one does two PRs one for the submodule and one for the commit hash outside or we "adjust" the commit hash in a different way once the submodule PR has been merged. 

> 
>     2. What I meant was excatly, that your branch has the same content as the commit hash. In that case, what I had understood from you was that we would risk changing the branch name in .gitmodules (not the commit hash in the submodule). In my understanding, this should NOT happen. Can you check?

no that's ok, no problem here. 
thanks Stefan

valassi:
Hi, good, thanks for checking.

> The only possible problem left I see is if they diverge then we should understand if one does two PRs one for the submodule and one for the commit hash outside or we "adjust" the commit hash in a different way once the submodule PR has been merged.

Hm I think that unfortunately we will need always two MRs anyway, one on madgraph4gpu and one on gpucpp. These are two different repos, I do not think we can combine change sto two repos in a single MR?

If indeed we need two MRs. then I guess it is better to first merge mg5amcnlo and only later merge madgraph4gpu, as the latter contains the hash to the former, not viceversa. This is almost exactly what I had with my commit.GIT anyway. Sometimes I was making changes in mg5amcnlo gpucpp in my fork and create a MR (so the hash would be in my fork anyway) and change madgraph4gpu even if that commit of gpucpp was not yet merged. But we should try to avoid it.

So I guess, ideally, first push/merge on mg5amcnlo gpucpp so that the commit hash is there, and then push/merge the reference to that hash in madgraph4gpu. Then sometimes there can be exceptions/deviations...

Issue #678, opened 2023-06-04
Check that Fortran results are the same with and without vector interfaces


valassi:
Around the time of the release (not a blocker technically, but necessary for physics validation), we should check that Fortran results are the same with and without vector interfaces.

I think Olivier is already making a lot of tests in this direction. But just opening this as an explicit issue as a reminder...

cc @oliviermattelaer @roiser

oliviermattelaer:
# Issue with chcluster

program that reproduce the issue:

```
define p = p / g
add process u g > t t~ j
output PROC_TEST --vector_size=8
launch
set nevents 500k
set ickkw 1
set mc_grouped_subproc F
set pdfwgt T
set chcluster T # Note this is the problematic parameter that impacts the black box.
```

This is the result that I expect:

    46.27 +- 0.02233 pb (correct value from 3.5.0)
    This is the result that I got:
    46.13 +- 0.02268 pb(vecsize=64) [4 sigma away]
    46.15 +- 0.02265 pb (vecsize=8) [3 sigma away]
    46.22 +- 0.02365 p (vecsize=2) [1 sigma away]
    46.26 +- 0.02338 pb (vecsize=1) [perfect]

The "problematic" lines are l466 SubProcesses/cluster.f
           if(chcluster)then
              if(id_cl(iproc,idij,i).ne.iconfig) cycle
           endif

And the problematic channel is here G4 (all other channel are just perfect spot on).
Note that 
  - G4 does not have any symmetric channel, so iconfig is always 4 
  - The above setup has only one matrix-element (so iproc=1)


Since the above line are part of the findmt function, I have check that the spectrum of output given by that function are the same in 3.x and this branch. However, I have noted that this function is call less often in the new branch. I do not worry about that (at least for the moment).

oliviermattelaer:
Information for the computation of scale uncertainty should be correctly store/written now.
I have open #708 to follow up in that direction and be sure that the writting of the event is correct in all cases

valassi:
Hi @oliviermattelaer thanks.
I reopen and assign to myself - this is a reminder that I need to move madgraph4gpu to your latest
https://github.com/mg5amcnlo/mg5amcnlo/commit/f98ca2586ae05c692b9907aa41b91b9ced2c8d34
Andrea

oliviermattelaer:
Hi re-open it for me (even if you did not reclose it --you should have right--
Just to report here the latest issue with pp > t t~ process giving wrong cross-section 
[...]

PR #754, opened 2023-08-23
avoid that mirroring is reset by the plugin

oliviermattelaer:
Hi @valassi,

This is a PR to fix the issue that 
p p > t t~ returns the wrong cross-section (compare to pure madevent case: (see #678)

The issue that I found is that the plugin reset the "mirror" flag to False and since this is done "before" the creation of the madevent metadata file, the madevent loose track of such symmetry (and therefore of a part of the cross-section).
This explains why g g > t t~ was not impacted but that q q > t t~ was wrong by a factor of two.

The question is why did you implement such line (and what is going to break by removing such line).
To avoid too much side effect, I have just removed them for the madevent+cpp/cuda mode?
But we likely need to take a more drastic option here (but want your opinion on it)

Olivier

### DIFF
--- a/epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/model_handling.py
+++ b/epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/model_handling.py
@@ -1299,9 +1299,10 @@ class PLUGIN_OneProcessExporter(PLUGIN_export_cpp.OneProcessExporterGPU):
             misc.sprint('self.include_multi_channel is already defined: this is madevent+second_exporter mode')
         else:
             misc.sprint('self.include_multi_channel is not yet defined: this is standalone_cudacpp mode') # see issue #473
-        if self.matrix_elements[0].get('has_mirror_process'):
-            self.matrix_elements[0].set('has_mirror_process', False)
-            self.nprocesses/=2
+            # I move those line to standalone_cudacpp mode (but do we need those at all???)
+            if self.matrix_elements[0].get('has_mirror_process'):
+                self.matrix_elements[0].set('has_mirror_process', False)
+                self.nprocesses/=2
         super(PLUGIN_export_cpp.OneProcessExporterGPU, self).generate_process_files()
         self.edit_CMakeLists()
         self.edit_check_sa()


oliviermattelaer:
let's merge this and then see if we break something (release first fix later strategy)

oliviermattelaer:
This patch fixes the issue for the fortran mode but not for the cuda/cpp mode.
In the cpp mode, all the events written on disk have (IMIRROR=1)
and the reason for that is clear when running the ../madevent_cpp < input_app.txt
```
 selecting           1           1           1
 selecting           1           1           1
 selecting           1           1           1
 selecting           1           1           1
 selecting           1           1           2
WARNING! flagging abnormal ME for ievt=0
WARNING! flagging abnormal ME for ievt=1
WARNING! flagging abnormal ME for ievt=2
WARNING! flagging abnormal ME for ievt=3
WARNING! flagging abnormal ME for ievt=4
WARNING! flagging abnormal ME for ievt=5
WARNING! flagging abnormal ME for ievt=6
WARNING! flagging abnormal ME for ievt=7
WARNING! flagging abnormal ME for ievt=8
WARNING! flagging abnormal ME for ievt=9
 Error in matrix element
 Error in matrix element
 Error in matrix element
 Error in matrix element
 Error in matrix element
 Error in matrix element
 Error in matrix element
 Error in matrix element
 Error in matrix element
 Error in matrix element
 selecting           1           1           2
WARNING! flagging abnormal ME for ievt=0
```

So all event with imirror=2 have a cpp issue and returns 0 (and therefore cross-section is ~.5 the correct value and events are biased in the way they are written.

oliviermattelaer:
I guess that the issue is that the generated code for the amplitude is

```
      // Wavefunction(s) for diagram number 1
#if not( defined __CUDACC__ and defined MGONGPU_TEST_DIVERGENCE )
      imzxxx<M_ACCESS, W_ACCESS>( momenta, cHel[ihel][0], +1, w_fp[0], 0 ); // NB: imzxxx o\
nly uses pz
#else
      if( ( blockDim.x * blockIdx.x + threadIdx.x ) % 2 == 0 )
        ipzxxx<M_ACCESS, W_ACCESS>( momenta, cHel[ihel][0], +1, w_fp[0], 0 ); // NB: imzxxx\
 only uses pz
      else
        ixxxxx<M_ACCESS, W_ACCESS>( momenta, 0, cHel[ihel][0], +1, w_fp[0], 0 );
#endif

      omzxxx<M_ACCESS, W_ACCESS>( momenta, cHel[ihel][1], -1, w_fp[1], 1 ); // NB: opzxxx o\
nly uses pz
```

which due to the "mz" and "pz" assumes the sign of the z direction of the events.
This is problematic when using mirroring since those assumptions are wrong. 

- weirdly using ixxxxxx seems problematic as well (100% of the event fail in that case)
- switching ipz to imz has identical issue (but event writting are now only the mirrored one)

valassi:
Hi @oliviermattelaer thanks for asking the review :-)
I had a look, in my tests I do not see the benefits (but I vaguely understand you may see a factor 2x in the xsec in your own tests), but I also see no problems. Or better, there were some hickups, but you fixed them in a later commit, so that's sll good for me.

PR #757, opened 2023-08-29
WIP: if needed fix the order of couplings 

### LINKED ISSUE #748: xsec from fortran and cpp differ in gq_ttq tmad tests (wrong order of couplings)
This is a follow-up to #630, which had originally been files about gq_ttq but eventually focused on gg_uu: xsec from fortran and cpp differ in gq_ttq tmad tests.

In #630, I had also found out that xsec from fortran and cpp also differed in gg_uu tmad tests, but with Zenny's help this was identified as a problem in getGoodHel and fixed. This however fixed gg_uu, but not gq_ttq, where xsecs are still different. For clarity, I move here gq_ttq, while I will soon close #630 about gg_uu.

roiser:
override the prepare_couplings function in the in PLUGIN_UFOModelConverter class
the additional functionality will re-order the dictionary keys of the dependent couplings such that they follow the correct ordering. The wanted_couplings parameter of the function contains the correct ordering as it is discovered in the OneProcessExporter class

fixes #748


oliviermattelaer:
I do approve this... (But we should do the merge in the other repo as well)

valassi:
Thanks @roiser ! I confirm this fixes #748. It introduced a few minor issues (gqttq ref log, eemumu generation), fixed later on.

PR #760, opened 2023-09-08
remove symbolik links in the .sa directories makefile->kokkos.mk

roiser:
because of the parallel existence of a file Makefile (upper case M) this creates a problem on standard MacOS filesystems which is case insensitive.


roiser:
self-merging

PR #761, opened 2023-09-08
fix running couplings ordering

roiser:
…e base class called "wanted_couplings", this list also contained the independent couplings which need to be removed / filtered out before applying the ordering


roiser:
self merging

PR #762, opened 2023-09-08
Remove smatrix counters and regenerate all processes

hageboeck:
- Add a script to regenerate code and automatically stage changes to git.
- Remove smatrix counters from smatrix function in Fortran MEs. The counter is never read, but dramatically slows down the matrix elements. See [here](https://indico.cern.ch/event/1318938/#3-round-table).
- Remove pickle files from version control
- Regenerate all processes. Due to mg5 updates, a few more things change:
  - zero-width t-channel is enabled for some processes
  - ident_card.dat is generated differently
  - The argument `FK_MDL_WT` for FFV functions changes to `ZERO`


roiser:
Thanks @hageboeck very good. Just a few comments inline. Only thing, especially for the last point of your initial message I would ask @oliviermattelaer to check that this is all good? 

I also see some re-ordering happening, e.g. in Cards files or a variable ALOHA_FUNCTIONS, do we understand that this is all good?

oliviermattelaer:
Hi,

Yes the ordering for the variable in the makefile is irrelevant and the ordering in that file too.
So I'm not worry about any of those changes.

Thanks a lot @hageboeck, It is ready to go!

hageboeck:
> I also see some re-ordering happening, e.g. in Cards files or a variable ALOHARoutine, do we understand that this is all good?

At least for the makefile, I can easily fix this in upstream mg5. For the cards, I haven't found the code yet.

@roiser 
Could you please check
https://github.com/madgraph5/madgraph4gpu/pull/762/files#diff-33c49abaf9d894a2f30e3eb3767639cef0f25755f905817f963fe343f43dd50e

Is this the coupling bug you were mentioning?

oliviermattelaer:
No need to fix it, those ordering have no meaning (and no impact)

oliviermattelaer (approved):
Here is an "official" comment to approve the merge

roiser:
> @roiser Could you please check https://github.com/madgraph5/madgraph4gpu/pull/762/files#diff-33c49abaf9d894a2f30e3eb3767639cef0f25755f905817f963fe343f43dd50e
> 
> Is this the coupling bug you were mentioning?

This is a fix done for respecting the order of the couplings as they appear in the code generation, see also #757 

There was another fix done last Friday which fixes the problem that the non-running couplings need to be filtered out before this re-ordering is applied, see #761

PR #764, opened 2023-09-12
New interface

oliviermattelaer:
This includes various bug fix but also a starting point for the new interface 
including
 - the possibility to use the launch command within MG5aMC
 - the possibility to control the backend via the run_card
 - the possibility to control vector_size via the run_card (recompilation correctly handle)


oliviermattelaer:
Looks like the update of the master branch by @hageboeck  creates an issue with this PR.
I'm now having this crash:
```
	The compilation fails with the following output message:
	    make: *** No rule to make target `madevent_cpp_link'.  Stop.
```

I'm investigating it (so sorry for having open the PR too early)

oliviermattelaer:
With this latest fix and on my mac.
- [x] tt~(0-2)j is working 
  - [x] from the launch commmand 
  - [x] from the ./bin/generate_events (but important to correctly set the run_card (and the first time) ! (in particular vector_size!)
- [x] DY 0+1j 
   - [x] from the launch commmand 
   - [x] from the ./bin/generate_events
- [ ] DY 0+1+2j: 
   - issue with "Note: The following floating-point exceptions are signalling: IEEE_INVALID_FLAG IEEE_DIVIDE_BY_ZERO"  
 
Improvment required (but let's not wait for those):
 - [ ] improve the handling of the vector_size in the run_card such that the code does not complains if the run_card does not defines it
 - [ ] checking recompilation if backend is changed
 - [ ] checking recompilation if compilation flag are updated
 - [ ] cudacpp backend does not appear in the default run_card (need to think how this would be handle)
 - [ ] define way to have shortcut for the complex option and better default value from the run_card (especially if parameter are not defined) (like for SDE_strategy)

So note that it is currently MANDATORRY to add within the run_card the definition of vector_size (otherwise it is set back to one and the code crash)

oliviermattelaer:
nice user interface are now available for the users:
Note that previous command is STILL working (so this should not break anything)
but here is the new command
```
generate p p > t t~
output madevent_simd MYPROC
launch
```
(which is equivalent to:
```
generate p p > t t~
output MYPROC --hel_recycling=False --vector_size=16 --me_exporter=standalone_cudacpp
launch
```
and for GPU
```
generate p p > t t~
output madevent_gpu MYPROC
launch
```
which has the same meaning (but a default vector_size=2048)

roiser:
This is really great !!! A huge simplification, thanks a lot 

I was just wondering if

> which has the same meaning (but a default vector_size=2048)

this vector_size is enough? Didn't we run with 16384 in the past, in the sense that we need to fill the GPU with events. The open question is maybe if there is a "physics case" for doing this?

hageboeck:
> > which has the same meaning (but a default vector_size=2048)
> 
> this vector_size is enough? Didn't we run with 16384 in the past, in the sense that we need to fill the GPU with events. The open question is maybe if there is a "physics case" for doing this?

It sounds a bit small to me. The sweet spot for gg-->ttgg(g) was as 8k and 16k, but I don't know this number for other processes.

oliviermattelaer:
Ok will change to 16k.
But this worries me since this is too large to avoid bias in the phase-space.

I guess that this means that we have to change the bridge such that the channel can either be set event by event or wrap by wrap. Would you be able to do that?

oliviermattelaer:
So I have updated it to 16k.

and started to work on a branch which will fix my worry about bias:
(where vector_size will be "small" (< 100) and a second number (nb_wrap) could be large (=512 for gpu)
https://github.com/madgraph5/madgraph4gpu/tree/new_interface_wrap
But that would be another merge request/discussion.

epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/madgraph/iolibs/template_files/gpu/process_function_definitions.inc
```diff
@@ -405,8 +405,7 @@ namespace mg5amcCpu
     {
       // nprocesses>1 was last observed for "mirror processes" in uux_ttx in the 270 branch (see issue #343 and PRs #360 and #396)
       constexpr int nprocesses = %(nproc)i;
-      static_assert( nprocesses == 1, "Assume nprocesses == 1" );
-      // process_id corresponds to the index of DSIG1 Fortran functions (must be 1 because cudacpp is unable to handle DSIG2)
+      static_assert( nprocesses == 1 || nprocesses == 2, "Assume nprocesses == 1 or 2" );
```
valassi, line 408: This is probably related to #872

PR #769, opened 2023-09-26
add missing vector.inc to the lhapdf wrapper

roiser:
When using lhapdf there is a fortran wrapper file around it. It includes the coupl.inc include which needs to be prepended with the usual include of vector.inc

This PR adds the corresponding patch to the code generating part of the plugin patching mechanism


oliviermattelaer:
Ok I'm confused (one more time today) because I thought I did merge the exact same fix last week... Whatever error I did does not matter, thanks for doing it

oliviermattelaer:
I will actually revert this change since it is incompatible with "my change" within the upstream madgraph (which is better on that point of view). Did I miss to update the upstream madgraph?

roiser:
Great that's very fine with me if its fixed upstream, pls go ahead. Good that we have code reviews ;-)

PR #770, opened 2023-09-27
remove the patch for the pdf wrapper, it was already upstream

roiser:
remove a fix for the pdf wrapper fortran file, this was already included in the upstream mg5amcnlo repo


roiser:
self-merging

PR #771, opened 2023-09-27
fix a problem when using sm models which are not standard

roiser:
This fixes a problem when using a model which is sm but not standard, e.g. a line

```
import model sm-no_b_mass
```

would produce an error in the model_handling.py. This PR will fix the problem, but I'm a bit wary about this piece of code in general. I merge this PR anyway now to let our users continue working but I think would be good to discuss this in more depth. I open an issue for it.


roiser:
this was tested and works, I self-merge this

PR #773, opened 2023-09-29
[Profiling/GitHub workflow] Added profiling infrastructure & improvements to GitHub Action workflows

Jooorgen:
This consists of all of work related to the profiling infrastructure and GitHub Action workflow improvements.

For full commit history look here: https://github.com/Jooorgen/madgraph4gpu/tree/master/tools/profiling


Jooorgen:
The 15 remaining checks are for gcheck on GPU, but since there is no active github runner with the a100 tag in this repo they are not run.

PR #775, opened 2023-09-29
Includes all the changes to the makefiles, adding new targets for CUDA/cppavx builds

Jooorgen:
This includes the new targets for the makefile, cudacpp.mk and cudacpp_src.mk files.

The new targets are: `cuda`, `cppnone`, `cppsse4`, `cppavx2`, `cpp512y` and `cpp512z`.


valassi:
This has been effectively merged when merging #798 which was based on it - thank again Jorgen!

Issue #759, opened 2023-09-07
Issue with DY plus two jet


oliviermattelaer:
DY +2 jets is crashing (even with fortran backend)

Note that the crash is linked to the second exporter and not to the madevent vectorization since the code runs correctly when not setting the CPP exporter

### cross-section 
- up to 2 jet:
  - [x] LTS:  1019 +- 1.676 pb
  - [x] pure fortran (vecsize=10):  1022.5173084926286
  - [x] cudacpp exporter, backend fortran:  1021 +- 1.584 pb
  - [x] cudacpp exporter, backend cpp: 1021 +- 1.459 pb
- up to 2 jet:
  - [x] LTS: 1090
  - [x] pure fortran (vecsize=10): 1090
  - [x] cudacpp exporter, backend fortran: 1090
  - [x] cudacpp exporter, backend cpp: 645
     -  Note: The following floating-point exceptions are signalling: IEEE_INVALID_FLAG IEEE_DIVIDE_BY_ZERO
     - more details is:
        ```
        WARNING! flagging abnormal ME for ievt=0 value is nan
        WARNING! flagging abnormal ME for ievt=2 value is nan
        WARNING! flagging abnormal ME for ievt=4 value is nan
        WARNING! flagging abnormal ME for ievt=6 value is nan
        WARNING! flagging abnormal ME for ievt=8 value is nan
        ```

PR #776, opened 2023-10-02
Fix DY + 2j

roiser:
There are two changes in this PR, one is to get the order of running and independent couplings right, the other to always make sure that the template parameter for the coupling access is correct.


roiser:
fixes #759

oliviermattelaer (approved):
Perfect for me. Thanks

PR #778, opened 2023-10-26
in CODEGEN, update gqttq ref file for runTest after fixing coupling order

valassi:
this is a minor issue missing in PR #757


valassi:
trivial PR, self merging, without waiting for the CI and with no review

PR #779, opened 2023-10-26
minor fixes (clang format and code comments)

valassi:
minor fixes (clang format and code comments)


valassi:
Another trivial PR, self merging with no review and no CI

Issue #628, opened 2023-04-03
wrong type argument to unary minus (FFV functions should not use "-COUPs" as argument)


valassi:
wrong type argument to unary minus (FFV functions should not use "-COUPs" as argument)

I have seen this issue in two places
- first in a SUSY process gg_tt of PR #625 (specifically https://github.com/madgraph5/madgraph4gpu/pull/625/commits/586195dced571514e6f5f903c2e4b4b7fb7fe9af), so I thought it was specifically a SUSY issue
- but then I also saw the same in a SM process gq_ttllq in PR #626 (specifically https://github.com/madgraph5/madgraph4gpu/pull/626/commits/2f571a1e25c87cad2371fea5b781ee769a8939fe), so actually it is not a SUSY issue

For SUSY for instance
https://github.com/madgraph5/madgraph4gpu/pull/625/commits/586195dced571514e6f5f903c2e4b4b7fb7fe9af
```
The CUDA_HOME=none HRDCOD=0 build also fails with
ccache g++  -O3  -std=c++17 -I. -I../../src -I../../../../../tools  -Wall -Wshadow -Wextra -ffast-math  -fopenmp -march=skylake-avx512 -mprefer-vector-width=256  -DMGONGPU_FPTYPE_DOUBLE -DMGONGPU_FPTYPE2_DOUBLE -DMGONGPU_HARDCODE_PARAM -DMGONGPU_HAS_NO_CURAND  -fPIC -c CPPProcess.cc -o CPPProcess.o
CPPProcess.cc: In function ‘void mg5amcCpu::calculate_wavefunctions(int, const fptype*, const fptype*, mgOnGpu::fptype*, fptype_sv*, int)’:
CPPProcess.cc:241:81: error: wrong type argument to unary minus
  241 |       FFV1_0<W_ACCESS, A_ACCESS, CD_ACCESS>( w_fp[3], w_fp[2], w_fp[4], -COUPs[1], &amp_fp[0] );
      |                                                                          ~~~~~~~^
CPPProcess.cc:251:62: error: wrong type argument to unary minus
  251 |       FFV1_1<W_ACCESS, CD_ACCESS>( w_fp[2], w_fp[0], -COUPs[1], cIPD[0], cIPD[1], w_fp[4] );
      |                                                       ~~~~~~~^
CPPProcess.cc:254:81: error: wrong type argument to unary minus
[...]

Issue #116, opened 2021-02-16
High CPU use in clock_gettime (only on itscrd03, not on pmpe04)?


valassi:
I wanted to reproduce a few performance numbers, in case we wanted to include some in the vchep paper. So I went back in time to older commits and tried to reproduce some numbers. In particular I went back to commit https://github.com/madgraph5/madgraph4gpu/commit/715d26ffe830c3c9794419c45e86d0c04f474365 because this is the first one where I added some numbers to the logs.

I got very surprised because I was unable to reproduce the c++ throughput on my itscrd03. Eventually, I understood that this node gives a VERY different performance (a factor 3 to 4 slower!) than an lxbatch GPU node (like the one I had used for the performance recorded in the commit log) or my physical node pmpe04.

The CPUs are similar Intel's. I wondered if there was a problem with memory access or swappiness, but it does not look like it. Eventually I played a bit with flamegraphs and found out that itscrd03 seems to spent 70% of the CPU time in clock_gettime, unlike the other nodes.

This is itscrd03
```
time ./check.exe 393216
***********************************
NumberOfEntries       = 393216
TotalTimeInWaveFuncs  = 3.760869e+00 sec
MeanTimeInWaveFuncs   = 9.564384e-06 sec
StdDevTimeInWaveFuncs = 4.128436e-06 sec
MinTimeInWaveFuncs    = 7.822000e-06 sec
MaxTimeInWaveFuncs    = 6.076190e-04 sec
-----------------------------------
NumMatrixElements     = 393216
MatrixElementsPerSec  = 1.045546e+05 sec^-1
***********************************
NumMatrixElements     = 393216
MeanMatrixElemValue   = 1.372012e-02 GeV^0
StdErrMatrixElemValue = 1.307144e-05 GeV^0
StdDevMatrixElemValue = 8.196700e-03 GeV^0
MinMatrixElemValue    = 6.071582e-03 GeV^0
MaxMatrixElemValue    = 3.374887e-02 GeV^0

real    0m7.065s
user    0m2.196s
sys     0m4.864s
```

And this is pmpe04 using the exact same executable (shared on AFS)
```
time ./check.exe 393216
***********************************
NumberOfEntries       = 393216
TotalTimeInWaveFuncs  = 1.143175e+00 sec
MeanTimeInWaveFuncs   = 2.907244e-06 sec
StdDevTimeInWaveFuncs = 6.344392e-07 sec
[...]

PR #780, opened 2023-10-28
Patches and test results over the latest PRs

valassi:
Hi Olivier, Stefan, Stephan, this is a comprehensive patch including several changes over the latest PRs. This is based on the current master including all PRs in the last two months, with some fixes that should be non controversial (clang format, build warnings, build errors etc), and other changes that instead need closer review (e.g. they may break Olivier's Mac). 

It also includes new test results for 'my standard processes', using a new Alma9 machine (including downfall mitigation, which seems to have no effect anyway), using the August codebase and the most recent codebase from this branch.

I try to give a summary below. Can you please review (I suggest that you focus on CODEGEN only and on the parts that are potentially controversial). Thanks!

---

Summary of changes (in addition to or including minor fixes already in PRs #778 and #779).

Non-controversial? changes:
- fix build error for eemumu float by adding missing `cxsmpl<float> operator*`
- fix build warning in counters.cc after #762
- fix clang format around Ccoeff following the fix for unary minus #628 
- add copyright and license, add comments, remove too many empty lines in python etc

More controversial changes, can you please review
- reenable lines commented out by Olivier in patchMad.sh: do this only in a special 'tmad mode' called from generateAndCompare.sh, and/or move the lines directly to the latter script (this in principle should also be non controversial and should not affect the full workflow?)
- go back from a ln to a cp for counters.cc and ompthreads.cc (Stephan, maybe yours was better: I just wanted to minimise differences so I got back my working version... but if you prefer and it works, I will try the ln)
- always dump stdout and stderr from subprocess.popen for patchMad.sh (eventually we may remove this maybe, but now I need this to more easily check the logs for errors and avoid silent failures... and personally I prefr to have it dumped anyway as it is a tricky part still)
- in cudacpp.mk, remove 'include ../../Source/make_opts' and remove '-fopenmp' (Olivier this is for you, why do you need that? for me it works without it)
- add a comment that in MatrixElementKernels.cc we should add the correct ifdef for Mac
- in launch_plugin.py, do not raise an Exception in reset_simd, but just print a warning (otherwise my generation fails: by the way I still use _cudacpp, not simd and not cuda)
- in patch.common, reenable some makefile fragments that had been commented out by Olivier (why? Mac?)

---

Coming to the performance and functionality tests.

I have the latest logs in these two commits:
https://github.com/madgraph5/madgraph4gpu/commit/ee83b62c1dcf164795debe5e838cb3561f4b309b
and
https://github.com/madgraph5/madgraph4gpu/commit/536ad49c0247d238d393ebcd15472405b147a277

1. The first log has 'tput' (check.exe, gcheck.exe, runTests.exe) results. This is a summary

Functionality:
** all NaN's have disappeared everywhere!
-- eemumu, ggtt, ggttg, ggttgg, ggttggg: all ok
** gqttq: some failures in runTest.exe

Performance:
-- eemumu: cuda ~20% slower (and 20% increase in registers used)
** eemumu: simd a factor 2 (inl0) to a factor 4 (inl1) slower
-- ggtt: cuda up to 10% slower
-- ggtt: simd ~10-15% faster (inl0), similar (inl1)
-- ggttg, ggttgg, ggttggg: generally similar performance, some minor speedups

Rephasing:
- GOOD: no NaNs anymore! (I would not be surprised if #733 will also disappear)
- TO DO: gqttq failures
- (bad? or who cares?... eemumu performance is degraded, I think by using ixxx instead of imzx)

2. The second log has 'tmad' tests (madevent executable: check xsec and lhe files). This is a summary

Functionality:
- ggttggg: madevent crashes
- gqttq: xsec differs again

Performance (Fortran overhead):
-- all very similar

Performance (MEs):
-- eemumu: fortran 20% faster, cuda slightly slower, simd a factor 2 to 3 slower
-- ggtt: fortran and simd 20% faster, cuda similar
-- ggttg: fortran 10% faster, simd and cuda similar
-- ggttgg: all very similar

+ERROR! ' ./madevent_fortran < /tmp/avalassi/input_ggttggg_x1_fortran > /tmp/avalassi/output_ggttggg_x1_fortran' failed
+d R # 5  >    -0.0    -0.0    -0.0     0.4     0.4
+d R # 6  >    -0.0    -0.0    -0.0    -0.0     0.4
+s min # 3>     0.0119716.0 29929.0 29929.0     0.0
+s min # 4>     0.0     0.0 29929.0 29929.0     0.0
+s min # 5>     0.0     0.0     0.0     0.0     0.0
+s min # 6>     0.0     0.0     0.0     0.0     0.0
+xqcutij # 3>     0.0     0.0     0.0     0.0     0.0
+xqcutij # 4>     0.0     0.0     0.0     0.0     0.0
+xqcutij # 5>     0.0     0.0     0.0     0.0     0.0
+xqcutij # 6>     0.0     0.0     0.0     0.0     0.0

+ERROR! xsec from fortran (0.26050333309703716) and cpp (1.2757941949814184) differ by more than 2E-14 (3.8974198518457603)

Rephrasing:
- performances are not surprising
- (I was expecting to see the throughput increase reported by Stephan in PR #762 ? also because I did have the impression I saw it at some point... or are we sure that this is not again a problem on one specific machine using system calls for timing as in #116 and the TSC source?)
- gqttq xsec is wrong again? I would therefore reopen #748... but Stefan please check (Stefan note: I am kind of sure this had been fixed by PR #757, but it was broken again by PR #761?)
- I think there is a new problem with ggttggg, I will open a new bug

---

Voila that's all. Thanks for reviewing!


valassi:
PS one more point for Olivier
- I commented out the copy of plugin_interface.py because I have file anywhere (did you forget to commit it to the repo?). If it is optional, we should add an "if file exists then copy else do nothing"

epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/output.py
```diff
@@ -1,7 +1,7 @@
 # Copyright (C) 2020-2023 CERN and UCLouvain.
 # Licensed under the GNU Lesser General Public License (version 3 or later).
 # Created by: O. Mattelaer (Sep 2021) for the MG5aMC CUDACPP plugin.
-# Further modified by: O. Mattelaer, S. Roiser, A. Valassi, Z. Wettersten (2021-2023) for the MG5aMC CUDACPP plugin.
+# Further modified by: S. Hageboeck, O. Mattelaer, S. Roiser, A. Valassi, Z. Wettersten (2021-2023) for the MG5aMC CUDACPP plugin.
```
oliviermattelaer, line 5: Here the "created by " me is weird,  if you want to have a main author per file, please put A. Valssi and not me as original author

epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/output.py
```diff
@@ -204,23 +204,40 @@ def finalize(self, matrix_element, cmdhistory, MG5options, outputflag):
 	    MG5options are all the options of the main interface
 	    outputflags is a list of options provided when doing the output command"""
         misc.sprint('Entering PLUGIN_ProcessExporter.finalize', self.in_madevent_mode, type(self))
+        misc.sprint(self.in_madevent_mode)
```
oliviermattelaer, line 207: misc.sprint need to be removed

epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/output.py
```diff
@@ -204,23 +204,40 @@ def finalize(self, matrix_element, cmdhistory, MG5options, outputflag):
...
-#            misc.sprint(path)
-            p = subprocess.Popen([pjoin(plugin_path, 'patchMad.sh'), self.dir_path , 'PROD', str(patchlevel)],
-                                 stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+            ###p = subprocess.Popen([pjoin(plugin_path, 'patchMad.sh'), self.dir_path , 'PROD', str(patchlevel)],
+            ###                     stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+            p = subprocess.Popen([pjoin(plugin_path, 'patchMad.sh'), self.dir_path , 'PROD', str(patchlevel)]) # AV always dump patchMad.sh stdout/stderr
             stdout, stderr = p.communicate()
-            if p.returncode != 0:
+            misc.sprint(p.returncode)
+            if p.returncode != 0: # AV: WARNING! this may silently fail, for instance because 'madevent treatcards run' may silently fail
```
oliviermattelaer, line 234: treatcards should not be called within patchMad.sh
the comment would therefore be not relevant.
(we need another way to have the setup that you want (which can be a script with a bunch of set command)

epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/output.py
```diff
@@ -204,23 +204,40 @@ def finalize(self, matrix_element, cmdhistory, MG5options, outputflag):
...
                 logger.debug("####### \n stdout is \n %s", stdout)
                 logger.info("####### \n stderr is \n %s", stderr)
                 logger.info("return code is %s\n", p.returncode)
                 raise Exception('ERROR! the O/S call to patchMad.sh failed')
-            
-            self.add_madevent_plugin_fct()
+            # Additional patching (OM)
```
oliviermattelaer, line 239: This is not patching (it is the creation of additional file) so not patch

epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/output.py
```diff
@@ -251,17 +268,15 @@ def add_madevent_plugin_fct(self):
         which contains a series of functions and one dictionary variable TO_OVERWRITE
         that will be used to have temporary overwrite of all the key variable passed as string by their value.
         all variable that are file related should be called as madgraph.dir.file.variable
-        """
-        
+        """        
         plugin_path = os.path.dirname(os.path.realpath( __file__ ))
-        files.cp(pjoin(plugin_path, 'plugin_interface.py'), pjoin(self.dir_path, 'bin', 'internal'))
+        ###files.cp(pjoin(plugin_path, 'plugin_interface.py'), pjoin(self.dir_path, 'bin', 'internal')) # AV FIXME (added by OM, but file is missing?)
         files.cp(pjoin(plugin_path, 'launch_plugin.py'), pjoin(self.dir_path, 'bin', 'internal'))
```
oliviermattelaer, line 274: Ok will take a look (maybe we do not need this anymore (or I forget to push that file)

epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/patchMad.sh
```diff
@@ -40,51 +40,59 @@ dir_patches=$2
...
-#\rm -f ${dir}/index.html
-#\rm -f ${dir}/madevent.tar.gz
-#\rm -f ${dir}/Cards/delphes_trigger.dat
-#\rm -f ${dir}/Cards/plot_card.dat
-#\rm -f ${dir}/bin/internal/run_plot*
-#\rm -f ${dir}/HTML/*
-#\rm -rf ${dir}/bin/internal/__pycache__
-#\rm -rf ${dir}/bin/internal/ufomodel/__pycache__
-#touch ${dir}/HTML/.keep # new file
+# AV Recover special 'tmad' mode used by generateAndCompare.sh, after OM's changes that commented this out in patchMad.sh
```
oliviermattelaer, line 43: Would not make more sense to have all those lines in the generateandcompare.sh script? (But no strong feeling) but maybe add a comment here that in "normal" mode tmadmode is zero

oliviermattelaer (changes requested):
I think this is good, I will need to check that it is still working on my mac, 
I have put comment in the code but no strong feeling at all.

I would say to merge this asap, and then check on mac what the status is (and do another PR if relevant --pretty sure that we will have one issue with one line that you revert back where we will need to sit together to find a way to have the flag when needed and missing when needed.)

valassi:
Hi @oliviermattelaer thanks! I will have a look tomorrow at your change requests.

In the meantime, I did manage to log in on your Mac (!thanks!) and I fixed all issues there.
This also nicely fixes the build errors in the CI on Mac, so the CI is all green.

Can you maybe check if at least the mac build part is ok for you too now? (I guess it should, I use the environment out of the box on your M1).

Thanks

oliviermattelaer:
Here is my run on mac: 
# Fortran mode (with vector phase-space integration): 
```
define q = p / g
generate p p > e+ e-
add process  p p > e+ e- j
add process p p > e+ e- j j
output madevent_simd PROC_TEST
launch
set cudacpp_backend Fortran
set vector_size 256
set nevents 25k
```
(so I put those command into a file and do ./bin/mg5_aMC PATH_TO_FILE)

and it crash with
```
	    gfortran -o gensym symmetry.o idenparts.o  -L../../lib/   -L../../lib/ -ldhelas -ldsample -lmodel -lgeneric -lpdf -lcernlib  -lbias  -lstdc++  -Wl,--no-relax
	    ld: unknown option: --no-relax
```
So I confirm that this flag is problematic (at least on my laptop, looks like you had it working on my desktop --which is more uptodate compare to my laptop)
If i revert the responsible line (so recommenting that line).
I get the following result.

```
  === Results Summary for run: run_01 tag: tag_1 ===

     Cross-section :   1090 +- 1.574 pb ("MG5 v3.5.2 returns  1087 +- 1.986 pb)
     Nb of events :  25000
```

allowing SIMD (so setting the backend to cpp in the script above):
it crashes with the following message:
```
	    f77 -I. -c fcheck_sa.f -o fcheck_sa.o
	    make: *** ../../../../../test: No such file or directory.  Stop.
	    make: *** ../../../../../test: No such file or directory.  Stop.
	    make[1]: f77: No such file or directory
	    make[1]: *** [../../../../../test/googletest/install_unknown/lib64//libgtest_main.a] Error 2
	    make[1]: *** Waiting for unfinished jobs....
	    make[1]: *** [../../../../../test/googletest/install_unknown/lib64//libgtest.a] Error 2
	    make[1]: *** [fcheck_sa.o] Error 1
	    make: *** [.cudacpplibs] Error 2
	    make: *** Waiting for unfinished jobs....
```

Which indicates two issues here
1. the fact that we should not compile the googletest in production
2. the fact that you do not follow the Source/make_opts which is the file that control the various compiler (and compiler flag) for all the madevent run (and it will force the compiler to be gfortran in my setup --or in most cases--). Having Source/make_opts used in ALL makefile is important (certainly in production but also in development such that we know that such file is impacting any compilation). If you think that this is a bad idea (I guess you prefer environment variable?), then we will need to remove that file from the MG5aMC branch and replace it by something else ...

Actually adding the "sourde of make_opts", remove the f77 issue but the code still fails for google test 
(but likely related to the way my directory are organised on my laptop). (Note that CMS has also issue with the google test).

Here I tried to remove the googletest by setting "TESTDIR=", not sure that this is the correct way to do it...

next error after that:
```
 Building in BUILDDIR=. for tag=sse4_d_inl0_hrd0_hasNoCurand (USEBUILDDIR is not set)
	    /Library/Developer/CommandLineTools/usr/bin/make -C
	    /Library/Developer/CommandLineTools/usr/bin/make -C
	    gfortran -I. -c fcheck_sa.f -o fcheck_sa.o
	    /Library/Developer/CommandLineTools/usr/bin/make: option requires an argument -- C
	    Usage: make [options] [target] ...
```

This is the first time that I see that one. So I do not know how to fix it...

valassi:
Hi Olivier, ah thanks then your desktop is ok but not your laptop?
- Can you try your self on your desktop just to be sure?
- And if indeed it only fails on the laptop, can you try to update the laptop, or alternatively allow me access? I cannot debug if I have no access to the same type of machine... (by the way, on the CI Mac it seems to work as well)

hageboeck:
> (I was expecting to see the throughput increase reported by Stephan in PR https://github.com/madgraph5/madgraph4gpu/pull/762 ? also because I did have the impression I saw it at some point... or are we sure that this is not again a problem on one specific machine using system calls for timing as in https://github.com/madgraph5/madgraph4gpu/issues/116 and the TSC source?)

That's exactly what it is, but I assume that you tested on a machine that is "lucky", and I run on a machine that's "unlucky" w.r.t. the clock sources. Given that we don't know where madgraph is going to run, I think the best course of action is to minimise the call to timers, so it's fast on all machines. Does that make sense?


The Makefile changes will create conflicts with mine #753 and Jørgen's work #775, but I think the only way is to merge this and I try to solve the conflicts again. Given that I have some time this week, I would like to rebase asap, to get the Makefile work done before more changes are introduced.

valassi:
> That's exactly what it is, but I assume that you tested on a machine that is "lucky", and I run on a machine that's "unlucky" w.r.t. the clock sources. Given that we don't know where madgraph is going to run, I think the best course of action is to minimise the call to timers, so it's fast on all machines. Does that make sense?

Hi @hageboeck thanks it makes perfect sense. I think it was good to remove useless code anyway! 

I am just very surprised that 
- we still have machines around with TSC issues (which one is it by the way?), an dpotentially this issue may reappear elsewhere
- we have been lucky that most of the overhead came from the 'useless' counters, while we do still have 'useful' counters enabled (I need those all the time) and they do not give an overhead... can you estimate on your machine what fraction of some typical workloads actually go in clock system calls? thanks!

hageboeck:
> 
> * we still have machines around with TSC issues (which one is it by the way?), an dpotentially this issue may reappear elsewhere

The A100

> * we have been lucky that most of the overhead came from the 'useless' counters, while we do still have 'useful' counters enabled (I need those all the time) and they do not give an overhead... can you estimate on your machine what fraction of some typical workloads actually go in clock system calls? thanks!

0.2%
See [here](https://shageboe.web.cern.ch/madgraph/profile_pp_tt0-2j//flame_ggttgg.mad_FOR.svg?s=hrtimer), search for hrtimer.

valassi:
Hi @oliviermattelaer ,  I just pushed a few changes to address your comments.

Can you have a look if it looks better?

Anyway my main messages for Mac are: as you suggested, lets push this PR first, and fix mac later; and for mac, I really would need a reproducible environment where I can log in. It turns out that your desktop I was using my gfortran downloaded last year, and instead Stefan is using MacPorts which is probably not what you are using. We should try a 'default' environment (as default as it may be, since there is no fortran by default IIUC).

Thanks!

valassi:
Hi @oliviermattelaer @roiser @hageboeck can this be merged?

As discussed, I would focus on mac later on.

epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/launch_plugin.py
```diff
@@ -50,15 +47,14 @@ def compile(self, *args, **opts):
...
 
 class CPPRunCard(banner_mod.RunCardLO):
-
     def reset_simd(self, old_value, new_value, name):
         if not hasattr(self, 'path'):
-            raise Exception
-
+            logger.warning('WARNING! CPPRunCard instance has no attribute path')
+            return
+            ###raise Exception('INTERNAL ERROR! CPPRunCard instance has no attribute path')
```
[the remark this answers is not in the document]
oliviermattelaer: Could you comment on this? This really scare me to convert an error to a warning without understanding what is going on. Could you comment on how you trigger it? This should not be put as resolved as long as this is not understood. (we can open an issue on this if needed to not block the merge but this MUST be understood. This should not be silently bypass this is super dangerous for the interface.

epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/output.py
oliviermattelaer: Please do not resolve, stuff not yet resolve.
So I believe that this was for an old attempt to modify the run_card.dat (the attempt I did during the hackaton).
So the best is to fully remove that line. Thanks for forcing me to look at this line.

oliviermattelaer (approved):

oliviermattelaer:
Good for me,
For me having it working on mac is actually crucial otherwise, I would not be able to test the non fortran part.
I have put some commit allowing to fix compilation issue on my mac.
But this is still not working:
log from MG5:
```
Error when reading /Users/omattelaer/Documents/git_workspace/3.1.1_lo_vectorization/PROC_TEST/SubProcesses/P1_uux_epem/G1/results.dat
Command "generate_events run_01" interrupted with error:
Exception : Reported error: End code 134.0
	 Full associated log:

	ls status:
	input_app.txt
	run1_app.log

```
when running executable mannually:
```
[G1]$ ../madevent < input_app.txt
madevent(17776,0x201efe500) malloc: *** error for object 0x106c7c7d0: pointer being freed was not allocated
madevent(17776,0x201efe500) malloc: *** set a breakpoint in malloc_error_break to debug
Abort trap: 6
```

oliviermattelaer:
I fix the last issue with the pointer. (it was brew that messed up with something, so I have removed fully brew from my system).

But I confirm that I do need to have make_opts to have it compiling on mac.
(More precisely, I need to have gfortran setup as FC which is one of the things done in make_opts).

So one hacky solution to have this PR pushed, is to hardcode FC=gfortran within your makefile (but this is obviously not sustainable). I have pushed that. and will open an issue to track andrea issue with make_opts

oliviermattelaer:
So the make_opts issue is tracked here: https://github.com/madgraph5/madgraph4gpu/issues/787

valassi:
Ok I have the impression that we converged :-)
Thanks @oliviermattelaer!

There is one pending issue in #790 about reset_simd, which is not understood. I assign this to you.

About make_opts, I fixed #787.
Just one point on Mac and gfortran: GNU make assumes FC=f77 if FC is not defined. So indeed in my Linux environment I explicitly export FC=gfortran. But I think this is the correct way
https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html 

As discussed, I then consider this as ready to go.
(@roiser and @hageboeck, last chance for comments!).
I will regenerate all processes and launch all tests first.
So I will merge this evening I think.

roiser (approved):
NB: I'm have not looked at makefile changes as those will probably be changed soon again

epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/madgraph/iolibs/template_files/gpu/cudacpp.mk
```diff
@@ -233,6 +233,8 @@ override OMPFLAGS = -fopenmp
 else ifneq ($(shell $(CXX) --version | egrep '^(Apple clang)'),)
 override OMPFLAGS = # AV disable OpenMP MT on Apple clang (builds fail in the CI #578)
 ###override OMPFLAGS = -fopenmp # OM reenable OpenMP MT on Apple clang? (AV Oct 2023: this still fails in the CI)
+else ifneq ($(UNAME), 'Darwin')
```
valassi, line 247: @oliviermattelaer I just realised there are two bugs here
- It should be ifeq not ifneq
- It should be UNAME_S not UNAME
That is it should read `else ifeq ($(UNAME_S), 'Darwin')`
The way it is done now disables OpenMP on Mac AND ALSO on Linux (except icpx and clang)
I will fix this

valassi: Ouf. There were three errors. No quotes needed. It should read `else ifeq ($(UNAME_S),Darwin)`. Fixing this again

valassi:
All tests in the CI including Mac are now good. Reviews addressesd. I am finally self merging.

PR #782, opened 2023-10-28
preserve order of coupling when filtering dependent ones out

roiser:
when filtering out the running couplings from wanted_couplings, preserve the order, the trick is to swap the two containers


valassi:
Thanks a lot Stefan! I confirm that it seems to work, I merge this

PR #784, opened 2023-11-01
support GPU builds without the curand library (incomplete patch)

roiser:
When on a machine without a curand library the code would currently not compile, mainly because of check.exe and gcheck.exe. The patches in this PR shall fix this problem. 

NB: There is an #undef MGONGPU_HAS_NO_CURAND in src/mgOnCpuConfig.h which probably also needs to be removed


valassi:
Hi @roiser, as discussed: I changed your PR to point to a new branch sapta_curand on upstream.
I will merge your changes and then complete them in a new PR.

PR #785, opened 2023-11-01
support GPU builds without the curand library (complete patch)

valassi:
This is a patch completing @roiser's previous changes in PR #784.

I have packaged this into a CODEGEN-only patch and removed gg_tt.mad changes for simplicity.


valassi:
Hi @roiser, as we discussed this together and it is relatively minor, I will self merge

PR #788, opened 2023-11-03
allow to choose fptype via the run_card

oliviermattelaer:
Hi Andrea,

This is one example (that I likely want to improve a bit) of a case where I want to pass information trought make_opts.
In this case the run_card is triggering a re-creation of make_opts in order to include the definition of FPTYPE requested by the user.

The point I want to improve is to include the check that fptype is correct at the compilation of the Source directory (for the moment it is at the level of the P directories.

So this relate to your request: https://github.com/madgraph5/madgraph4gpu/pull/780.

Olivier


oliviermattelaer:
Hi Andrea,

Could we move forward on this PR?

Olivier

epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/launch_plugin.py
```diff
@@ -57,13 +85,47 @@ def reset_simd(self, old_value, new_value, name):
...
+        self.add_param('cudacpp_backend', 'CPP', include=False, hidden=False,
+                       allowed=['Fortan', 'CPP', 'CUDA'])
+        self['vector_size'] = 16 # already setup in default class (just change value)
+        self['aloha_flag'] = '--fast-math'
+        self['matrix_flag'] = '-O3'
+        self.display_block.append('simd')
+        self.display_block.append('psoptim')
+
+
+>>>>>>> 746f16476 (better formatting for the card from the start and adding the possiblitity to choose avx)
```
valassi, line 128: This line is a mistake. I will try to clean this up and merge with my PR #793

valassi (commented):
Hi @oliviermattelaer thanks. 

I found one minor (but blocking) issue, that I will fix.

I am a bit puzzled howver who creates vector.inc now. All tghe comments I had added in the header have now disappeared.

I also saw a 'standalone_cudapp' directory appear in MG5AMC/mg5amcnlo in one test, I need to understand this better.

If that's ok with you, I would rather push through #793 (today) and then come back to this one (next week).

Sounds ok? Thanks Andrea

valassi (changes requested):
Hi @oliviermattelaer I fixed a minor blocker issue, an dnow it builds ok for me.

NB I have not tried to exercise this in "my" setup, and I will most likely need to make further changes.

One question however for you ("request for change"?) is about vector.inc: I do not understand why previously the vector.inc was generated with all of my comments. Also, there are vector.inc files that end up in the MG5AMC/mg5amcnlo directory after generation. So I would try to make sure that
- we get again the comments (like "THIS FILE CANNOT CONTAIN #ifdef DIRECTIVES" etc)
- we do not produce vector.inc files elsewhere

My vague impression is that there are issues related to the "path" (it's the same issue I had mentioned last week, you had an exception that I had turned into a warning, maybe something around that area).

Maybe this comes from my calling treatcards run? Not sure, I have no time to debug this now, maybe it rings a bell to you?

Thanks Andrea

oliviermattelaer:
Hi, 
let me reply now, but let's discuss this next week. (I'm just stuck since my laptop is running an heavy production for next week talk).

I do not think that I have changed the handling of vector.inc in this particular PR. (I did apply changes in the handling of make_opts but none in the handling of vector.inc. So this discussion is likely better in another issue (and/or PR)

Concerning the handling of vector.inc this file is first generated by the fortran output of the code.
either if you specify "--vector_size=XXXX" or if you use the plugin syntax (madevent_simd or madevent_gpu) that force some default value without the need to specify such value.

Now when you are handling the run_card.dat a check is done if the current vector_size requested in the run_card is 
the same or not compare to the value of vector.inc.
If the new value is bigger than the one within vector.inc, then the code will force recompilation of the simd/gpu library for the new value.
What I'm not fully sure if what happens for the fortran side if the value was modified. ( Did I re-create vector.inc and re-compile the fortran code or not).
So yes, this code is created at generation time and modify at launch time to follow the run_card.

For the "path" issue, this is on my todo list (but that will likely be another PR, likely targetting gpucpp with MG5aMC and not the plugin).

valassi:
Thanks Olivier, lets discuss next week. I will show you what happens in practice. Probably there is some unintended side effect somewhere, lets debug it together next week if confirmed.

valassi:
I have merged master into this (which was a bit tricky as there were a couple conflicts to solve). Lets discuss next week anyway.

valassi (approved):
Hi Olivier, thanks a lot. I fixed with a hack the vector.inc issue (I generate with comments in export_v4.py, then if updates are needed via banner.py and launch_plugin.py I simply keep the existing comments and update the rest).

I also took the opportunity to include your fix for #790.

I will resrun the CI tests and then merge

valassi:
Thanks Olivier I will now merge this.

Note: I tested individual madevent applications, but I have not tested the launch workflow. SO I have not tested if the runcard actually works here.

But I merge this anyway.

PR #774, opened 2023-09-29
Adding HIP/CUDA Abstraction to CODEGEN and gg_ttgg.mad

Jooorgen:
This PR adds the HIP/CUDA GPU abstraction.

PR #718, opened 2023-06-21
[WIP] GPU Abstraction to also target HIP in cudacpp impementation & Profiling infrastructure

Jooorgen:
Current work from the GPU abstraction I have been working on. Also includes the profiling infrastructure, but bear in mind the SYCL portion is not working as intended currently. Need some documentation from Nathan on how to compile it first. 

Just ran it successfully on LUMI-G with MI250x GPUs. On a sidenote I seem to have an issue running on the GPUs we got from LHCb, but on LUMI it apperantly runs fine. 

LHCb development GPUs (MI250x):
```
[jteig@n4051701 P1_epem_mupmum]$ ./gcheck.exe 32 32 10

ERROR! assertGpu: 'no ROCm-capable device is detected' (100) in ./GpuRuntime.h:62

gcheck.exe: ./GpuRuntime.h:21: void assertGpu(hipError_t, const char *, int, bool): Assertion `code == gpuSuccess' failed.

Aborted (core dumped)
```

VS

LUMI-G GPUs (MI250x):
```
teigjorg@nid005027:~/madgraph4gpu/epochX/cudacpp/ee_mumu.mad/SubProcesses/P1_epem_mupmum> ./gcheck.exe 32 32 10

..........

teigjorg@nid005027:~/madgraph4gpu/epochX/cudacpp/ee_mumu.mad/SubProcesses/P1_epem_mupmum>
```

PR #800, opened 2024-01-25
NOT TO BE MERGED - ("jthip24") Jorgen's earlier work on HIP before PR #774 and my fixes on that 

valassi:
This is a temporary PR that I file here only for reference and to keep the relevant code. I will close it immediately after creating it.

This PR uses my "jthip24" branch, which is itself an addition over Jorgen's branch
https://github.com/Jooorgen/madgraph4gpu/tree/gpu_abstraction
Jorgen's "gpu_abstraction" includes HIP work but also CI/profiling work by Jorgen. 
My "jthip24" branch includes the above, with the addition of some extra fixes which I was working on together with Jorgen in July-August (and some more recent ones).

Jorgen's "gpu_abstraction" branch never made it into a PR. 
Instead, Jorgen stripped off a smaller/different "gpu_abstraction_only" branch which he used to open PR #774.
https://github.com/Jooorgen/madgraph4gpu/tree/gpu_abstraction_only

When I tried to work over PR #774, I had several issues from "gpu_abstraction_only", which I had already fixed with Jorgen in this "gpu_abstraction". Therefore I had to go back and forth between the two branches to compare and pick up the better code. This is in my "jt774" branch which I will use for an upcoming PR.

I file this PR (which I will immediately close) for two reasons
- First, so that the full jthip24 is preserved anyway, in case of issues
- Second, because we may also decide later to pick up the CI and profiling changes by Jorgen (which he stripped off to PR #773), and it can be useful to have this.

Note that I have merged jt774 into this jthip24, so now the two lines of developmente have converged again, just in case.

Note also that Jorgen's "gpu_abstraction" is now what he used for PR #718, which instead is based on his master.


valassi:
As discussed, I am closing this immediately.

PR #816, opened 2024-02-06
Fix default run_card for GPU 

oliviermattelaer:
With Stefan we noticed that 
"output madevent_gpu" was producing a run_card with the default value set for SIMD mode...
This patch is fixing that.


oliviermattelaer:
Can i merge this?

roiser:
IIUC this will fix the problem that Jin has reported?

valassi:
Hi @oliviermattelaer sorry I will also look at this one in the afternoon. It changes the generated vector.inc, not a problem but I need to update a few scripts/tests.

oliviermattelaer:
Ok, I have therefore done a rebase of the master within this branch to avoid to have to do a merge (since I do not expect everyone to branch from this branch

valassi (approved):
This is good to go for me

Issue #821, opened 2024-03-06
Bug in GPUFOHelasCallWriter format_coupling (wrong params2order for SUSY processes)


valassi:
Hi @oliviermattelaer I think I hit a bug in your original implementation of GPUFOHelasCallWriter format_coupling (which I then overloaded in the cudacpp plugin, but keeping the same logic and hence the same bug).

The result of the bug is that params2order has weird values like the following, where two parameters have the same index,
```
{'mdl_Msu3': 3, 'mdl_Wsu3': 1, 'mdl_Msu6': 2, 'mdl_Wsu6': 3}
```
while it should instead be (I think)
```
{'mdl_Msu3': 0, 'mdl_Wsu3': 1, 'mdl_Msu6': 2, 'mdl_Wsu6': 3}
```

The problem is this code
```
    def format_coupling(self, call):
        """Format the coupling so any minus signs are put in front"""
        import re
        ###print(call) # FOR DEBUGGING
        model = self.get('model')
        newcoup = False
        if not hasattr(self, 'couplings2order'):
            self.couplings2order = {}
            self.params2order = {}
        if not hasattr(self, 'couporderdep'):
            self.couporderdep = {}
            self.couporderindep = {}
        for coup in re.findall(self.findcoupling, call):
            if coup == 'ZERO':
                ###call = call.replace('pars->ZERO', '0.')
                call = call.replace('m_pars->ZERO', '0.') # AV
                continue
            sign = ''
            if coup.startswith('-'):
                sign = '-'
                coup = coup[1:]
            try:
                param = model.get_parameter(coup)
            except KeyError:
                param = False
            if param:
                alias = self.params2order
                name = 'cIPD'
            elif model.is_running_coupling(coup):
                alias = self.couporderdep
                name = 'cIPC'
            else:
                alias = self.couporderindep
                name = 'cIPC'
            if coup not in alias:
                if alias == self.couporderindep:
                    if not len(alias):
                        alias[coup] = len(self.couporderdep)
                    else:
                        alias[coup] = alias[list(alias)[-1]]+1
                else:
                    alias[coup] = len(alias)
                if alias == self.couporderdep:
                    for k in self.couporderindep:
                        self.couporderindep[k] += 1
                newcoup = True
            if name == 'cIPD':
                call = call.replace('m_pars->%s%s' % (sign, coup),
                                    '%s%s[%s]' % (sign, name, alias[coup]))                        
            elif model.is_running_coupling(coup):
[...]

Issue #820, opened 2024-03-03
Further issues in builds of other SUSY processes


valassi:
The builds for one simple process gg>tt~ in the SUSY model is now fixed by PR #625 (for both HRDCOD=0 and HRDCOD=1).

I tried to investigate a few more SUSY processes, and unfortunately I still found other issues (which sometimes look very similar to those I already fixed).

- In "'import model MSSM_SLHA2; generate g g > t t~ t t~", both builds succeed, Strangely however the ME results are different in HRDCOD=1 with respect to the reference file produced with HRDCOD=0
```
  ME  4.578319386090301e-08
r.ME  4.847626218542335e-08
```

- In "import model MSSM_SLHA2; define q = u c d s u~ c~ d~ s~; generate g q > t t~ q", the HRDCOD=0 and HRDCOD=1 builds succeed and give the same results as each other. This is true in both P1 subdirectories.

- In "import model MSSM_SLHA2; generate g g > go go", the HRDCOD=1 build succeeds, but the HRDCOD=0 build fails
```
    ccache g++  -O3  -std=c++17 -I. -I../../src -Wall -Wshadow -Wextra -ffast-math  -fopenmp  -march=skylake-avx512 -mprefer-vector->
    CPPProcess.cc: In member function ‘virtual void mg5amcCpu::CPPProcess::initProc(const string&)’:
    CPPProcess.cc:554:38: error: ‘class mg5amcCpu::Parameters_MSSM_SLHA2’ has no member named ‘mdl_bsmIndepParam’; did you mean ‘nBs>
      554 |       memcpy( bsmIndepParam, m_pars->mdl_bsmIndepParam, Parameters_MSSM_SLHA2::nBsmIndepParam * sizeof( double ) );
          |                                      ^~~~~~~~~~~~~~~~~
          |                                      nBsmIndepParam
```

- In "import model MSSM_SLHA2; generate g g > t1 t1~", HRDCOD=0 builds fail
```
ccache g++  -O3  -std=c++17 -I.  -fPIC -Wall -Wshadow -Wextra -ffast-math  -fopenmp  -march=skylake-avx512 -mprefer-vector-width=256  -DMGONGPU_FPTYPE_DOUBLE -DMGONGPU_FPTYPE2_DOUBLE -fPIC -c Parameters_MSSM_SLHA2.cc -o Parameters_MSSM_SLHA2.o
In file included from Parameters_MSSM_SLHA2.cc:15:
[...]

PR #822, opened 2024-03-06
Fix bug in GPUFOHelasCallWriter format_coupling

valassi:
Hi @oliviermattelaer this is the PR closing #821
(This is also part of the fix for #820 and I will include it in PR #825 I think)
Can you please review?
Thanks
Andrea

### DIFF
--- a/epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/model_handling.py
+++ b/epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/model_handling.py
@@ -1629,22 +1629,27 @@ class PLUGIN_GPUFOHelasCallWriter(helas_call_writers.GPUFOHelasCallWriter):
                 param = False
             if param:
                 alias = self.params2order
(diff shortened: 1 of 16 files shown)


oliviermattelaer (approved):
see above.

I let you decide, in any case this is good to go!

valassi:
> Yes this is perfect. I guess replacing the "==" by "is" would also have done the trick. But while the above would have avoid to define one variable, this one has the advantages of clarity.

Thanks Olivier! Interesting, I had not thought of "is"... But yes, let;s keep the more verbose and more clear version.

> The only thing that one might consider if is the line ###if alias == self.couporderindep: # bug #821! this is incorrectly true when all all dictionaries are empty! are needed or not, (they certainly do not hurt). But I would advocate to at least modify them to ###if alias == self.couporderindep: # bug #821! this is incorrectly true when both dictionaries are empty!

Yes I have done this.

valassi:
> see above.
> 
> I let you decide, in any case this is good to go!

Ok I will merge this when the CI is done, just to get one out of the way. I will then update the other PRs.

Issue #627, opened 2023-04-03
SUSY HRDCOD=1 builds fail because sin/cos/atan are not constexpr


valassi:
SUSY HRDCOD=1 builds fail because sin/cos/atan are not constexpr

See for instance susy gg to tt in PR #625, and in particular https://github.com/madgraph5/madgraph4gpu/pull/625/commits/586195dced571514e6f5f903c2e4b4b7fb7fe9af

The log says
```
The HRDCOD=1 build now fails with

ccache /usr/local/cuda-12.0/bin/nvcc  -O3  -lineinfo -I. -I../../src -I../../../../../tools -I/usr/local/cuda-12.0/include/ -DUSE_NVTX -gencode arch=compute_70,code=compute_70 -gencode arch=compute_70,code=sm_70 -use_fast_math -std=c++17  -ccbin /usr/lib64/ccache/g++ -DMGONGPU_FPTYPE_DOUBLE -DMGONGPU_FPTYPE2_DOUBLE -DMGONGPU_HARDCODE_PARAM -Xcompiler -fPIC -c gcheck_sa.cu -o gcheck_sa.o
../../src/Parameters_MSSM_SLHA2.h(310): error: expression must have a constant value
../../src/Parameters_MSSM_SLHA2.h(310): note #2703-D: cannot call non-constexpr function "atan(double) noexcept(true)"
/usr/local/cuda-12.0/include/crt/math_functions.h(4137): here
../../src/Parameters_MSSM_SLHA2.h(725): error: expression must have a constant value
../../src/Parameters_MSSM_SLHA2.h(725): note #2703-D: cannot call non-constexpr function "cos(double) noexcept(true)"
/usr/local/cuda-12.0/include/crt/math_functions.h(553): here
../../src/Parameters_MSSM_SLHA2.h(726): error: expression must have a constant value
../../src/Parameters_MSSM_SLHA2.h(726): note #2703-D: cannot call non-constexpr function "sin(double) noexcept(true)"
/usr/local/cuda-12.0/include/crt/math_functions.h(520): here
3 errors detected in the compilation of "gcheck_sa.cu".
```

valassi:
Update Feb 2024. Actually, it is only nvcc that complains that sin/cos/atan are not constexpr. The C++ only build (CUDA_HOME=none using the old setup) succeeds, and the runtime tests are ok.

Maybe we can try to hardcode the gcc implementation of constexpr sin/cos/atan into our nvcc code?

[...]

PR #632, opened 2023-04-06
WIP: start investigating builds of smeft gg_tttt

valassi:
This is a WIP MR including the process SMEFT gg_tttt relevant for @zeniheisser .

The builds fail both in HRDCOD=0 and HRDCOD=1

PR #824, opened 2024-03-08
comprehensive set of fixes for SUSY (complete) and EFT (partial) - keep susy_gg_tt, susy_gg_t1t1, heft_gg_h in repo and test them in CI

### LINKED ISSUE #818: SUSY HRDCOD=0 builds in CUDA give MEs equal to 0 (warning in parameter access through getInstance in device code)
"SUSY HRDCOD=0 builds in CUDA give MEs equal to 0 (warning in parameter access through getInstance in device code)"

The title says it all. See https://github.com/madgraph5/madgraph4gpu/pull/625#issuecomment-1943265467

I suspect that we msut work around getInstance - or maybe, more likely, copy parameters like mdl_I51x1 from host to device so that they can be used in the alphas running computations there on the device.

### LINKED ISSUE #821: Bug in GPUFOHelasCallWriter format_coupling (wrong params2order for SUSY processes)

### LINKED ISSUE #823: Out-of-bounds array warning in SUSY gq_ttllq
I have essentially fixed all of the SUSY issues in #820 and am now systematically testing a large spectrum of SUSY processes. Since this is one process that I had used long ago, I am testing gq_ttllq. The build succeeds and the tests seem to succeed. however in one of the four subprocesses I get this warning, which looks worrying
```
/data/avalassi/GPU2023/madgraph4gpuBis/epochX/cudacpp/susy_gq_ttllq.sa/SubProcesses/P1_Sigma_MSSM_SLHA2_gu_ttxemepu> make cleanall; make HRDCOD=0 -j
...
ccache g++  -O3  -std=c++17 -I. -I../../src -Wall -Wshadow -Wextra -ffast-math  -fopenmp  -march=skylake-avx512 -mprefer-vector-width=256  -DMGONGPU_FPTYPE_DOUBLE -DMGONGPU_FPTYPE2_DOUBLE -fPIC -c CPPProcess.cc -o CPPProcess.o
In file included from CPPProcess.cc:22:
MemoryAccessCouplingsFixed.h: In function ‘void mg5amcCpu::calculate_wavefunctions(int, const fptype*, const fptype*, mgOnGpu::fptype*, mg5amcCpu::fptype_sv*, int)’:
MemoryAccessCouplingsFixed.h:41:44: warning: array subscript 12 is outside array bounds of ‘mgOnGpu::fptype [10]’ {aka ‘double [10]’} [-Warray-bounds]
   41 |       return &( buffer[iicoup * nx2 + ix2] ); // STRUCT[idcoup][ix2]
      |                                            ^
CPPProcess.cc:91:17: note: while referencing ‘mg5amcCpu::cIPC’
   91 |   static fptype cIPC[10];
      |                 ^~~~
In file included from CPPProcess.cc:22:
MemoryAccessCouplingsFixed.h:41:44: warning: array subscript 12 is outside array bounds of ‘mgOnGpu::fptype [10]’ {aka ‘double [10]’} [-Warray-bounds]
   41 |       return &( buffer[iicoup * nx2 + ix2] ); // STRUCT[idcoup][ix2]
      |                                            ^
CPPProcess.cc:91:17: note: while referencing ‘mg5amcCpu::cIPC’
   91 |   static fptype cIPC[10];
      |                 ^~~~
```
This is using gcc11.3. Also gcc12.1 gives the same warning. Strangely, clang14 does not complain.

Looking at the code is complex because this is deep inside some templated methods.

However the following looks wrong, even if it is just a comment.

In Param.h:
```
  namespace Parameters_MSSM_SLHA2_dependentCouplings
  {
    constexpr size_t ndcoup = 2; // #couplings that vary event by event because they depend on the running alphas QCD
    constexpr size_t idcoup_GC_51 = 0;
    constexpr size_t idcoup_GC_6 = 1;
    ...
  }
...
  namespace Parameters_MSSM_SLHA2_independentCouplings
  {
    constexpr size_t nicoup = 7; // #couplings that are fixed for all events because they do not depend on the running alphas QCD
    //constexpr size_t ixcoup_GC_1 = 0 + Parameters_MSSM_SLHA2_dependentCouplings::ndcoup; // out of ndcoup+nicoup
    //constexpr size_t ixcoup_GC_2 = 1 + Parameters_MSSM_SLHA2_dependentCouplings::ndcoup; // out of ndcoup+nicoup
    //constexpr size_t ixcoup_GC_3 = 2 + Parameters_MSSM_SLHA2_dependentCouplings::ndcoup; // out of ndcoup+nicoup
    //constexpr size_t ixcoup_GC_246 = 3 + Parameters_MSSM_SLHA2_dependentCouplings::ndcoup; // out of ndcoup+nicoup
    //constexpr size_t ixcoup_GC_248 = 4 + Parameters_MSSM_SLHA2_dependentCouplings::ndcoup; // out of ndcoup+nicoup
    //constexpr size_t ixcoup_GC_249 = 5 + Parameters_MSSM_SLHA2_dependentCouplings::ndcoup; // out of ndcoup+nicoup
    //constexpr size_t ixcoup_GC_250 = 6 + Parameters_MSSM_SLHA2_dependentCouplings::ndcoup; // out of ndcoup+nicoup
  }
```

In Param.cc:
```
void
Parameters_MSSM_SLHA2::setIndependentCouplings()
{
  GC_1 = -( mdl_ee * mdl_complexi ) / 3.;
  GC_2 = ( 2. * mdl_ee * mdl_complexi ) / 3.;
  GC_3 = -( mdl_ee * mdl_complexi );
  GC_246 = -( mdl_cw * mdl_ee * mdl_complexi ) / ( 2. * ( -1. + mdl_sw ) * mdl_sw * ( 1. + mdl_sw ) );
  GC_248 = -( mdl_cw * mdl_ee * mdl_complexi * mdl_sw ) / ( 3. * ( -1. + mdl_sw ) * ( 1. + mdl_sw ) );
  GC_249 = ( 2. * mdl_cw * mdl_ee * mdl_complexi * mdl_sw ) / ( 3. * ( -1. + mdl_sw ) * ( 1. + mdl_sw ) );
  GC_250 = -( ( mdl_cw * mdl_ee * mdl_complexi * mdl_sw ) / ( ( -1. + mdl_sw ) * ( 1. + mdl_sw ) ) );
}
```

In CPPProcess.cc:
```
../../SubProcesses/P1_Sigma_MSSM_SLHA2_gux_ttxemepux/CPPProcess.cc:  __device__ const fptype cIPC[14] = { (fptype)Parameters_MSSM_SLHA2::GC_3.real(), (fptype)Parameters_MSSM_SLHA2::GC_3.imag(), (fptype)Parameters_MSSM_SLHA2::GC_2.real(), (fptype)Parameters_MSSM_SLHA2::GC_2.imag(), (fptype)Parameters_MSSM_SLHA2::GC_246.real(), (fptype)Parameters_MSSM_SLHA2::GC_246.imag(), (fptype)Parameters_MSSM_SLHA2::GC_250.real(), (fptype)Parameters_MSSM_SLHA2::GC_250.imag(), (fptype)Parameters_MSSM_SLHA2::GC_249.real(), (fptype)Parameters_MSSM_SLHA2::GC_249.imag(), (fptype)Parameters_MSSM_SLHA2::GC_1.real(), (fptype)Parameters_MSSM_SLHA2::GC_1.imag(), (fptype)Parameters_MSSM_SLHA2::GC_248.real(), (fptype)Parameters_MSSM_SLHA2::GC_248.imag() };
...
../../SubProcesses/P1_Sigma_MSSM_SLHA2_gu_ttxemepu/CPPProcess.cc:  __device__ const fptype cIPC[10] = { (fptype)Parameters_MSSM_SLHA2::GC_3.real(), (fptype)Parameters_MSSM_SLHA2::GC_3.imag(), (fptype)Parameters_MSSM_SLHA2::GC_2.real(), (fptype)Parameters_MSSM_SLHA2::GC_2.imag(), (fptype)Parameters_MSSM_SLHA2::GC_246.real(), (fptype)Parameters_MSSM_SLHA2::GC_246.imag(), (fptype)Parameters_MSSM_SLHA2::GC_250.real(), (fptype)Parameters_MSSM_SLHA2::GC_250.imag(), (fptype)Parameters_MSSM_SLHA2::GC_249.real(), (fptype)Parameters_MSSM_SLHA2::GC_249.imag() };
```

Rephrasing: **why does cIPC have different dimensions in P1_Sigma_MSSM_SLHA2_gu_ttxemepu (10, gives a warning) and in P1_Sigma_MSSM_SLHA2_gux_ttxemepux (14, no warning)??**

valassi:
Hi @oliviermattelaer I have completed many more fixes for SUSY so I am including them here.

This PR includes also all the changes that are in PR #625 (first set of susy changes) and in PR #822 (fix for a helas bug affecting also non SUSY processes). I filed separate PRs to make it easier (maybe) to review.

But if you prefer, I can also just merge this straight away and close the other two.

Can you please review?

Notes
- This closes #818 aand #627 (through the changes in PR #625)
- This closes #821 (helas bug, through the changes in PR #822)
- It also closes #823 (out of bound arrays bug in cIPC), not in any other PR
- And it especially closes all of the issues in #820, not in any other PR

The status is: **all SUSY processes that I tried are now correctly code generated, built and tested**. Can you maybe also suggest new processes to try?

Thanks
Andrea

cc @roiser


valassi:
Note: in spite of all the SUSY issues this fixes, this still does NOT fix SMEFT builds.

valassi:
Hi @oliviermattelaer I confirm that this is now ready for review, can you please have a look? Let me know if you want to discuss it via zoom or similar.

As mentioned, this includes also PR #625 and PR #822. So easier to only review this one if you prefer.

oliviermattelaer:
thanks @valassi sorry to be slow for the moment, I will try to review this one today  (or maybe the two others depending of this review I guess).

Sorry,

Olivier

valassi:
> thanks @valassi sorry to be slow for the moment, I will try to review this one today (or maybe the two others depending of this review I guess).
> 
> Sorry,
> 
> Olivier

Thanks Olivier. Just to clarify (sorry if confusing): this PR is the most advanced of them all. The other two are earlier ones and less complete. But, apart from one issue, the comments you sent to the simpler #625 still make sense here.

One notable thing that was broken in #625 but is fixed here is that previously only real BSM parameters were handled, now instead here also the complex ones are handled.

By the way this confirms what I said in the other comments: somehow, the BSM couplings are NOT handled by the code that handles the cIPC for SM couplings. They go through a diferent part of teh python code, which was broken and which here I fixed.

valassi:
Hi @oliviermattelaer , as discussed, I have merged #822 and also #625. So now this contains only the additional fixes, which fix SUSY as far as I can tell.

As discussed in #625, before considering this good to merge, I will stil investigate whether I can relax the "if model starts with sm" condition, because this is dangerous and may include things like SM with Z prime, etc.

I also quote your comments from https://github.com/madgraph5/madgraph4gpu/pull/625#issuecomment-1998308440

> This is helpfull, Maybe the only change that need to be done (but not required to be done in this PR) is to run all the BSM trick for the SM model as well and check that SM is still working correctly.
> 
> We can also sit together on Friday to see if I can design a quick and nice way to fix your issue with those coupling. But, I guess that the answer will not be quick...
> 
> One reason of the complexity is that the structure of the code is likely:
> 
>     1. read the UFO model -> convert it to a MG5 object
> 
>     2. potential optimization of the model by MG5
> 
>     3. generate the diagrams
> 
>     4. generate the helas call -> write the process file
> 
>     5. retrieve from step 3 or 4 the list of couplings/lorentz structure needed for the particular process and create code to compute only such list. (Note that all parameters are handle separetly and likely do not have such optimization ...)
> 
> 
> So I guess the confusion might partly comes from the difference between coupling and parameter (and the difference is in a way arbitrary) and the fact that some parameter can/should be real (coupling should always be treated as complex --even if they might sometimes/often be real or pure imaginary--).

I will not address these points yet. I will just try to remove the 'SM' condition

valassi:
> This is helpfull, Maybe the only change that need to be done (but not required to be done in this PR) is to run all the BSM trick for the SM model as well and check that SM is still working correctly.

Hi @oliviermattelaer again, this is now done. Thanks for the good suggestion! The code is now cleaner.

En passant, I also identified and fixed a couple more minor issues for EFT. In particular, heft_gg_h now builds and runs ok also in HRDCOD=0 mode (it was not before). The issue here was that I was computing my list of "BSM" independent parameters that must go to device constant memory only looking at those needed to compute dependent _couplings_, but there are also computations of dependent _parameters_, and this is now fixed.

Unfortunately the SMEFT processes still have other issues #614 #616 #632, but it looks better anyway.

By the way I added back susy_gg_tt to the repo. There are now both susy_gg_tt (needed to test double BSM parameters, 1 value each) and susy_gg_t1t1 (needed to test complex BSM parameters, 2 values each). And I added these two processes and also heft_gg_h to the CI tests.

This is again ready for review, I remove the draft status. Let me know if you have comments.

I will still run a few tests though before merging. Thanks!

cc @roiser

valassi:
By the way @oliviermattelaer have you got a better HEFT suggestion that gg>h? This process works but has no phase space as there is no degree of freedom. Maybe something like HEFT gg>bb (with massive bs) would be better? Thanks

valassi:
Ok I have completed my tests. 
@oliviermattelaer I think this is ready to be merged when it is reviewed.

Note, I have also implemented the "tmad" tests comparing fortran to cuda/cpp, an dthey fail for both susy_gg_tt #825 and for susy_gg_t1t1 #826. But I suggest meging this one already (which fixes many other things!) and then addressing those other issues later.

By the way, again, eventually I'd like to remove heft_gg_h by something more interesting, since heft seems fixed (unlike smeft).

epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/model_handling.py
```diff
@@ -823,15 +823,30 @@ def super_generate_parameters_class_files(self):
...
-                            bsmparam_indep_real_used.append( param.name )
-            bsmparam_indep_real_used = set( bsmparam_indep_real_used ) 
+        bsmparam_indep_complex_used = []
+        for param in self.params_indep: # NB this is now done also for 'sm' processes (no check on model name, see PR #824)
+            if param.name == 'mdl_complexi' : continue
+            if param.name == 'aS' : continue
+            # Add BSM parameters which are needed to compute dependent couplings
+            # Note: this seemed enough to fix SUSY processes, but not EFT processes
+            for coupdep in self.coups_dep.values():
+                if param.name in coupdep.expr:
```
oliviermattelaer, line 834: things to check, do we need regular expression here to avoid overlap
but I guess 
mdl_g
will be return true for 
mdl_g1 as expression

But likely not critical since just too much
-> Andrea answer (thinks this is include in a next PR -> I will check)

epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/model_handling.py
```diff
@@ -1165,54 +1162,49 @@ def get_process_function_definitions(self, write=True):
...
-        if self.model_name[:2] == 'sm' :
-            replace_dict['bsmMemcpySym'] = ''
-            replace_dict['bsmMemcpy'] = ''
-            replace_dict['bsmdump'] = ''
-        else:
-            replace_dict['bsmMemcpySym'] = '\n    if( Parameters_MSSM_SLHA2::nBsmIndepParam > 0 )\n      gpuMemcpyToSymbol( bsmIndepParam, m_pars->mdl_bsmIndepParam, Parameters_MSSM_SLHA2::nBsmIndepParam * sizeof( double ) );'
-            replace_dict['bsmMemcpy'] = '\n    if( Parameters_MSSM_SLHA2::nBsmIndepParam > 0 )\n      memcpy( bsmIndepParam, m_pars->mdl_bsmIndepParam, Parameters_MSSM_SLHA2::nBsmIndepParam * sizeof( double ) );'
-            replace_dict['bsmdump'] = '\n    //for ( int i=0; i<Parameters_MSSM_SLHA2::nBsmIndepParam; i++ ) std::cout << std::setprecision(17) << "m_pars->mdl_bsmIndepParam[i] = " << m_pars->mdl_bsmIndepParam[i] << std::endl;'
+            replace_dict['cipdhrdcod'] = '//__device__ const fptype* cIPD = nullptr; // unused as nIPD=0'
+        # FIXME! Here there should be different code generated depending on MGONGPUCPP_NBSMINDEPPARAM_GT_0...
```
oliviermattelaer, line 1207: Need to check the status here for @valassi

oliviermattelaer (approved):
Sounds good some minor point to check -> let's say approve!!!! 
Thank you

epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/model_handling.py
```diff
@@ -823,15 +823,30 @@ def super_generate_parameters_class_files(self):
...
-            for param in self.params_indep:
-                if param.type == 'real':
-                    for coup in self.coups_dep.values():                
-                        if param.name in coup.expr:
-                            bsmparam_indep_real_used.append( param.name )
-            bsmparam_indep_real_used = set( bsmparam_indep_real_used ) 
+        bsmparam_indep_complex_used = []
+        for param in self.params_indep: # NB this is now done also for 'sm' processes (no check on model name, see PR #824)
+            if param.name == 'mdl_complexi' : continue
+            if param.name == 'aS' : continue
```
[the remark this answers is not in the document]
valassi: Thanks @oliviermattelaer . As I imagined, we already have since a long time a treatment of 'ZERO':
https://github.com/madgraph5/madgraph4gpu/blob/df29acb070abeb219b65facd8b2b451b27f36e5e/epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/model_handling.py#L1710

So nothing changed in this MR about this. This check is in another part of the code, and I replace 'ZERO' with a hardcoded '0'. Let me know if you see an issue with this.

epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/model_handling.py
valassi: Thanks @oliviermattelaer 

I tried to use ' '+param.name+' ' instead, but this is not enough. See cfd56a4ff in PR #832. 
I understand this is not ideal, maybe something to be addressed in mg5amcnlo#90?
Not much else I can do here I think I am afraid

epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/model_handling.py
valassi: This comment is essentially about open issue #827. I have mentioned the issue in the code (in a patch in #832). It is still open.

epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/model_handling.py
valassi: I have also added a comment in a patch in #832

valassi:
> I will merge to ease the comparison to smeft and then heft.
> 
> Keep here the three comments for later:
> 
>     * what about param.name == 'ZERO' ?
> 
>     * mdl_g will be return true for mdl_g1 as expression?
> 
>     * check the status of "FIXME! Here there should be different code generated depending on MGONGPUCPP_NBSMINDEPPARAM_GT_0"?

Ok I have added comments on all three points (previously resolved without answers). All done for me.

PR #834, opened 2024-04-08
Default fptype

oliviermattelaer:
change default fptype to "m"

Note that the value from the run_card is written inside the make_opts file for the moment.
I can write it in another file if needed by the refactorization of the matrix-element but for the current master branch this is at least not needed (I can do it in the associated branch if needed)


valassi:
Hi @oliviermattelaer I am a bit confused by this one. You asked for a review, but then I see you closed this, so I guess that there is nothing you expect me to do? Let me know otherwise. Thanks

Issue #700, opened 2023-06-13
Add support for AVX, HRDCOD, HELINL, FPTYPE in the runcards


valassi:
Add support for AVX, HRDCOD, HELINL, FPTYPE in the runcards

valassi:
fptype was done by Olivier in #788

PR #835, opened 2024-04-09
change default for fptype

oliviermattelaer:
change default fptype to "m" (second try)

Note that the value from the run_card is written inside the make_opts file for the moment.
I can write it in another file if needed by the refactorization of the matrix-element but for the current master branch this is at least not needed (I can do it in the associated branch if needed)


valassi (approved):
Hi @oliviermattelaer this is essentially good for me, thanks.

One point: in PR #798, the avx_level and cudacpp_backend are merged: the AVX is included in the backend (because cpp by itself does not exist, it is always either cppnone, cppsse4, etc: and cpp defaults to cppauto i.e. let MG choose).

Another point: I would suggest to add aslo helinl and hrdcod here, see #700. And when we do that, I would be much more consistent in the naming, and name all these cudacppp options with a prefix: so in teh runcard we would have cudacpp_backend, cudacpp_fptype, cudacpp_helinl, cudacpp_hrdcod. These four variables only make sense for the cudacpp plugin.

Anyway, feel free to merge and we discuss these points another time. Thanks!

Issue #844, opened 2024-05-15
Remove gen_ximprove.py and madevent_interface.py from patch.common?


valassi:
Hi @oliviermattelaer this is a followup to your comments in PR #798:
https://github.com/madgraph5/madgraph4gpu/pull/798#discussion_r1585038141
https://github.com/madgraph5/madgraph4gpu/pull/798#discussion_r1585040722

Essentially, you suggest that patch.common should not touch madevent_interface.py, do I understand this correctly?

Now, this issue is *already in upstream/master*. My changes in PR #798 modify the patch, but for the moment I have no way of doing that differently. *I suggest to strip that off PR #798 and solve it in a separate PR for this new issue*.

Note also, this is not strictly a new issue: you already mentioned this in #756 and specifically https://github.com/madgraph5/madgraph4gpu/issues/756#issuecomment-1699237967
```
python file modify by patch (create issue with launch)
- [x] banner.py (no need)
   - [x] add parameter in the run_card (fixed via generic method)  
- [ ] gen_ximprove.py
   - [ ]  change compile code
- [ ] madevent_interface.py
   - [ ] change compile code 
```

By the way, I suspect that these changes were introduced here (cc @roiser who may know more)
```
commit 2a188c04d945726b5f2ac071a750360e3ec303c3
Author: Stefan Roiser <stefan.roiser@cern.ch>
Date:   Fri Mar 31 18:07:55 2023 +0200

    add compile for individual targets in the refine step

diff --git a/epochX/cudacpp/gg_tt.mad/bin/internal/madevent_interface.py b/epochX/cudacpp/gg_tt.mad/bin/internal/madevent_interface.py
index 9a308cc60..50392be67 100755
--- a/epochX/cudacpp/gg_tt.mad/bin/internal/madevent_interface.py
+++ b/epochX/cudacpp/gg_tt.mad/bin/internal/madevent_interface.py
@@ -3615,8 +3615,20 @@ Beware that this can be dangerous for local multicore runs.""")
                 logger.info('    %s ' % subdir)
     
                 if os.path.exists(pjoin(Pdir, 'ajob1')):
-                    self.compile(['madevent'], cwd=Pdir)
-                    
+
+                    exec_mode = 0
+                    if 'exec_mode' in self.run_card:
+                        exec_mode = self.run_card['exec_mode']
+
+                    if exec_mode == 0:
+                        self.compile(['madeventfortran'], cwd=Pdir)
+                    elif exec_mode == 1:
[...]

PR #849, opened 2024-05-23
remove un-necessary patch

oliviermattelaer:
The point is that the function
self.compile('madevent')
is already catched by the CPP/CUDA interface of launch_plugin.py and does ensure that 
compile('madevent') actually compiles the required exectuble. So they are no need to change the local copy of the  "upstream" code (which might not be used anyway since the upstream one can have priority).

So since this is already handle correctly where you run the unpatch file, they are no reason to patch them...
So I did remove the patch in this PR. (see also [#844](https://github.com/madgraph5/madgraph4gpu/issues/844))


oliviermattelaer:
This CI crash on a quite funy way:

```
 ../../src/Parameters_MSSM_SLHA2.h:826:58: error: conversion from ‘mg5amcCpu::fptype_sv’ {aka ‘__vector(2) double’} to non-scalar type ‘const mgOnGpu::cxsmpl<double>’ requested
  826 |         constexpr cxsmpl<double> mdl_G__exp__2 = ( ( G ) * ( G ) );
      |                                                  ~~~~~~~~^~~~~~~~~
make[2]: *** [cudacpp.mk:700: build.sse4_d_inl0_hrd0/runTest_cpp.o] Error 1
```
So this sounds like not related to this PR but to (another) type missmatch?

What do you think @valassi?

oliviermattelaer:
@valassi, just to remind us about this PR here, would you have any conceptual problem with this?
if not, I can probably update the mg5amc commit to avoid the conflict and repropose this.

valassi (approved):
Hi @oliviermattelaer that's good for me.

Or better: I trust you on this, as it only touches gen_ximprove.py and madevent_interface.py, and I am not routinely testing those. So you certainly know better.

Only two issues:
- the MG5AMC conflict must be fixed
- the processes must be regenerated

Both issues are solved in my PR #939. So you could merge that branch of mine here and then merge into master, or merge directly my 939.

Thanks
Andrea

valassi:
> Only two issues:
> 
>     * the MG5AMC conflict must be fixed
> 
>     * the processes must be regenerated
> 
> 
> Both issues are solved in my PR #939. So you could merge that branch of mine here and then merge into master, or merge directly my 939.

Thanks Olivier for approving my #939 :-)

I had forgotten the conflicts. I thought I would merge this 849 into master first, and then 939 into master, but this will not work. I will instead merge 939 here, and then this 849 into master.

valassi:
Ok I merged 939 here into less_patch, fixing the conflicts.

Waiting for the CI to complete, then I will merge this into master.

valassi:
Ok CI completed with 6 failures as expected (susy and pptt012j), merging

image

Issue #826, opened 2024-03-17
No cross section in SUSY gg_t1t1 log file


valassi:
In PR https://github.com/madgraph5/madgraph4gpu/pull/824 I fixed SUSY codegen, builds and internal tests cuda/cpp.
But now I ALSO added the test comparing to fortran and this fails

In tmad/logs_susyggt1t1_mad/log_susyggt1t1_mad_d_inl0_hrd0.txt
```
*** (2-none) EXECUTE MADEVENT_CPP x1 (create events.lhe) ***
--------------------
CUDACPP_RUNTIME_FBRIDGEMODE = (not set)
CUDACPP_RUNTIME_VECSIZEUSED = 8192
--------------------
8192 1 1 ! Number of events and max and min iterations
0.000001 ! Accuracy (ignored because max iterations = min iterations)
0 ! Grid Adjustment 0=none, 2=adjust (NB if = 0, ftn26 will still be used if present)
1 ! Suppress Amplitude 1=yes (i.e. use MadEvent single-diagram enhancement)
0 ! Helicity Sum/event 0=exact
1 ! Channel number (1-N) for single-diagram enhancement multi-channel (NB used even if suppress amplitude is 0!)
--------------------
Executing ' ./build.none_d_inl0_hrd0/madevent_cpp < /tmp/avalassi/input_susyggt1t1_x1_cudacpp > /tmp/avalassi/outpu
t_susyggt1t1_x1_cudacpp'
 [OPENMPTH] omp_get_max_threads/nproc = 1/4
 [NGOODHEL] ngoodhel/ncomb = 4/4
 [XSECTION] VECSIZE_USED = 8192
 [XSECTION] MultiChannel = TRUE
 [XSECTION] Configuration = 1
 [XSECTION] ChannelId = 2
 [XSECTION] ERROR! No cross section in log file:
   /tmp/avalassi/output_susyggt1t1_x1_cudacpp
   ...
xqcutij # 3>     0.0     0.0
 RESET CUMULATIVE VARIABLE
 NGOODHEL =           4
 NCOMB =           4
 MULTI_CHANNEL = TRUE
 CHANNEL_ID =           2
 RESET CUMULATIVE VARIABLE
        4096  points passed the cut but all returned zero
 therefore considering this contribution as zero
 Deleting file events.lhe
```

oliviermattelaer:
After PR #850, the code does provide a cross-section (even if the statement might be machine specific).
But in any case the cross-section does not match
[...]

PR #850, opened 2024-05-30
Fix 825

### LINKED ISSUE #825: Cross section mismatch (fortran/cpp) in SUSY gg_tt
Cross section mismatch (fortran/cpp) in SUSY gg_tt

In PR #824 I fixed SUSY codegen, builds and internal tests cuda/cpp.
But now I ALSO added the test comparing to fortran and this fails

In tmad/logs_susyggtt_mad/log_susyggtt_mad_d_inl0_hrd0.txt
```
ERROR! xsec from fortran (44.598860065419856) and cpp (171.81273026311101) differ by more than 3E-14 (2.85240183294
1188)
```

oliviermattelaer:
This does fix #825 where the problem was that the unary handling was not propagated within the aloha routine.
This patch include such handling now but I'm a bit worried about other processes that can be impacted.
This PR (via the CI) is a simple way to test many processes to spot potential issue, if it goes trough, I will anyway have to check the corner case, I'm worried about...

### DIFF
--- a/epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/model_handling.py
+++ b/epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/model_handling.py
@@ -482,7 +482,7 @@ class PLUGIN_ALOHAWriter(aloha_writers.ALOHAWriterForGPU):
                     if self.routine.denominator:
                         raise Exception('modify denominator are not compatible with complex mass scheme')
                     # This affects 'denom = COUP' in HelAmps_sm.cc
-                    out.write('    %(declnamedenom)s =  %(pre_coup)s%(coup)s%(post_coup)s / ( ( P%(i)s[0] * P%(i)s[0] ) - ( P%(i)s[1] *P%(i)s[1] ) - ( P%(i)s[2] * P%(i)s[2] ) - ( P%(i)s[3] * P%(i)s[3] ) - ( M%(i)s * M%(i)s ) );\n' % mydict) # AV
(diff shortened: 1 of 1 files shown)


oliviermattelaer:
@roiser, ok this is good to be squashed and merged.

Olivier

valassi:
Hi @oliviermattelaer I had a look since I saw that you posted this.

I confirm that #825 seems fixed: the cross section from fortran and cuda/c++ now matches in susy_gg_tt. 

Conversely for #826 I still see that there is NO cross section for susy_gg_t1t1. So I am a bit surprised when you say that you do see a cross section. Maybe it depends on the number of events and the exact setup? Anyway, I agree with you that this needs more investigation.

One point, I see that this patch also changes the Parameters.h files in other processes including eemumu. Is this expected? Anyway, I will run a few tests on all processes and see what this gives.

Just in case, this is in https://github.com/valassi/madgraph4gpu/tree/susy (but I will not make a PR out of this, at least not yet)

Thanks! Andrea

oliviermattelaer:
Hi Andrea,

in Parameters.h  this sounds surprising (I do not see any of those files changed in your repo actually)
But in all process/model, the file HelAmps_sm.h is going to be modified since the issue has nothing specific to the MSSM model and we could --maybe we should--  have detected in the SM has well.

For #826, I will "first" try to fix the missmatch that I do observe on my laptop, and then we might need to work together to understand why your machine is just crashing (if it is still happens at that stage obviously).

valassi:
> Hi Andrea,
> 
> in Parameters.h this sounds surprising 

OPS! SORRY! Yes I meant that HelAmps.h changed, not Paramaters.h!

valassi:
For me this is good to go and should be merged. However it will then also need an update of refence files for runTest (#859), which is in my PR #860. So I suggest, 
- either merge this #850 first and then my #860
- or merge directly my #860 which includes this #850
The result should be the same. Thanks

valassi:
I merged my #860. As I expected, github automatically closed this one as merged too. Thanks Olivier!

PR #851, opened 2024-05-30
Add bin/madevent to the madgraph4gpu repo for all processes

valassi:
Hi @oliviermattelaer this is a very simple PR that __only__ affecs madgraph4gpu repos. It does not touch CODEGEN.

It adds bin/madevent to the madgraph4gpu repo for all processes, in particular BSM processes where it was missing (while for SM processes it was there). I had to modify the .gitignore (again __only__ of madgraph4gpu, no change in CODEGEN).

The reason is that I need bin/madevent to launch bin/generate_events from a repo process when I modify the runcards. (The treacards step uses it).

I hope it is not controversial, can you approve please? I need this to investigate the cross section crashes in susy #826 

Thanks
Andrea

### DIFF
--- a/epochX/cudacpp/.gitignore
+++ b/epochX/cudacpp/.gitignore
@@ -4,17 +4,5 @@ results.dat*
 results.pkl
 run_[0-9]*
 events.lhe*
-.libs
-.cudacpplibs
-madevent
-madevent_fortran
-madevent_cpp
-madevent_cuda
-
-G[0-9]*
-ajob[0-9]*
-input_app.txt
-symfact.dat
-gensym
 
 py3_model.pkl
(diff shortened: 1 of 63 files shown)


valassi:
PS Ah yes this includes many changes also because it is the first full regeneartion with your PR 835: so launch_plugin.py and runcards are modified from your patch Olivier

valassi:
PPS and just for further explanation, I removed madevent and a few other files from epochX/cudacpp/.gitignore because it was/is ALREADY in the  .gitignore of all SubProcesses/P* generated directories

oliviermattelaer (approved):
Sure go ahead.

Thanks

Olivier

valassi:
Ok thanks! I will merge this. Andrea

PR #854, opened 2024-06-01
Add matrix.pdf files to the git repo

valassi:
Hi @oliviermattelaer this is another small PR that should be non controversial. I am just adding matrix.pdf files instead of matrix.ps files to the madgraph4gpu git repo, to make it easier to debug diagram related issues like #826.

THis does not affect CODEGEN in standard user mode. Can you please approve?

Thanks
Andrea


valassi:
I have now updated this by removing pdf and xmp metadata from the matrix.pdf file to make it reproducible.

This PR is now ready for approval and merging

valassi:
I have updated this after merging #860 into master.

@oliviermattelaer this is ready to be reviewed and merged. Thanks

oliviermattelaer (approved):
I'm not a fan of commiting binary format in a git repo, (see xz recent issue).
But if that helps you, lets go for it! 
But why not keeping both the ps and pdf? This will make more clear what is change in the pdf file if we see the change from git in the ps file (since they are text file)

valassi:
> I'm not a fan of commiting binary format in a git repo, (see xz recent issue).
> But if that helps you, lets go for it!

Thanks Olivier :-)

(What was the xz recent issue?)

> But why not keeping both the ps and pdf? This will make more clear what is change in the pdf file if we see the change from git in the ps file (since they are text file)

Yes I have done this, good suggestion. And then if at some point we feel we must remove pdfs, we can remove them...

I will merge now, thanks Andrea

valassi:
CI tests succeded, merging

Issue #859, opened 2024-06-03
runTest fails in susy_gg_tt and susy_gg_t1t1 after fixing issue 825 in PR 850


valassi:
runTest fails in susy_gg_tt and susy_gg_t1t1 after fixing issue #825 in PR #850

I think that most likely the expected results of runTest were wrong.... they must be updated with the new code which includes Ccoeff in PR #850

valassi:
This is fixed in two commits of PR #860. Closing this as fixed in that PR (which will soon be ready to be merged).

PR #857, opened 2024-06-02
Fix SIGFPE crash (855) in rotxxx by adding volatile in aloha_functions.f

valassi:
Hi @oliviermattelaer I think this is ready for review, can you please have a look?

This fixes the SIGFPE crash in rotxxx (#855, also mentioned in #826 - and maybe this is also related to what you discuss as a crash in #852?). It uses an ugly workaround of just adding 'volatile' to disable optimizations (from SIMD?) deep inside the fortran code. Ugly, but it works, and has no performance penalties I think here. And this is also exactly the same technique I used to fix many SIGFPEs in cudacpp in the ixx/oxx functions last here. 

It also slightly modifies my test scripts to enable testing a different config which is not 1. I do not understand why the SIGFPE crash only seems to happen with some specific iconfig (which may also have other issues), but it seems wuite clear that the issues are independent. Rephrasing: I would add volatile in any case, and then go on and investigate channel/iconfig mappings as in your #853.

Note, this IMO brings us one step of progress, but other issues remain open
- the susy_gg_t1t1 still has no cross section (#826)
- after fixing the SIGFPE crash, iconfig 104 in SM gg_ttgg gives an LHE color mismatch (#856)
As mentioned above, I would go ahead and add volatile, and then investigate the rest.

I will still run large scale tests tonight, but the code will almost certainly not change. So this is ready for review. Thanks.
Andrea

Issue #855, opened 2024-06-02
tmad test crashes in rotxxx (SIGFPE erroneous arithmetic operation)


valassi:
"tmad test crashes for some iconfig (channel/iconfig mapping issues and SIGFPE erroneous arithmetic operation)"

Hi @oliviermattelaer this is a follow up to the discussions in #826 and PR #853.

I prefer to open this as a clean issue and investigate this independently of SUSY, or in any case of zero cross section #826.

In these discussions from your patch #853 I realised that we risk having a MAJOR problem not only for BSM but also for SM, namely: all of my 'tmad' tests test only iconfig=1. These were ok so far (in some cases by luck maybe), but for different iconfig (i.e. if we put a number different from 1 in the input_app.txt piped to madevent.

Indeed I found a crash on the first test I executed, ggttgg with iconfig=104.
```
 ./tmad/madX.sh -ggttgg -iconfig 104
...
On itscrd90.cern.ch [CPU: Intel(R) Xeon(R) Silver 4216 CPU] [GPU: 1x Tesla V100S-PCIE-32GB]:
Working directory (run): /data/avalassi/GPU2023/madgraph4gpuX/epochX/cudacpp/gg_ttgg.mad/SubProcesses/P1_gg_ttxgg

*** (1) EXECUTE MADEVENT_FORTRAN (create results.dat) ***
 [OPENMPTH] omp_get_max_threads/nproc = 1/4
 [NGOODHEL] ngoodhel/ncomb = 64/64
 [XSECTION] VECSIZE_USED = 8192
 [XSECTION] MultiChannel = TRUE
 [XSECTION] Configuration = 104
 [XSECTION] ChannelId = 112
 [XSECTION] Cross section = 0.4632 [0.46320556621222242] fbridge_mode=0
 [UNWEIGHT] Wrote 11 events (found 187 events)
 [COUNTERS] PROGRAM TOTAL          :    4.4430s
 [COUNTERS] Fortran Overhead ( 0 ) :    0.2478s
 [COUNTERS] Fortran MEs      ( 1 ) :    4.1953s for     8192 events => throughput is 1.95E+03 events/s

*** (1) EXECUTE MADEVENT_FORTRAN x1 (create events.lhe) ***
 [OPENMPTH] omp_get_max_threads/nproc = 1/4
 [NGOODHEL] ngoodhel/ncomb = 64/64
 [XSECTION] VECSIZE_USED = 8192
[...]

PR #853, opened 2024-05-31
WIP: susy tests for issue 826

valassi:
This is a WIP PR with various tests for issue #826. This includes in particular Olivier's PR #850 and #852.

Presently not meant for merging... issue #826 is not fixed.

PR #860, opened 2024-06-03
Test and complete the fix for 825 in PR 850

valassi:
Hi @oliviermattelaer this is a PR where I test and complete your fix in PR #850 for #825 (xsec mismatch in susy_gg_tt).

Your patch is good and we should merge it. In these tests however I found that a few additions are needed, in particular the reference files for runTest.exe must be updated (the code changed after adding Ccoeff, so I would say that the old reference files were wrong, see #859).

Still in WIP for now but will be ready soon.


valassi:
Note: after fixing #825 with PR #850, again a SIGFPE reappeared also in gqttq. This is what I had reported in #845. I think that most likely this is the same SIGFPE crash in rotxxx #855 and can be fixed by adding volatile in PR #857... but I will check that.

valassi:
Hi Olivier, this is now also ready for merging.It includes your #850 and complements it with an update of the susy reference files for runTest (#859), plus various other tests. Can you please have a look and approve?

I would suggest to first merge your #850 and then this one. Or I think that one can merge this directly and it will automatically close your #850 as merged.

Reminder, there are two other PRs from me ready to be reviewed and merged: PR #854 (just adding matrix.pdf files for easier visualization than ps) and PR #857 (fixing the SIGFPE crash in rotxxx by adding volatile).

Instead, I think that issues #826 (no xsec in susy_gg_t1t1) and #845 (SIGFPE crash in gqttq) are still pedning to be debugged. I think that a combination of your PR #852 and my PR #853 might fix those, but as far as I am concerned this will wait for end June.

Thanks!
Andrea

oliviermattelaer (approved):
Thanks Andrea, yes this can be merged.
(and no need to merge my PR first)

Cheers,

Olivier

valassi:
Thanks Olivier, merging now

PR #863, opened 2024-06-05
Fixing the segmentation fault detected on the haswell machine

oliviermattelaer:
This is a copy of the PR #852 but targetting a different master such that we can move forward waiting for Andrea.

This contains:
1) An array (that acts a dictionary) to know which index to use in the array linking which color are valid.
2) This includes a bunch of comment to explain the meaning of all variables 
3) This follows relatively closely (but not perfectly) the convention of Andrea

So to my point of view this can be merged swiftly.

Cheers,

Olivier


roiser (approved):
what we discussed yesterday, thanks

PR #830, opened 2024-03-26
channelIDs

roiser:
Hi @oliviermattelaer this is the PR for the channelIds, this goes together with the mg5amcnlo:gpucpp_wrap branch. Functionality wise this should be it now. I still mark it as WIP as not all processes have been re-generated and also the comparison of the generated code fails because of some missing spaces here and there. cc @valassi

PR #864, opened 2024-06-07
Madgraph testsuite

oliviermattelaer:
Adding a CI that allow to test the code from the "madgraph interface"
this use the madgraph standard test suite framework


oliviermattelaer:
Let me close this one and then allow ourself to make the merge from #865
and then it will go forward when #830 will go trough.

Issue #758, opened 2023-08-31
Why do we need to compile with openmp flag?


oliviermattelaer:
Should we remove the openmp flag completely?
or have a way to compile with it on request?

I had to change a makefile to compile on my mac (but will likely create issue with the CI)

valassi:
Hi @oliviermattelaer good point.
(Sorry for the delay in replying).

I have implemented a simple mechanism whereby you can disable OpenMP if you externally set OMPFLAGS= as an empoty string

See https://github.com/madgraph5/madgraph4gpu/pull/874/commits/cb67911ccca86144273015c398bd5b1f43367f20

Conversely, if OMPFLAGS is set externally to something else, for the moment I prefer to ignore this and keep control.

oliviermattelaer:
Perfect thanks,

I think that we should have an agreement on how we handle OpenMP by default from the madgraph5 interface.
 (i.e. should I set OMPFLAGS= by default or not, should we have an entry in the run_card to control that /...) 

Cheers,

Olivier

valassi:
> Perfect thanks,
> 
> I think that we should have an agreement on how we handle OpenMP by default from the madgraph5 interface. (i.e. should I set OMPFLAGS= by default or not, should we have an entry in the run_card to control that /...)
> 
> Cheers,
> 
> Olivier

Thanks a lot Olivier! As discussed in #874 (that I just merged), I reopen this so that we can discuss waht to do... assign it also to you so you remember to provide more suggestions

oliviermattelaer:
I wanted to know yours and @roiser opinion

On my side, I would naively move for OMP OFF by default.
For SIMD case, they are no gain to use OMP.
For the GPU case, this is more blurr but the OMP but I also guess that having multi-process is better than OpenMP (helping to overlap cpu/gpu compuation between the thread)

roiser:
Yes, having multiple processes hammering the GPU is better, but I'm not sure its really needed for the real world use case where we can control how many processes we want to run via the mg5_configuration.txt with nb_core. This is also what I partly (re)exercised when doing the A100/H100 comparison (slides on the indico of the last MG5 dev meeting).

valassi:
[...]

Issue #872, opened 2024-06-27
Cross section mismatch in pp_tt012j (P2_gu_ttxgu) in CI tmad tests - reset_cumulative_variable was called twice in fortran and once in cudacpp


valassi:
There is a cross section mismatch in pp_tt012j in CI tmad tests.

This only appears after fixing the rotxxx crash (which otherwise hides it).
I have seen this in the CI for PR #857, where volatile is added to fix rotxxx.
https://github.com/madgraph5/madgraph4gpu/actions/runs/9694817881/job/26753418805
```
*** (2-none) Compare MADEVENT_CPP xQUICK xsec to MADEVENT_FORTRAN xsec ***

ERROR! xsec from fortran (0.82047507505698292) and cpp (0.67821465194602271) differ by more than 3E-13 (0.1733878669026967)
```

![image](https://github.com/madgraph5/madgraph4gpu/assets/3473550/56aaf894-3e59-4882-97b5-54a5e6f52706)

valassi:
I forgot to mention: the CI now runs tmad tests in each P* subprocess. The issue appears in P2_gu_ttxgu. (Other subprocesses succeed).

valassi:
I am having a look at this.

I confirm that the CI fails in P2_gu_ttxgu. This is a recent run in PR #934 where the pptt012j tests are the only pending issues
https://github.com/madgraph5/madgraph4gpu/actions/runs/10045773913/job/27763863225?pr=934
```

*******************************************************************************
*** tmad_test pp_tt012j.mad (P2_gu_ttxgu)
*******************************************************************************

Testing in /home/runner/work/madgraph4gpu/madgraph4gpu/epochX/cudacpp/pp_tt012j.mad/SubProcesses/P2_gu_ttxgu

*** (1) EXECUTE MADEVENT_FORTRAN (create results.dat) ***
 [OPENMPTH] omp_get_max_threads/nproc = 1/4
 [NGOODHEL] ngoodhel/ncomb = 32
32/64
64
 [XSECTION] VECSIZE_USED = 32
 [XSECTION] MultiChannel = TRUE
 [XSECTION] Configuration = 1
 [XSECTION] ChannelId = 1
 [XSECTION] Cross section = 0.8205 [0.82047507505698292] fbridge_mode=0
 [UNWEIGHT] Wrote 5 events (found 37 events)
 [COUNTERS] PROGRAM TOTAL          :    0.0631s
[...]

PR #874, opened 2024-06-28
Fix intermittent FPE crashes for SIMD (add volatile) and allow no-OpenMP builds

valassi:
This is a WIP PR to allow no-OpenMP builds (#758) and to debug intermittent FPEs (#845).

This is starting from MR #873 which fixes color mappings and related issues.

The reason why OMP and FPEs are related is that I had some indication that the FPE crash is intermittent with OMP, but happens all the time without OMP. So disabling OMP might trigger the issue more often.
See https://github.com/madgraph5/madgraph4gpu/pull/873/commits/f1e0d42349a97574647cd09b088b2092cfe4ee2b


valassi:
Note: even after disabling OMP in the CI tests, issue #845 has not shown up. The only 6 usual errors are from #826 and #872.

I think that I understand this better now: this issue ONLY happens for AVX512 (and specifically 512z with FPTYPE=f). Our CI on standard github nodes does not offer AVX512 (and it would be an overkill to set up our own nodes just for that). So I will debug this in manual tests.

valassi:
Hi @oliviermattelaer yet another one (we are getting there...).

I have fixed this intermittent crash #845 again by using 'volatile'. This time it should be less controversial I hope. The crash happens deep in SIMD code and only in specific configurations like 512z and fptype=f. So I am quite confident that disabling optimizations with volatile makes sense. Note this interesting post on SIMD and float, https://gcc.gnu.org/bugzilla/show_bug.cgi?id=90993

En passant I added the option (that you had asked for in #758) to disable OpenMP. You just need to export OMPFLAGS= (set an empty string: this is different from unsetting it).

Can you please review?
NB This is based on #873, so I would FIRST review and merge that, and THEN review this (so it becomes easier to see only the additional differences).

Thanks!
Andrea

valassi:
I have rerun the manual tests. This is now ready to be merged (but it would be best to first agree on and merge https://github.com/mg5amcnlo/mg5amcnlo/pull/115 and #873)

valassi:
Putting this back to draft for the moment until we agree on #873 and #877

valassi:
I have merged the latest upstream/master (including PR #852 and PR #877) into this and marked it again as ready for review.

valassi:
Hi @oliviermattelaer I have updated this again with upstream/master, it is ready to be merged
- it fixes an intermittent segfault (#845), again with a 'volatile', but this is very clearly a SIMD AVX512 issue 
- and it allows no-OpenMP builds by setting 'export OMPFLAGS=', which is more or less what you had requested (#117 
Thanks
Andrea

oliviermattelaer:
Perfect to be merged (but I would like to keep #758 open, since I think that we should have an agreement on how we handle OpenMP by default from the madgraph5. (i.e. should I set OMPFLAGS= by default or not, should we have an entry in the run_card to control that /...) -> will copy-paste this message in #758 (since this is the proper place to discuss that).

Thanks a lot,

Olivier

valassi:
Very good, thanks Olivier!

I will merge this then, and reopen #758 

Note, there are 6 errors in the CI but they are the usual failing ones
image

Issue #889, opened 2024-07-05
master_june24: xsec mismatch in ggttggg for FPTYPE=f (must increase tolerance)


valassi:
In WIP PR #882 I added a workaround (NOT a fix) for crash #885 just to allow the CI tests to proceed further. Apart from other 'expected' failures, there is a xsec mismatch for ggttggg. This can be fixed by increasing tolerance
https://github.com/madgraph5/madgraph4gpu/actions/runs/9807713341/job/27082079782
```
*** (2-sse4) Compare MADEVENT_CPP xQUICK xsec to MADEVENT_FORTRAN xsec ***

ERROR! xsec from fortran (2.6007004181875136E-006) and cpp (2.6020625819045072E-006) differ by more than 4E-4 (0.0005237680232093034)
```

valassi:
I increased it to 6E-4. Enough for sse4, not enough for avx2...

https://github.com/madgraph5/madgraph4gpu/actions/runs/9807953772/job/27082866125#step:12:112
```
*** (2-avx2) Compare MADEVENT_CPP xQUICK xsec to MADEVENT_FORTRAN xsec ***

ERROR! xsec from fortran (2.6007004181875136E-006) and cpp (2.6034626920972927E-006) differ by more than 6E-4 (0.001062126914142647)
```

valassi:
Ok fixed in #882

PR #890, opened 2024-07-05
Port to master a CI fix (tmad xsec tolerance increase for 889) from master_june24 work

valassi:
Port to master a CI fix (tmad xsec tolerance increase for #889) from master_june24 work in WIP PR #882


valassi:
This is a totally minor fix in the CI. I merge it immediately without approval.

It is not strictly needed in master, but I prefer to keep the CIs the same while I try to merge wth master_june24

valassi:
This is a totally minor fix in the CI. I merge it immediately without approval.

It is not strictly needed in master, but I prefer to keep the CIs the same while I try to merge wth master_june24

valassi:
The CI has the 6 expected failures from #872 and #826. Self-merging, this is a trivial change in the CI

Issue #904, opened 2024-07-11
Fix build errors with clang16


valassi:
Fix build errors with clang16 - this is a subtask in #734 

The first error is the following
```
ccache /data/avalassi/GPU2023/madgraph4gpuX/tools/compilers/mg-lcg-clang++-16.0.3 -o fcheck_cpp.exe ./fcheck_sa_fortran.o -fopenmp ./fsampler_cpp.o -L../../lib -lmg5amc_common_cpp -Wl,-rpath='$ORIGIN/../../lib'  -lgfortran -L../../lib -lmg5amc_gg_ttx_cpp ./CommonRandomNumberKernel_cpp.o ./RamboSamplingKernels_cpp.o
/cvmfs/sft.cern.ch/lcg/releases/binutils/2.40-acaab/x86_64-el9/bin/ld: ./fcheck_sa_fortran.o: relocation R_X86_64_32 against `.rodata' can not be used when making a PIE object; recompile with -fPIE
/cvmfs/sft.cern.ch/lcg/releases/binutils/2.40-acaab/x86_64-el9/bin/ld: failed to set dynamic section sizes: bad value
clang-16: error: linker command failed with exit code 1 (use -v to see invocation)
```

This can be fixed by adding -no-pie

Then I also disabled OMP for this specific compiler (while waiting to disable OMP generally when merging PR #900)

valassi:
This is closed in PR #905. Closing the issue even if the MR is not yet merged

valassi:
I later fixed also clang17 and added wrappers

(NB I think that clang formatting would behave very differently instead...)

Issue #903, opened 2024-07-11
segfault in constexpr_sin_quad when running runTest.exe through valgrind


valassi:
I am working on adding tests for two warps with different channel #896 

The working idea is to have two tests in runTests.cc (one with and one without channelids)

As a preliminary check of the infrastructure, I duplicated the existing no-multchannel test. This gives a strange and unexpected segfault in a surprising place

The code change is
```
diff --git a/epochX/cudacpp/gg_tt.mad/SubProcesses/runTest.cc b/epochX/cudacpp/gg_tt.mad/SubProcesses/runTest.cc
index 3cc650d66..2adddc2ab 100644
--- a/epochX/cudacpp/gg_tt.mad/SubProcesses/runTest.cc
+++ b/epochX/cudacpp/gg_tt.mad/SubProcesses/runTest.cc
@@ -245,21 +245,35 @@ struct CUDATest : public CUDA_CPU_TestBase
 // Use two levels of macros to force stringification at the right level
 // (see https://gcc.gnu.org/onlinedocs/gcc-3.0.1/cpp_3.html#SEC17 and https://stackoverflow.com/a/3419392)
 // Google macro is in https://github.com/google/googletest/blob/master/googletest/include/gtest/gtest-param-test.h
-#define TESTID_CPU( s ) s##_CPU
-#define XTESTID_CPU( s ) TESTID_CPU( s )
-#define MG_INSTANTIATE_TEST_SUITE_CPU( prefix, test_suite_name ) \
+#define TESTID_CPU1( s ) s##_CPU1
+#define XTESTID_CPU1( s ) TESTID_CPU1( s )
+#define MG_INSTANTIATE_TEST_SUITE_CPU1( prefix, test_suite_name ) \
 INSTANTIATE_TEST_SUITE_P( prefix, \
                           test_suite_name, \
                           testing::Values( new CPUTest( MG_EPOCH_REFERENCE_FILE_NAME ) ) );
-#define TESTID_GPU( s ) s##_GPU
-#define XTESTID_GPU( s ) TESTID_GPU( s )
-#define MG_INSTANTIATE_TEST_SUITE_GPU( prefix, test_suite_name ) \
+#define TESTID_CPU2( s ) s##_CPU2
+#define XTESTID_CPU2( s ) TESTID_CPU2( s )
+#define MG_INSTANTIATE_TEST_SUITE_CPU2( prefix, test_suite_name ) \
[...]

PR #900, opened 2024-07-10
Disable OpenMP by default (enable it only if USEOPENMP=1)

valassi:
Hi Olivier, this is a followup to #758.

As discussed at the meeting yesterday, I am disabling OpenMP by default.

It is now enabled only if USEOPENMP=1.

Can you please check and approve? Thanks
Andrea

Issue #896, opened 2024-07-08
master_june24 add runTest (ME comparison) tests for two warps with different channel


valassi:
This is related to PR #830 and issue #765 and the ongoing work in #882.

In particular it is related to https://github.com/madgraph5/madgraph4gpu/pull/882#issue-2390983670. I assume that at some point the input.txt will allow specifying a range of iconfigs, rather than a single one, but I also assume we are not there yet in the Fortran code in #765.

Nevertheless, on the cudacpp side, I think (unless I did not see them) that we are missing simple tests of the cudacpp machinery where different channelid arrays are sent to the ME kernels. One simple idea is to extend runTest, which presently uses only no-multichannel tests, to also include two different channelids like 1 and 2 (the values are process dependent probably, but can be code generated).

Some sanity checks should also be included in the code eg about fortran warp sizes vs SIMD sizes (or actual cuda warp sizes), and about the fact that channelids in a warp are all the same. Self assigning this.

oliviermattelaer:
>  I assume that at some point the input.txt will allow specifying a range of iconfigs, rather than a single one, but I also assume we are not there yet in the Fortran code in #765.

Just to confirm what we discussed last week and adding more details.
So yes the fortran code is ready to handle more than one ifconfigs even if the input.txt is only designed to receive a single one. The point is that we can only handle configuration that are connected by symmetry since we only have a single grid.

The fortran code is actually reading the file ../symfact.dat where the information about symmetric channel is provided and the fortran code then knows the list of symmetric channel (and the associated symmetry to apply for the phase-space) that can be consider.  (and if ../symfact.dat does no exists, then no symmetry is taken into account). 

One side effect about this is since symfact.dat is in the PXXXXXX directory, if you run madevent from 
the PXXXXX directory or the PXXXXX/G1 directory for the same input.txt then you will get different event file/cross-section/...

Let me stress that this handling of symfact.dat is actually present in the official version of MG5aMC (since more than 10 years). And therefore is also present for the current master branch, which is why this branch is so important (at least when going to GPU) since for the moment the event can be bias in the current master since only one config was used in practise.

oliviermattelaer:
And to state the obvious, yes dedicated test are important here.
I was thinking to add some with "my" new CI but this is still waiting to be accepted...
(anyway more test is good wathever the framework used)

valassi:
[...]

PR #905, opened 2024-07-11
Fix clang16 and clang17 builds

valassi:
Hi @oliviermattelaer these are small fixes for clang16 builds, can you have a look please? Thanks Andrea


valassi:
This closes #904

valassi:
In the meantime I also included similar fixes, and some wrappers, for clang16

valassi:
And I improved the fix, adding -fPIC to fortran compilation instead of -no-pie to clang linking.

One word of motivation:
- I am doing this clang stuff to try an duse address sanitizer (with limited success)
- And I want that to debug the segfault #903 when dulicating a test
- Which in turn I want to implement #896 (adding tests for the channelid functionality)
- Which is part of my work on master_june24...

oliviermattelaer (commented):
Here I'm confused...

I guess I will need to check #908 for the final code... So I put this in hold for the moment.

What confuse me is that you seems to want to fully forbid OpenMP for clang16 and clant17?  But will check in #908

oliviermattelaer:
Hi Andrea, 

Back on this one, given that #900 is merged and that our default is no openmp, would it not make more sense to simply let the code to crash (or refuse to compile) if the user ask for openmp with such compiler?

I guess that you face issue when compiling with such compiler or was it just to simplify some test?

Cheers,

Olivier

valassi:
Hi @oliviermattelaer thanks!

True OpenMP is now disabled by default, but one can always reenable it.

Here what I wanted to makerk is that we know that we have not ported the openmp stuff to clang, we (well I) did not find a way to make it work. If you prefer I can replace the
```
else ifneq ($(shell $(CXX) --version | egrep '^clang version 16'),)
    override OMPFLAGS = # disable OpenMP on clang16 #904
else ifneq ($(shell $(CXX) --version | egrep '^clang version 17'),)
    override OMPFLAGS = # disable OpenMP on clang17 #904
```

by something like
```
else ifneq ($(shell $(CXX) --version | egrep '^clang version 16'),)
$(error OpenMP is not supported by cudacpp on clang16 (#904))
else ifneq ($(shell $(CXX) --version | egrep '^clang version 17'),)
$(error OpenMP is not supported by cudacpp on clang17 (#904))
```

Would that be better? This was impossible when OpenMP was the default, but now that it is disabled by default, this can be a way to signal that it is not supported.

In any case I would definitely do something to signal that this was tested and found problematic. There's so many use cases and combinations, that I think it is important to makrk in the code what we are able to do and what we are not able to do.

oliviermattelaer:
Yes, I like it :-)
As soon as you have done it, this is good to go.

Thanks,

Olivier

valassi:
The CI is showing
163 successful, 3 in progress, and 3 failing checks 
This is expected to end with the usual 6 failures (what is pending is pp_tt012j, which is expected to fail.

Merging now.

PR #918, opened 2024-07-17
Couplings order

oliviermattelaer:
This is based on the work of @roiser (thanks so much for it).
Trying to avoid to do what I suggest him to do (i.e. avoiding global variable).

Putting this in PR to have the CI/CD.
Next step here would be to merge my new test to check other problematic process (to see if it was the same issue)

### DIFF
--- a/epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/model_handling.py
+++ b/epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/model_handling.py
@@ -13,7 +13,8 @@ PLUGINDIR = os.path.dirname( __file__ )
 import logging
 logger = logging.getLogger('madgraph.PLUGIN.CUDACPP_OUTPUT.model_handling')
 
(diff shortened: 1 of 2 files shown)


oliviermattelaer:
SO this 
- [x] fix the missmatch for stop pair production (confirmed by the CI )
- [x] The status of the CI (non merged one) for e+ e- > mu+ mu- a should be consider as fixed (i.e. LTS agree with current value BUT only if setting sde_strategy=1... So they are a bug in LTS and not in the plugin for this process. So the CI might have been checking the value for sde_strategy=2, explaining the report, anyway this can be consider as fixed for this plugin.
- [ ] The CI for g g > t t~ 2g is still crashing (need investigation)

This is not as good as we would have like, but this is progress :-D 
So I remove the draft status on this one.

valassi:
This is also related to #748 (xsec mismatch in gqttq, which was also due to ther order of couplings)

valassi:
Thanks Olivier for approving #924 :-)

I will merge this one and then merge #934.

Andrea

PR #935, opened 2024-07-23
WIP investigate pp_tt012j bug 872

valassi:
This is a WIP PR to investigate the pp_tt012j bug 872


valassi:
I have completed a patch. This fixes the 3 failing tests in pptt012j (the remaiining 3 failing tests are with susy, which is fixed in another set of PRs).

image

I will add more comments later

valassi:
Hi @oliviermattelaer this is a potentially controversial PR, lets discus next week (but I hope it is not).

In practice: 
- bug #872, the cross section mismatch fortran/cudacpp for pp_tt012j, was ultimately caused by the fact that mirror processes are enabled in fortran (MIRRORPROCS=TRUE), but they are not enabled in cudacpp (which expects nprocesses=1)
- we discussed for many years the question whether nprocesses>1 should be supported, and I have done many tests and changes related to that, dicussing with you (see #272 #337 #340 #343 #360 #396 #619  #626, possibly also #759)
- at some point I had added a static_assert that nprocesses=1 is expected in cuda, because the code handles several variables as scalar for nprocesses=1 and does not have the extra array... this of course would be possible if needed, but it is a **HUGE** piece of work, which I hope we can avoid for a release
- my understanding (but this is the question for you, maybe I am wrong) was that it was safe to assume nprocesses=1 and to disable mirror processes, simply because *in that case you generate one more P subdirectory for the mirror process as a separate subprocess* (which gives a performance penalty, but *produces correct physics results*?)
- now, there was one PR #754 that you made in August which actually reintroduces mirror processes, and nprocesses=2, for things like pp_tt012j (you asked me to review it but - my bad - I failed to see the consequence, maybe I did not have a pp_tt012j process yet or I did not look at that, then later I was absent anyway...)
- there was then a second PR #764 in September, while I was away, which removed the static assert nprocesses=1, and also allowed nprocesses=2... this means that from that moment on we have internally inconsistent code
- this PR 935 that I ask you to review is essentially undoing 754 (ie it is disabling mirror processes again) and it is undoing part of 764 (introducing back the sanity check that nprocesses should be 1): the tests now pass for pp_tt012j
- if all I said above makes sense, then I guess this can be approved without issues... otherwise, we have a problem

Specifically
- can you confirm that we can set NMIRRORPROCS=FALSE in Fortran, e.g. because the relevant subprocesses are generated twice?
- otherwise, if madevent fortran must use NMIRRORPROCS=TRUE, can you confirm that there is some way where we can still use nprocesses=1 in cudacpp? because otherwise we are in BIG trouble... (many more weeks of work)
- (by the way, I noticed that when I set 8192 events in the input.txt file, 1 1 iterations, then NMIRRORPROCS=FALSE processes 8192 events in fortran, but NMIRRORPROCS=TRUE proceses more than 8192... is this coming from what SMATRIX1 does internally, i.e. what cudacpp should do internally with nprocesses=2, or is there a switch in the outside calling sequence of madevent that causes this? or maybe I guess it is both?)

Lets discuss next week, thanks
Andrea

valassi:
I change this to WIP again and change the title.

I did more tests and my proposed approach of undoing 754 looks wrong. I compared cross sections with and without patch 754, and the correct cross sections look correct when the patch is there - so removing it would be a mistake.

I will add more comments in #872

valassi:
I have undone the CODEGEN changes - bug #872 will show up again the CI, there will be 6 failing tests normally

valassi:
Hi @oliviermattelaer can you please review? This is now ready.
(I will still add some manual tests, but it can be reviewed).

Essentially, this fixes #872 by calling reset_cumulative_variable once per mirror, to ensure that the same numbers of events are processed in fortran and cudacpp. But it still computes helicities only once (assuming that the two helicity lists in fortran are both the same...).

I also added pp_tt.mad to the repo (and to the CI tests) as this is an easier process than pp_tt012j an dis enough to show and study the issue.

I will also add comments on #940, but essentially there is nothing to do there

Thanks
Andrea

valassi:
PS reset_cumulative_variable issues are related to #486 and https://github.com/oliviermattelaer/mg5amc_test/issues/22

valassi:
The CI has now only 3 failures (in susy)
image

oliviermattelaer (commented):
Hi Andrea,

Here I still have to make my mind on what is the correct approach to this problem.
But two/three things that are not perfect
 - the use of the patch method and you know what I think of that
 - the implicit use of the assumptions that the helicity are identical (asymmetric compare to Fortran)
 - adding a reset of cumulative variable which is not needed for the physics 

Maybe we can use such patch only for the generateandcompare (and equivalent script)?

I do not think that the correct approach is to implement such double helicity filtering within cudacpp. The reason is that, I believe that we should likely investigate the way it is done in Fortran and check if such method is still needed. 
This was done to circomvent a bug. If I'm correct the same bug was faced (again) recently when using helicity-recycling. During that investigation the real cause of the bug was not the helicity but rather a grid initialization error.
So if I'm right, we could just remove this double initialization on the Fortran side.
But this obviously needs to be checked carefully. 

But let discuss this more tomorrow and then decide how to move forward.

valassi:
> Hi Andrea,
> 
> Here I still have to make my mind on what is the correct approach to this problem. 

Hi Olivier, thanks a lot for the comments :-)

> But two/three things that are not perfect
> 
>     * the use of the patch method and you know what I think of that

Well, this one I even agree with you that *eventually* we should get rid of that. (Example your #849 just goes in that direction). But as long as we have it, well that's what we have and we should use that. We cannot block everything else until we fix patch I think.

>     * the implicit use of the assumptions that the helicity are identical (asymmetric compare to Fortran)

Ok here I thought you had no problem, I must have misunderstood your earlier message.

I agree, in principle. But maybe I would leave this for later?

Anyway there are two options
- we add a check in fortran that the two ngoodhel are identical... (incidentally I just added that as a an -posteriori check in my test scripts): I think this is the fastest and should be enough for most cases... in practice, it is about adding a SANITY CHECK and I like the idea
- or, we implement two separate helicity filterings in cudacpp (see also #940)... but this sounds both overly complicated and an overkill, I would avoid that

So I will start looking into adding a sanity check

>     * adding a reset of cumulative variable which is not needed for the physics

No this I disagree. It might not be relevant fo rthe physics of cudacpp, but it is **essential to get the same results in fortran and cudacpp**. I continue to think that our bit-by-bit tests are fundamental. Letting them go would be very wrong.

(Think of the CMS meeting - you were not there - there is a not understood difference between fortran and cudacpp for DY+4jets... maybe this issue here is totally irrelevant there, but what I mean is that we should make sure that if we start from the same random numbers we **expect** to get the same results, unless something strange happens).

Rephrase: adding reset_cumulative_variable has an almost completely negligible cost in performance, and a huge benefit in testing reproducibility. For me this is pure gold, no question we keep this out. 

> Maybe we can use such patch only for the generateandcompare (and equivalent script)?

No, see the example of CMS. We will also get real use cases to debug where fortran and cudacpp give different results. We need this in production.

> I do not think that the correct approach is to implement such double helicity filtering within cudacpp. The reason is that, I believe that we should likely investigate the way it is done in Fortran and check if such method is still needed. This was done to circomvent a bug. If I'm correct the same bug was faced (again) recently when using helicity-recycling. During that investigation the real cause of the bug was not the helicity but rather a grid initialization error. So if I'm right, we could just remove this double initialization on the Fortran side. But this obviously needs to be checked carefully.

I see your point. Listen I propose
- (1) now, I add a sanity check that the two ngoodhel are equal; and then we merge this, including the second reset_cumulative_variable call
- (2) I open a ticket for later on, on you, to eventually recheck if this can be removed also in fortran

Does that sound ok?

> But let discuss this more tomorrow and then decide how to move forward.

Anyway yes we can discuss tomorrow. Thanks again

valassi:
> (2) I open a ticket for later on, on you, to eventually recheck if this can be removed also in fortran

THis is now open #941

valassi:
Change back to WIP while I look into adding a sanity check that the two lists of helicities in fortran are identical

valassi:
In the meantime I merged upstream/master with #918 and #849

oliviermattelaer:
Found a process where this is not identical at the fortran stage... (see #941) So we clearly need to be more careful here...
So this sounds more tricky than I thought... (this CI offers a good catch to find such tricky issue).
But ok, I need to go back to the physics to understand why such situation makes sense physically (or what is the real bug behind --unlikely?--)

valassi:
As I mentioned, I created a sanity check here
https://github.com/madgraph5/madgraph4gpu/pull/935/commits/77f13eebfe0b3fdc7c6cb72ed6eed9add9851958

For the moment I see no failures.

As I mentioned, I think that the general fortran case is different from fortran in cudacpp, because in the former case we mix all quark/antiqurk initial states in qq, in the latter we separate uux, uc, ucx etc etc. So we are really only left with mirroring from inverting right/left beams... I guess only polarised beams could make a physics difference then. 

Anyway, I confirm my proposed strategy
- merge this now to have fortran and cudacpp give bit by bit identical
- have the sanity check in, so that we are notified if/when a process gives two different helicity lists
- wait and see: for the moment, I see no processes where this gives a problem... rather than over-engineer now, I would be pragmatic and address a problem only if it shows up (and I do not think it will)

valassi:
I have rerun all my manual tests too, I confirm this as good to go, unless we decide on a different strategy

oliviermattelaer:
Ok, so #941 concluded 
that 
 - Yes, this can be removed from the fortran side
 - That if using limhel (which is important to do) then the list can be different due to the different random number used to fill both list.
 - that for the moment limhel is set to zero  (this need to be clarified when this is happening)
 - that the list of helicity was common for all matrix#.f (irrelevant for the gpucpp case since we ensure to have only one)

So I'm working on a PR for the MG5aMC: https://github.com/mg5amcnlo/mg5amcnlo/pull/124
which, in particular: remove the need of imirror but add the support to have different list for different matrix_element.

Let discuss next week.

valassi:
Thanks Olivier. On my side I updated this PR with #949 upstream/master.

My proposal is the following:
- now/soon we merge this PR with at least a temporary patch (the 2nd call to reset cumulative variable) that makes fortran and cudacpp tests identical
- then/later you work on your https://github.com/mg5amcnlo/mg5amcnlo/pull/124 and when this is done I remove the second cumulative_variable call and adapt the code more generally
- also then/later at some point I work on #950 and I move back to LIMHEL>0 in production but add a runcard to allow LIMHEL=0 and use that as default in madgraph4gpu

But I would not wait to have all of the above done before fixing these pptt012j tests... I want to see the cross sections identical, and be able to debug CMS without having that issue, too

Anyway lets discuss next week! Have a good weekend :-)

Andrea

oliviermattelaer:
So my counter proposal is to move directly on this PR: https://github.com/madgraph5/madgraph4gpu/pull/955
where all CI are now green (and which also removes the first reset_cumulative call since now fortran has a smarter trigger to check if such call is needed and therefore does not need it for the cases tested by your test (this is only needed for MC over helicity).

And then move on https://github.com/madgraph5/madgraph4gpu/issues/950
(which is now in the run_card).

valassi:
I will look at 955 from Olivier, but in the meantime I resynced this to the latest upstream/master

valassi:
I renamed this and marked this as WIP.

The initial focus of this PR (fixing 872 via reset_cumulative_variable) is obsolete. The issue has been fixed by Olivier in his #955, which will be included in the upcoming #992.

I keep this open just in case I want to add pp_tt (temporarely?) to the repo. Low priority, will have a look after #992.

Issue #952, opened 2024-08-06
Fix clang-format version and eventually upgrade from v15 to v17


valassi:
I am using two different machines
```
[avalassi@itgold91 gcc11/usr] ~> clang-format --version
clang-format version 17.0.6 (AlmaLinux OS Foundation 17.0.6-5.el9)
[avalassi@itgold91 gcc11/usr] ~> more /etc/redhat-release 
AlmaLinux release 9.4 (Shamrock Pampas Cat)

[avalassi@itscrd90 bash] ~> clang-format --version
clang-format version 15.0.7 (Red Hat 15.0.7-2.el9)
[avalassi@itscrd90 bash] ~> more /etc/redhat-release 
AlmaLinux release 9.2 (Turquoise Kodkod)
```

All code so far was formatted with the latter ie clang-format v15.

I noticed some differences between clang-format 15 and 17.
The templates for codegen etc must be ported to v17.

For the moment I will simply fix the version to <=15 using clang-format from cvmfs.

PR #953, opened 2024-08-06
Fix clang-format version to between v13 and v15

valassi:
Fix clang-format version to v15 see #952


valassi:
Hi @oliviermattelaer I'd like to merge this, just to have fewer PRs up in the air.

This should be totslly non controversial, only changes a script that I use in generateAndCompare. It needs to use clang format not higher than v15, else all formatting will change.

Let me know if you have comments please, but I might merge anyway.

valassi:
Note: I could not test this fully on itgold91 because of https://github.com/mg5amcnlo/mg5amcnlo/issues/127

oliviermattelaer:
no objection.

valassi:
> no objection.

Thanks :-)

Merging

Issue #950, opened 2024-08-02
Move LIMHEL=0 to madgraph4gpu only and add a runcard allowing LIMHEL=0


valassi:
Hi @oliviermattelaer this is a followup to #941 and PR #935. Two points

> You can set limhel to 0 for test if you want but in production this should be the same fortran code/design.

(1) In https://github.com/madgraph5/madgraph4gpu/issues/941#issuecomment-2264800750 you suggest that LIMHEL=0 should not be the default in production cudacpp. Now it is. So it should be changed eventually. I still want to use LIMHEL=0 for all my tests in madgraph4gpu, so I can move that to generateAndCompare.sh for the code generation in the madgraph4gpu repo.

(2) However, I still think that LIMHEL=0 is a useful setting for tests, also for users, when debugging issues (eg the CMS xsec difference). So I suggest adding it as a hidden parameter in the runcards. Objections or other suggestions?

Thanks
Andrea

oliviermattelaer:
Yes make sense to allow to control such parameter via the run_card. Will look at that

Sent from Outlook for iOS<https://aka.ms/o0ukef>
________________________________
From: Andrea Valassi ***@***.***>
Sent: Friday, August 2, 2024 6:34:45 PM
To: madgraph5/madgraph4gpu ***@***.***>
Cc: Olivier Mattelaer ***@***.***>; Mention ***@***.***>
Subject: [madgraph5/madgraph4gpu] Move LIMHEL=0 to madgraph4gpu only and add a runcard allowing LIMHEL=0 (Issue #950)


Hi @oliviermattelaer<https://github.com/oliviermattelaer> this is a followup to #941<https://github.com/madgraph5/madgraph4gpu/issues/941> and PR #935<https://github.com/madgraph5/madgraph4gpu/pull/935>. Two points

You can set limhel to 0 for test if you want but in production this should be the same fortran code/design.

(1) In #941 (comment)<https://github.com/madgraph5/madgraph4gpu/issues/941#issuecomment-2264800750> you suggest that LIMHEL=0 should not be the default in production cudacpp. Now it is. So it should be changed eventually. I still want to use LIMHEL=0 for all my tests in madgraph4gpu, so I can move that to generateAndCompare.sh for the code generation in the madgraph4gpu repo.

[...]

PR #955, opened 2024-08-06
fix patch to be compatible with mg5 branch gpucpp_goodhel

oliviermattelaer:
This is a simple patch to make the plugin work for the gpucpp_goodhel branch
(removing the double list for the helicity initialization and allowing limhel to be control from the run_card.)

The fact that limhel is now a user parameter makes it important to be correctly handle by the cudacpp interface as well which is not done in this PR.

### DIFF
--- a/epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/MG5aMC_patches/PROD/patch.P1
+++ b/epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/MG5aMC_patches/PROD/patch.P1
@@ -304,13 +304,13 @@ index e73e654d4..27fbe7302 100644
  C     ----------
  
 +      call counters_smatrix1_start()
-       NTRY(IMIRROR)=NTRY(IMIRROR)+1
-       THIS_NTRY(IMIRROR) = THIS_NTRY(IMIRROR)+1
+       NTRY(1)=NTRY(1)+1
        DO I=1,NEXTERNAL
+         JC(I) = +1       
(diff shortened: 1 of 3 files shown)


oliviermattelaer:
@valassi 

The PR obviously now crashes since limhel !=0 since now that parameter is part of the run_card setup.
I can do a quick/temporary fix which is setting that parameter in the run_card by default when doing output in the simd/cuda mode to have this PR going trough quickly.
But I guess that we need to 
1) change the test to enforce limhel=0
2) change cudacpp to support correctly limhel (even if the algorithm is not 100% identical to the fortran one, like when this was done via a running sum rather than comparing one helicity to the average of the helicity for that event (like done in fortran).

Do you want to do 1 or 2 directly? or should I implement the different default value?
(which is super easy to do)

oliviermattelaer:
I had to remove a useless call to reset_cumulative_call in the cudacpp since now I discard those in the fortran.
And now all tests are passing.
Just we can either review/merge this or first fix the test and put back in cudacpp some support of limhel.

(maybe we can skype on this tommorow --if you are available--)

oliviermattelaer:
Hi,

Just to comment that the check that MC over helicity was not used for the cudacpp plugin was already in place.

So do not need to do anything here.

Olivier

valassi:
This #955 'goodhel' as well as the additional #986 'additional goodhel' (which would be needed to fix the master conflicts) are included in the #992 which is about to be merged. I close this for simplicity.

In any case I have marked that #992 fixes also the issues #872 and #950 that are fixed in this #986 and the underlying #955

Closing this for simplicity. All this work will be merged via #992.

Issue #958, opened 2024-08-08
Reduce madevent 'Fortran overhead' by restricting cudacpp helicity calculation (which scales with SIMD) to only 16 events


valassi:
Understand if madevent 'Fortran overhead' actually includes a parallelizable c++ component

See the discussion in #943 for CMS DY+3jets

```
[avalassi@itgold91 gcc11/usr] /data/avalassi/GPU2024/madgraph4gpuX/epochX/cudacpp/tlau/fromgridpacks> ./parseGridpackLogs.sh pp_dy3j.mad
pp_dy3j.mad/fortran/output.txt
[GridPackCmd.launch] OVERALL TOTAL    442.6794 seconds
[madevent COUNTERS]  PROGRAM TOTAL    438.456
[madevent COUNTERS]  Fortran Overhead 131.711
[madevent COUNTERS]  Fortran MEs      306.744
--------------------------------------------------------------------------------
pp_dy3j.mad/cppnone/output.txt
[GridPackCmd.launch] OVERALL TOTAL    734.9419 seconds
[madevent COUNTERS]  PROGRAM TOTAL    730.685
[madevent COUNTERS]  Fortran Overhead 424.426
[madevent COUNTERS]  CudaCpp MEs      306.259
--------------------------------------------------------------------------------
pp_dy3j.mad/cppsse4/output.txt
[GridPackCmd.launch] OVERALL TOTAL    437.2589 seconds
[madevent COUNTERS]  PROGRAM TOTAL    433.014
[madevent COUNTERS]  Fortran Overhead 279.465
[madevent COUNTERS]  CudaCpp MEs      153.548
--------------------------------------------------------------------------------
pp_dy3j.mad/cppavx2/output.txt
[GridPackCmd.launch] OVERALL TOTAL    260.5085 seconds
[madevent COUNTERS]  PROGRAM TOTAL    256.272
[madevent COUNTERS]  Fortran Overhead 193.151
[madevent COUNTERS]  CudaCpp MEs      63.1221
--------------------------------------------------------------------------------
pp_dy3j.mad/cpp512y/output.txt
[GridPackCmd.launch] OVERALL TOTAL    253.3131 seconds
[madevent COUNTERS]  PROGRAM TOTAL    249.01
[madevent COUNTERS]  Fortran Overhead 189.606
[madevent COUNTERS]  CudaCpp MEs      59.4043
--------------------------------------------------------------------------------
pp_dy3j.mad/cpp512z/output.txt
[...]

Issue #564, opened 2022-12-11
Allow LIMHEL>0 in cudacpp? (Or allow an input set of pre-filtered helicities)


valassi:
Before I forget, this is another followup of #419.

Presently fortran can use LIMHEL>0 to use a more relaxed helicity filtering (include helicities only if one of a few events has a ME above a threshold). In cudacpp the comparison is always to 0, and as a consequence the fortran code is also modified to use LIMHEL=0. If physicists can accept LIMHEL>0 to further speed up the code (at the cost of some precision), maybe this should be implemented in cudacpp too. 

HOWEVER, I would like this to be done in such a way that the results can be compared exactly to fortran. Anpther aoption is to pass to cudacpp a pre-filtered set of helicities computed in fortran 'the fortran way'.

valassi:
I have changed the cudacpp getGoodHel implementation in MR #705 to solve the mismatch in fortran/cpp cross sections for gg_uu described in issue #630.

The new implementation in cudacpp is now using an algorithm that is a bit closer to that in fortran, and it would therefore be possible in principle to reimplement the exact same LIMHEL algorithm in cudacpp as in fortran. I actually tried a quick test in https://github.com/madgraph5/madgraph4gpu/pull/705/commits/dcfababbadb3cd5fa668c2ff7b58feccc91c31b7 but the fortran test crashed (not visible in the log, unfortunately).

In any case I would not try the option to pass from fortran to cudacpp an input set of prefiltered helicities: I would indeed try to reimplement the excat same algorithm in cudacpp as in fortran.

This task includes several sub tasks, including
- First of all, make sure that LIMHEL>0 does not crash in fortran. 
- Then, in cudacpp, implement the same algorithm. The point is that one should not compare to LIMHEL just the ME contribution from one helicity: one should compare it to the ME contribution for one helicity, times ncomb (all possible helicities), divided by the sum of ME contributions from all helicities. This is what is done in fortran. One should use debug printouts to check we are really comparing the same thing.
- There is then also the issue of understanding the effect of LIMHEL on event selection and on cross sections, or at least checking that this goes on in the same way in fortran and cudacpp. As discussed in #630, small changes in the helicity selection may end up having large effects on event selection (and on cross sections, when few events are used and the cross sections are divergent like gguu), because helicity selection has a large effect ion numerators and denominators of the multichannel mode.

valassi:
Note this I had discussed here https://indico.cern.ch/event/1162062/

[...]

PR #960, opened 2024-08-08
remove unnecessary first pass on all bridge events in cudacpp helicity calculation from madevent (and improve timers)

valassi:
This is a WIP PR to improve timers and cudacpp helicity computation in madevent #958 and #546


valassi:
Hi @oliviermattelaer this is essentially ready for review. Can you please reviww?
(But I will still launch some manual tests on all processes)

I renamed it as "remove unnecessary first pass on all bridge events in cudacpp helicity calculation from madevent (and improve timers)". Essentially what this does is
- I added a boolean parameter "goodhelonly" to the fortran-cpp interface of the bridge (this already existed in the cpp bridge but was not exposed to fortran yet)
- in the madevent executable, during the first pass on cudacpp bridge that is meant to compute only helicities, I call the bridge with this flag=true: this ensures that only a few events (chosen by cudapp, typically 16) are used for helicity computation
- without this patch, what was happening was that, with 16384 events in the bridge, the first pass was computing MEs for all 16k events, *even if only 16 were used to cmpute helicities*
- note that the computation scales with SIMD, which explains why it was faster with AVX2 than in no SIMD
- note also that this was adding a huge overhead to the cudacpp calculation with respect to fortran, because fortran does not have the extr acalculation of these 16k events
- and note also that the overhead was reduced when running with only 32 events in the grid, because then this forst pass was only 32 events and not 16k

So I would say that all is understood. This was a nasty performance overhead. Now cudacpp should look a bit better with respect to fortran, especially if only few events are generated

epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/MG5aMC_patches/PROD/patch.P1
```diff
@@ -299,15 +301,7 @@ index e73e654d4..27fbe7302 100644
...
-@@ -140,6 +143,7 @@ C     ----------
- C     BEGIN CODE
- C     ----------
- 
-+      call counters_smatrix1_start()
-       NTRY(IMIRROR)=NTRY(IMIRROR)+1
-       THIS_NTRY(IMIRROR) = THIS_NTRY(IMIRROR)+1
-       DO I=1,NEXTERNAL
-@@ -217,6 +221,17 @@ C     ----------
+@@ -217,6 +220,17 @@ C     ----------
```
oliviermattelaer, line 304: This is weird no?

Those lines are also heavily changed in the goodhel PR.

epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/MG5aMC_patches/PROD/patch.P1
```diff
@@ -325,22 +319,3 @@ index e73e654d4..27fbe7302 100644
...
-         ENDIF
-       ENDIF
-       ANS=ANS/DBLE(IDEN)
--
-       CALL SELECT_COLOR(RCOL, JAMP2, ICONFIG,1,  ICOL)
--
-+      call counters_smatrix1_stop()
-       END
- 
-
```
oliviermattelaer, line 346: same here ....

oliviermattelaer (commented):
I guess that this would make more sense to include this in the master_goodhel
branch since the two are in a way related.

This will allow to move the HELONLY from a boolean to a float and pass limhel to the code to have this selection of helicity to be done as it should.

Or we can merge this one, and then I do the switch in master_goodhel to have that parameter as a float. (but we should not wait for merging master_goodhel anyway)

But can you comment way some file are modified on matrix.f this is weird/bad

Cheers,

epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/MG5aMC_patches/PROD/patch.P1
oliviermattelaer: Ah this is in the patch ... not in the matrix.f (not sure why I was confused)...
So this makes more sense ...

valassi:
> I guess that this would make more sense to include this in the master_goodhel
> branch since the two are in a way related.

Hi Olivier, thanks for loking at this.

One point: I really think that we should try NOT to have too many master_xxx branches, and rather have everything in master. Otherwise it becomes unmanageable. I work on master, and I also worked on master_june24 (to merge it in master, essentially ready), but I would avoid more branches. (What I try to do is to have several PRs onmaster in parallel, and in case I need one inside another then I do include them, but always trying to make sure that they can be merged to master one after the pther).

This specific PR makes sense against master, I would put it on master. And it does NOT have much to do with the other hel work you are doing in fortran (IIUC), because you are reducing from two fortran helicity calculations to one. Here I am just modifying how many events are used in the cudacpp helicity computation, I am not changing one vs two computations in cudacpp nor am I touching what is done in fortran. So I would really merge it in master.

> This will allow to move the HELONLY from a boolean to a float

Uh? This I really do not understand, I should see what you have done. But again, I'd prefer to merge this to master first and then look at the rest. All the tests I am doing for CMS use this now.

Thanks :-)
Andrea

epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/MG5aMC_patches/PROD/patch.P1
valassi: Here I just removed the addition of a ` call counters_smatrix1_start()` line. This is a counter that is no longer used (I think StephanH had removed it from cudacpp, but it was still there in fortran). So my PR on this is actually making the patch smaller.

The other counter calls that are still there are in a template of cudacpp rather than a patch I think, but I should check.

oliviermattelaer:
> One point: I really think that we should try NOT to have too many master_xxx branches, and rather have everything in master.

Approve the merge of the CI, and master_goodhel and ... they will be in master.

The multiplication of master_xxx is just that I have parallel work on therefore multiple PR (which for some weird reason takes a while to be merged). While master_june24 was special the others are quite "normal branch" (and the master_ prefix is just  my typical naming of branch indicating from which branch I'm originating from.

> Uh? This I really do not understand, I should see what you have done. But again, I'd prefer to merge this to master first and then look at the rest. All the tests I am doing for CMS use this now.

No problem, I will do a master_goodhel_limhel branch then to show you what I want to do for it (but if we merge master_goodhel in master  --which should be uncontroversial-- first such that I can do a master_limhel branch).

valassi:
I have updated this PR including
- the fixes for cms nvcc without nvtx PR #966
- the fix in the mac CI PR #974 
- conflict fixes in codegen logs

This should be ready to merge, when the CI tests have passed

valassi:
> > > This will allow to move the HELONLY from a boolean to a float
> >
> > Uh? This I really do not understand, I should see what you have done. But again, I'd prefer to merge this to master first and then look at the rest. All the tests I am doing for CMS use this now.
> 
> No problem, I will do a master_goodhel_limhel branch then to show you what I want to do for it (but if we merge master_goodhel in master --which should be uncontroversial-- first such that I can do a master_limhel branch).

Hi @oliviermattelaer thanks again for the discussion yesterday. Ok now I understand better what you mean. We could reuse the same parameter as a float in the following way
- if it is >0 or =0: consider this as a helicity filtering call (ie goodhelonly=true), with the float representing LIMHEL, ie start the cudacpp helicity filtering on a subset of events and with that value of LIMHEL (see #564)
- if it is <0 (eg say -1 by convention as you suggested), then consider this as a normal ME computation, which assumes that helicities have been precomputed.

Note also that I opened #975 about possibly adding channelid to helicity filtering (which now instead is done on all channels).

valassi:
The CI completed, there are the usual expected 3 failures
image

Hi @oliviermattelaer as discussed yesterday, I now self-merge this. Some followup work remains to do in
- #564 (and maybe elsewhere) for passing LIMHEL>0 to cudacpp
- #975 for using channels in helicity filtering

I also still need to review and merge your comprehensive helicity changes in fortran (which I think is #955 and related work)

Thanks Andrea

Issue #965, opened 2024-08-13
Compilation Error in cudacpp_backend=cpp for CMS DY+nj and TT+nj (nvcc exists but not NVTX or curand)


choij1589:
When attempting to compile the project using the ```cudacpp_backend = cpp``` configuration, encounter a compilation error with CMS DY+nj and TT+nj processes.

## Steps to reproduce
1. Used madgraph4gpu master branch (Aug 13., 2024)
2. proc_cards and run_cards are here: [DY0j_LO_5f_CPP](https://github.com/choij1589/genproductions/tree/mg4gpu/bin/MadGraph5_aMCatNLO/cards/13p6TeV/mg4gpu/DY0j_LO_5f_CPP)
3. Please modify $DEFAULT_PDF_SETS to 325300 and erase the line with $DEFAULT_PDF_MEMBERS: [run_card.dat](https://github.com/choij1589/genproductions/blob/mg4gpu/bin/MadGraph5_aMCatNLO/cards/13p6TeV/mg4gpu/DY0j_LO_5f_CPP/DY0j_LO_5f_CPP_run_card.dat#L34-L35)
4. launch event generation

## Environements
Using gcc-12 with CUDA 12.2, but also encountered same error in CUDA 11.8 

Compilation fails with the logs below:
[run_01_tag_1_debug.log](https://github.com/user-attachments/files/16599615/run_01_tag_1_debug.log)

valassi:
Hi Jin, thanks a lot for the details :-)

I think that the problem here is NOT related to the process DY or TT.

The problem is instead most likely related to your build environment. I managed to reproduce it by creating a dummy `nvcc` script and adding it to my path, but without a real cuda installation
```
echo "echo HALLO" > nvcc
chmod +x nvcc
\mv nvcc ~/bin
rehash
nvcc
  HALLO
```
and then
```
[avalassi@itgold91 gcc11/usr] /data/avalassi/GPU2024/madgraph4gpuX/epochX/cudacpp/gg_tt.mad/SubProcesses/P1_gg_ttx> make -j -f cudacpp.mk
...
[...]

PR #966, opened 2024-08-14
Fix cudacpp makefile for installations (e.g. in CMS) where nvcc exists but curand and nvtx headers are missing 

valassi:
Hi @oliviermattelaer this is a very simple PR, which only modifies the cudacpp.mk for installations (e.g. in CMS) where nvcc exists but curand and nvtx headers are missing.

It fixes the issues reported by @choij1589 in #965 and at the meeting yesterday.

Jin could already work around the issue by setting `USE_NVTX= HASCURAND=hasNoCurand` but I think that t is much better if some sanity checks for the existence of the headers is done elsewhere. (Yes I know cmake would make that out of the box, but we are not there).

Can you have a look please? This is rather standalone and independent from all other PRs I submitted.

Thanks
Andrea


valassi:
Hi @oliviermattelaer can you review this one? It fixes the issue #965, @choij1589 confirms it is good.

(Note, some CI tests are failing because of Mac configuration issues #971... orthogonal to this issue)

Thanks

oliviermattelaer (approved):
This can move forward (even if it is indeed a strong argument to move to a more automatic/flexible way to compile the code, like you said

valassi:
> (Note, some CI tests are failing because of Mac configuration issues #971... orthogonal to this issue)

This is now fixed in PR #974 which I will merge. But I will not rerun MAc tests for this 966

valassi:
> This can move forward (even if it is indeed a strong argument to move to a more automatic/flexible way to compile the code, like you said

Thanks @oliviermattelaer ! Ok so I am merging this one.

PR #979, opened 2024-08-30
PR where I only add CI but not trying to fix them

oliviermattelaer:
Hi @valassi,

This is the PR where I only add the file for the new CI.

The new CI is based on the test suite of MG5aMC, and therefore the github file is kind of a copy of what we have in MG5aMC

However, this file is auto-generated (but need to be committed for github to run the CI --maybe this is not needed but then I do not know how to do it).

So if you want to edit the CI, you need to run (from the acceptance directory)
```
python create_acceptance_from_file.py
```

This will create/overwrite two files:
 - .github/.github/workflows/madgraph_launch_test.yml (the CI itself)
 - acceptance_tests/test_simd_madevent.py (the file compatible with the test suite of MG5aMC, used by the CI)

The content of those two files are related to the directory `simple_cross_check` which contains MG5aMC script and metadata for example the file `simd_cpp_eemumua_float`
```
 #title check eemumua in single precision
 import model sm
                  set automatic_html_opening False --no_save
                  set notification_center False --no_save
                  generate e+ e- > mu+ mu- a
                  output madevent_simd %s -f -nojpeg
                  launch
                  set nevents 100
                  set floating_type f
 #check run_01 0.0239204 0.0002854 100
```
line starting with `#XXXXX` are metadata (title is for the comment/information of the CI, check is doing check of the expected cross-section/error/number of events)   

So if you add one of such file, and run the above python script, you do have one more CI, checking the cross-section for one additional case (ensuring that cross-section are stable when changing commit).

To run this locally, you need to run (on the mg5amcnlo side)
./tests/test_manager.py -P PATH_TO_CUDACPP_PLUGIN_ACCEPTANCE_DIR
And you can use all the feature of that script to filter which script you want to run.
for example if you want only `test_simd_cpp_ggtt_mixed`

you can do 
./tests/test_manager.py -P PATH_TO_CUDACPP_PLUGIN_ACCEPTANCE_DIR test_simd_cpp_ggtt_mixed
to do all those related to ggtt
```
./tests/test_manager.py -P PATH_TO_CUDACPP_PLUGIN_ACCEPTANCE_DIR .*_ggtt_.*
```

I would propose to merge this super quickly since this starts to be a blocker for the nd_warp_used/...
(especially since this will interact with the broken CI)


valassi (approved):
Approve, but we need also #980 and #984

valassi:
Hi @oliviermattelaer some of your confusion in https://github.com/madgraph5/madgraph4gpu/pull/984#issuecomment-2323026877 probably came from the fact that this testsuite_only did not include the latest upstream/master. I have merged master into it now.

valassi:
Hi @oliviermattelaer as discussed, I am merging this 979 into master.

Then I will merge into master #980, which includes also #984 merged into it.

PR #1002, opened 2024-09-16
Ensure that the order in param_card.inc is reproducible in the madgraph4gpu repo

valassi:
Hi @oliviermattelaer this should be a no brainer. 

I need this to ensure reproducibilit when I run various tests. I noticed that Param_card.inc was sometimes changed during a rebuild. I understood where it comes from, fixed it and regenerated. This is the same issue as in https://github.com/oliviermattelaer/mg5amc_test/issues/2, which has resurfaced after various changes in patchMad.sh etc.

**This ONLY affects the generated code in the madgraph4gpu repo.**

Can you please approve? Thanks
Andrea


oliviermattelaer (approved):
Ok a change only in generateAndCompare.sh script is something that is indeed not a problem.

valassi:
Thanks! Merging

PR #1008, opened 2024-09-26
v1.0.0 release: tarball creation/installation, plugin symlink management, release management

valassi:
Hi @oliviermattelaer this is a WIP PR for tarball creation/installation and plugin symlink management

The part that is done is the usage of the MG5aMC_PLUGIN trick for the symlink management (nice trick, thanks!)

What I am still testing (also in the CI here) is the package creation

NB there is not yet any renaming of plugin directories or of repositories, this can come later


valassi:
@oliviermattelaer This is the WIP that goes with https://github.com/mg5amcnlo/HEPToolsInstallers/pull/6

As I mentioned there, the problem is I am unable to delete/recreate tags, there is some confusion between tags/releases and what you can do with the web, with git tag, with gh, with the CI plugins, So some confuson to be sorted out, I think it will come together at some point.

As I mentioned, the idea would be to have a running tag like cudacpp_for360_latest. I was able to do it frm the web interface, then I added a script bbut that is not enough, and now the web interacce does not allow me to recreate a tag even if I think I deleted it. (amybe also something to do with the meaning of /ref/tag).

Will see tomorrow, but I think this is slowly converging

valassi:
Hi @oliviermattelaer this is now good to go for me. (cc @roiser for info)

This PR includes many things related to 
- 1 plugin symlink management (for the mg5amcnlo which is a submodule in madgraph4gpu)
- 2 tarball creation, tarball installation and release management (for the tarball to be downloaded into a user mg5amcnlo, with no git submodules)

In practice, this corresponds to 'option 3' in the discussion 10 days ago https://indico.cern.ch/event/1355161/
(This is an option I was not considering before the meeting, but the discussion with Olivier during/after the meeting made it clear that it was simpler and better)

More in detail

(1) Using cudacpp within madgraph4gpu

Here I used the approach suggested by Olivier, of having a separate MG5aMC_PLUGIN directory (the 'ATLAS approach'?). This is much easier and better than what I was imagining of using 'install cuda' with some options to add a symlink. Instead, I created a permanenet symlink MG5aMC/MG5aMC_PLUGIN/CUDACPP_OUTPUT, a bit like we used to have. 

This was quite easy. The tricky part was adapting all scripts and tests. Note that in practice from inside madgraph4gpu one must now do the following to use mg5amcnlo:
```
PYTHONPATH=.. ./bin/mg5_aMC -m CUDACPP_OUTPUT
```

Note that mg5amcnlo is still a git submodule at the same place.

(2) Using cudacpp within a user standard mg5amcnlo

Here there are no git submodules needed.

This is based on a modified version of HEPInstaller as in https://github.com/mg5amcnlo/HEPToolsInstallers/pull/6 . You find there the instructions to run the mg5amcnlo code. Using HEPInstaller essentially downloads https://github.com/valassi/madgraph4gpu/releases/download/TEST_cudacpp_for3.6.0_latest/cudacpp.tar.gz

In this PR #1008 instead there is the magic to create that cudacpp.tar.gz tarball at that web address. Everything is based on a github CI using the action https://github.com/softprops/action-gh-release which is extremely handy and easy to use. (Yesterday I was reporting some problems, but I was just tored and doing stupid mistakes: things are really smooth, instead).

An archiver.sh script creates the tarball (and also a VERSION.txt file). This is called by an archiver.yml workflow, whenever a new tag is created. Using the action above, the *git tag* is used to create a *github release* (two different things), and the cudacpp.tar.gz and VERSION.txt are uploaded to that github release as release assets (which I think should be permanent - unlike standard workflow CI artifacts that are limited to 90 days).

To create the tags, I added a gitTag.sh script. This is now connected to the valassi/install branch and uses a TEST_cudacpp_ tag convention, but this can be changed. I created the last tags as follows
```
./gitTag.sh  1.0.0_test001
```
This creates these two tags and releases
https://github.com/valassi/madgraph4gpu/releases/tag/TEST_cudacpp_for3.6.0_v1.0.0_test001
https://github.com/valassi/madgraph4gpu/releases/tag/TEST_cudacpp_for3.6.0_latest

The 3.6.0 is taken directly from the mg5amc submodule inside madgraph4gpu at that point. The VERSION.txt also prints out the minimal and last validated version in the plugin __init__.py.

The 1.0.0 MUST match the plugin version number inside __init__.py. The format can be simply x.y,z, or x.y.z_txt to have some test/prerelease tags (example here _test001). This is because eventually the leading TEST_ will disappear, and it will be handy to have cudacpp_for3.6.0_v1.0.0_pre1 tags for instance.

The idea is that the v1.0.0_xxx tag is a one off and should normally not be changed (unless -f is passed), while the 'latest' tag is by definition a moving tag and is always rebuilt.

Note, with respect to the discussion in https://github.com/mg5amcnlo/HEPToolsInstallers/pull/6: this mechanism *avoids the need of a info.dat file* to have the database matching. Inside mg5amcnlo, using `install cudacpp` would always automatically download from github the latest tarball for release x.y.z. There is no need for more complex matching.

But @oliviermattelaer you know better, maybe it is still better to have that management. To be discussed.

Ah one detail, whenever you create releases there are by default two archives of the Source Code. Since madgraoh4gpu now is huge and contains a lot of useless old stuff, I configured these two archives to be empty (it is only possible to remove stuff from these two archives - fo rinstance you cannot change the location of files, so I could not use them to create the cudacpp.tar.gz if I do not first move all files around).

(3) Some cleanup to do in any case

Amongst other things
- I added a HEPInstaller symlink in madgraph4gpu, but this is not needed, I will remove it
- If we decide to keep (or at least to try initially) this mechanism for downloading the tarball, then the prefixes and directories in all scripts must change (now it is a TEST_ tag inside valassi/install, it should become a non test tag inside the upstream master)
- If instead the usual file.dat database is used, hepinstaller must be changed accordingly and Olivier has to do all the magic... but I think that the releasing of cudacpp still can stay as is 
- Probably, in archiver.yml should avoid creating a release if the tag is not a release tag (the logic is in gitTag.sh but not in archiver.sh yet), but this is not really important

AH IMPORTANT
- This whole stuff now uses AS-IS the madgraph4gpu repo. But it would still be good to clean it up: remove epoch1, epoch2 etc; rename the repo from madgraph4gpu to something else (mg5amcnlo_gpucpp? so that eventually may also have sycl or whatever); move it to the mg5amcnlo hat. All thes echanges can be done in a second moment, after agreeing the rest above. In any case the scripts to create the tarball would remain almost unchganged apart from the directory structure.
- I would definitely keep in any case the madgraph4gpu and just strip it down, because this keeps all ssues and PRs which contain a lot of documentation. I made some simple tests that show that (unless you then recreate the same project after renaming!) all links are nicely maintained across github, so from mg5amcnlo pointing to madgraph4gpu, the links woul dbe automatically relinked to the new mg5amcnlo_gpucpp.

Voila that's all. @oliviermattelaer up to you, either for approval now (only if you also approve hepinstaller), or to be discussed at the meeting or offline, and changed. In any case **the whole chain works**.

Cheers, bon weekend Andrea

oliviermattelaer:
Here is a proposal of a script to create the info.dat automatically based on your tarball system and the system of tags.

```
import subprocess
import sys

def get_tag():
    out = subprocess.check_output(['git', 'tag']).decode()
    return out.split('\n')

def fake_tag():
    return ['golden', 'cudacpp_for3.6.0_v1.1', 'cudacpp_for3.6.1_v1.1', 'cudacpp_for3.6.1_v1.2']

def get_supported_version(tags):
    tags  = [ t.split('_',2)[1][3:] for t in tags if t.startswith('cudacpp_for')]
    tags = set(tags)
    tags.add('3.6.1')
    return tags

def create_file(path, versions):

    line = "%(version)s https://github.com/madgraph5/madgraph4gpu/archive/refs/tags/cudacpp_for%(version)s_latest.tar.gz \n"
    fsock = open(path, 'w')
    for v in versions:
        fsock.write(line % {'version':v})

if "__main__" == __name__:
    tags = get_tag()
    #tags = fake_tag()
    supported_version = get_supported_version(tags)
    print(supported_version)
    create_file(sys.argv[1], supported_version)
```

After that this script can be either 
 - integrated in the CI of this repo (and therefore the output file will be store on github)
 - or run via the cron on our MG5aMC server (the file itself does not have to be store on the server it can part of this repo and just executed on the MG5aMC server --so will be in sync with the master repo (like other CI works)
    - in that case the script will be in the plugin but the output store on the MG5aMC servers

valassi:
Thanks @oliviermattelaer for the comments, very good idea :-)

I implemented the full chain this morning based on your script.

This is the dat file generated by the github CI
https://raw.githubusercontent.com/valassi/madgraph4gpu/refs/heads/INFO/version_info.dat

I created a protected *orphan* branch INFO in the madgraph4gpu repo where the file is stored

If you agree with the whole machinery, I can move this to madgraph4gpu (I am still using the valassi fork for tests).

Let me know!

Thanks, have a good weekend
Andrea

oliviermattelaer:
Great excellent, This is super nice.

Please go ahead and thanks a lot,

Olivier

oliviermattelaer (approved):
Perfect nothing to comment on.

But please can we do a squash here?

Cheers,

Olivier

valassi:
Hi @oliviermattelaer thanks!

> But please can we do a squash here?

Sorry, no. Amongst other things, there are some test results that are _only_ visible as individual commits (I then build all my summary tables for cnferences based on those). Eventually I/we need to find a better way, but so far, no.

And in any case I really prefer to see the full history, in case we need to go back and check some stuff. You might not believe it, but I am _already_ squashing a lot (there are here around 100 commits visible, but in reality these were maybe 400, debugging the CI was really complex, and I want to keep track of that debugging).

Also for HEPInstaller by the way I would really prefer no squashing. Same thing, at least I can easily find back if ever I want to bypass info.dat (not that I think this will ever be needed, but one never knows).

Can I merge like this then? Thanks!
Andrea

valassi:
PS Sorry I know I am difficult :-)

But this point easily gets very very very religious...!
https://dev.to/wesen/squash-commits-considered-harmful-ob1
Quote "A recurring conversation in developer circles is if you should use git --squash when merging or do explicit merge commits. The short answer: you shouldn't. People have strong opinions about this. The thing is that my opinion is the correct one. Squashing commits has no purpose other than losing information. It doesn't make for a cleaner history."

I am not saying "my opinion is the correct one", but definitely I do have a strong opinion that squashing is often bad. 

Then I do squash sometimes, my own commits. With other people's commits I would even be more reluctant to squash, I want to be able to trace back what they did, if/when needed...

valassi:
Thanks again @oliviermattelaer ! As discussed, merging now.

Then I will bump up the version in a later PR or commit

Andrea

PR #1016, opened 2024-10-05
Remove add_input_for_banner from output.py (plugin_run_card is not needed in cudacpp)

### LINKED ISSUE #1015: cudacpp_backend card should be removed from output.py (it is only needed in launch_plugin.py)
We should understand if cudacpp_backend card should be defined only in output.py or launch_plugin.py

This is a followup to #1009 (which was closed because it included also other things, now fixed by Olivier)

valassi:
Hi @oliviermattelaer this is a very simple PR to remove add_input_for_banner from output.py (plugin_run_card is not needed in cudacpp)

It fixes #1015. As described there:
- In output.py cudacpp_backend is only needed to define plugin_run_card
- In banner.py plugin_run_card is only used in RunCard.plugin_input
- HOWEVER in launch_plugin.py there is a specific CPPRunCard that overloads RunCard.plugin_input and DOES NOT USE plugin_run_card
- **THEREFORE plugin_run_card is not used at all in cudacpp and cudacpp_backend can be removed from output.py**

Can you please review? 

I need to get this out to then mak eprogress on other things for CHEP

Thanks Andrea


valassi:
Hi @oliviermattelaer as discussed via email.

This is N=1 in the pipeline of PRs I would like to merge. This is ready as is, should be merged to master. Please tell me if I can go ahead, thanks!

oliviermattelaer:
Ok, 

I do not like this PR ... I either see this has half full or half empty PR. (meaning it either due too much or too little).

So here are two counter proposal:
1) fully remove "plugin_run_card" mechanism from 3.6.1 and the associate "plugin_input" API, and see this as deprecated by the launch_plugin API (both API were designed for the cudacpp case so this will not break anything) 

2) keep this plugin_run_card completes as easy way to see new parameter (even if this is not used) and in that case add in it "floating_type".

> I need to get this out to then mak eprogress on other things for CHEP

With all due respect, this is NOT something needed for CHEP (since this is a pure cleaning of the code which has zero impact on anything, so this can be put in super low priority). So let's discuss this at our meeting today.

Cheers,

Olivier

valassi:
> Ok,
> 
> I do not like this PR ... I either see this has half full or half empty PR. (meaning it either due too much or too little).
> 
> So here are two counter proposal:
> 
>     1. fully remove "plugin_run_card" mechanism from 3.6.1 and the associate "plugin_input" API, and see this as deprecated by the launch_plugin API (both API were designed for the cudacpp case so this will not break anything)
> 
>     2. keep this plugin_run_card completes as easy way to see new parameter (even if this is not used) and in that case add in it "floating_type".
> 

Hi @oliviermattelaer thanks :-)

I would vote for option 1 then. I do not see the need for plugin_run_card since we have launch_plugin.py. I think you introduced first plugin_run_card and then launch_plugin.py, but I thought that maybe you had other plugins in mind too. If they are both for cudacpp, then I think that launch_plugin.py is enough, and we can remove plugin_run_card.

Can you do the removal of plugin_run_card from mg5amcnlo yourself, whenever you have time?

This is low priority I agree. The reason I put this first in my 'pipeline' is that the subsequent PRs also touch runcards, and I did not want to have to do things there in plugin_run_card  if I knew that this was not needed anyway.

By the way I still think that this PR as is makes sense, if we eventually remove plugin_run_card. It is "too little" because it only removes it from cudacpp and not from mg5amcnlo, but inside cudacpp it should be ok? Well I can COMPLETELY remove all mentioning of plugin_runcard in cudacpp, rather than leave it commented out. But otherwise I do not see what elese needs to change in cudacpp itself?

> > I need to get this out to then mak eprogress on other things for CHEP
> 
> With all due respect, this is NOT something needed for CHEP (since this is a pure cleaning of the code which has zero impact on anything, so this can be put in super low priority). So let's discuss this at our meeting today.

My point was that I will produce profiling numbers for CHEP and I wanted to do it if possible from stuff merged in master. But not a problem, I will do it from my branch and then we can merge in master later on also after CHEP.

By the way there is no dev meeting today, it's next week.
https://indico.cern.ch/event/1355163/

valassi:
> I do not like this PR ... I either see this has half full or half empty PR. (meaning it either due too much or too little).
> 
> So here are two counter proposal:
> 
>     1. fully remove "plugin_run_card" mechanism from 3.6.1 and the associate "plugin_input" API, and see this as deprecated by the launch_plugin API (both API were designed for the cudacpp case so this will not break anything)

I prepared a WIP PR
https://github.com/mg5amcnlo/mg5amcnlo/pull/149

Is this what you had in mind?
Thanks
Andrea

oliviermattelaer:
Yes I merged the associated PR

valassi:
> Yes I merged the associated PR

Hi @oliviermattelaer thanks :-)

I have modified this PR accordingly
- I updated the mg5amcnlo submodule with the upstream removal of plugin_input
- I have completely removed plugin_input from cudacpp (previously only commented out)
- I have regenerated patch.P1
- I have regenerated all processes

Is this ok for merging now or are there other things to fix/improve?

Thanks
Andrea

oliviermattelaer (approved):
Thanks, this is good to go.

Thanks,

Olivier

valassi:
Thanks to you :-)

Merging now

Issue #972, opened 2024-08-19
Faster timers based on rdtsc instead of chrono


valassi:
Faster timers based on rdtsc instead of chrono

I file this explcitly as an issue to make this clearer. This was quite a bit of work in the last few days.

I have completed the whole task
- the new timer.h (I kept the same name for simplicity, but it's a different beast) has a new chrono timer with a different API and a nanosecond tick granularity, but especially includes also a rdtcs timer
- the calibration of tcs ticks to time is done only ONCE at the end... so maybe the relative ticks are not very precise if frequency varies wildly during a job, but I think that's precise enough
- the new timers have lower overhead (not too much, but still useful) when calling them very often eg inside sample_get_x
- I have completed the new counters.cc an dthe new timermap.h to adapt to this new ticks based API with one calibration at the end

The changes are used in many PRs, but I would say that they can be merged from PR #962

valassi:
The code is essentially here
https://github.com/madgraph5/madgraph4gpu/blob/9a03440ebf8dfd8f9b7302747549c90b23ed5c01/epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/madgraph/iolibs/template_files/gpu/timer.h

Note, this was developed thanks to a lot of documentation by many people, including these and many others
- https://github.com/jdmccalpin/low-overhead-timers/tree/master
- https://github.com/MengRao/tscns

valassi:
(Note: the overhead of std chrono clocks is in any case much less than we had in the past due to O/S issues #116)

valassi:
See this commit in PR #970: https://github.com/madgraph5/madgraph4gpu/pull/970/commits/c863d69aee13f89f3f09d3da5e4aa005924ee822
```
CUDACPP_RUNTIME_DISABLEFPE=1 ./build.cuda_d_inl0_hrd0/madevent_cuda < /tmp/avalassi/input_dy3j_x1_cudacpp
 [COUNTERS] *** USING RDTSC-BASED TIMERS ***
 [COUNTERS] PROGRAM TOTAL                         :    3.9808s
 [COUNTERS] Fortran Other                  (  0 ) :    0.1248s
 [COUNTERS] Fortran Initialise(I/O)        (  1 ) :    0.0676s
 [COUNTERS] Fortran Random2Momenta         (  3 ) :    2.7899s for  1170103 events => throughput is 2.38E-06 events/s
[...]

PR #962, opened 2024-08-10
WIP further improve timers and profile other fortran components

valassi:
This is a very WIP PR extending the work in #960. Again related to the CMS #943 issue reported by @choij1589 

The idea is to further improve timers and profile other fortran components

I added profiling to
- random to momenta translation in phase spaces sampling ie x_to_f_args (X2F)
- nnevolvepdf (PDF)
- put_sample_point (I/O)

The first two are related to the findings with nice flamegraphs by @Qubitol 

So far I only had time on a simple gg_tt. Even here it is quite interesting
- sampling is actually not much (but it is there)
- PDF is a lot (even in ggtt, I thought it would only appear in pp)
- I/O is also potentially a lot

The problem is also that the pverhead from the counters themselves starts being important, so it is difficult to do it well. Especially the i/o counters have a large overhead.

Anyway, this is WIP. For ggtt
```
    ./build.none_d_inl0_hrd0/madevent_cpp < /tmp/avalassi/input_ggtt_x1_cudacpp
     [COUNTERS] PROGRAM TOTAL          :    0.7442s
     [COUNTERS] Fortran Overhead ( 0 ) :    0.2437s
     [COUNTERS] CudaCpp MEs      ( 2 ) :    0.0871s for    16384 events => throughput is 5.32E-06 events/s
     [COUNTERS] CudaCpp HEL      ( 3 ) :    0.0008s
     [COUNTERS] Fortran X2F      ( 4 ) :    0.0162s for    16399 events => throughput is 9.86E-07 events/s
     [COUNTERS] Fortran PDF      ( 5 ) :    0.1335s for    98304 events => throughput is 1.36E-06 events/s
     [COUNTERS] Fortran I/O      ( 6 ) :    0.2629s for    16399 events => throughput is 1.60E-05 events/s
    
    ./build.none_d_inl0_hrd0/madevent_cpp < /tmp/avalassi/input_ggtt_x10_cudacpp
     [COUNTERS] PROGRAM TOTAL          :    1.9099s
     [COUNTERS] Fortran Overhead ( 0 ) :    0.3233s
     [COUNTERS] CudaCpp MEs      ( 2 ) :    0.5203s for    98304 events => throughput is 5.29E-06 events/s
     [COUNTERS] CudaCpp HEL      ( 3 ) :    0.0007s
     [COUNTERS] Fortran X2F      ( 4 ) :    0.0956s for    98371 events => throughput is 9.71E-07 events/s
[...]

PR #1018, opened 2024-10-07
Faster RDTSC-based timers and new timer/counter APIs

valassi:
This is a PR including faster RDTSC-based timers and new timer/counter APIs. It completes #972.

This PR ("prof0") was derived from the pre-existing PR #962 ("prof"), by stripping off the second part (additional profiling of non-ME fortran components) and keeping only the first part (new RDTSC based timers and new APIs).

The idea is that the additional profiling of non-ME fortran components will be done at a later time in #962, but it will be modified to include patches in upstream mg5amcnlo as suggested by @oliviermattelaer , rather than relying on patchMad.sh with much larger patches, as is done presently.


valassi:
Hi @oliviermattelaer as discussed via email, this N=4 and the last PR in the pipeline that I would like to merge.

Again, I changed this to target N=3 for easier review, but then I would merge it to master once approved.

Let me know what you think please.. thanks!

Andrea

valassi:
For the moment I moved back to draft while working on the base PRs

PR #1030, opened 2025-04-24
fixed bug where precision wasn't passed properly from the run_card

zeniheisser:
In cudacpp.mk, cudacpp_config.mk is included before make_opts. When setting floating_type from the run_card, it is appended to make_opts, but since FPTYPE is set with the "override" directive in cudacpp_config the default value is automatically taken unless FPTYPE is given as a command line argument.

Circumventing this by adding "override" to the make_opts declaration.

Note: Not a perfect fix. Appending FPTYPE using common_run_interface.CommonRunCmd.update_make_opts_full, which adds the declaration to the TOP of make_opts, regardless of the existence of the same argument name further down in the file


oliviermattelaer:
Thanks,

I'm confused about this override can you add a comment within the code to explain it maybe?
(I feel ashamed since I likely be the one implementing such things)

As you I'm surprised about the test failing which seems unrelated to your change... Do you have an idea of why this is happening? (change of hardware? or of python version?)

PR #1031, opened 2025-05-07
Update c-cpp.yml workflow to run on NGT runners

Qubitol:
Add also `workflow_dispatch` trigger to be able to trigger it manually.

I created the pull request, so the workflow becomes available to test it.
Let's wait I can test it properly before merging.


roiser:
Great ! Please go ahead and merge !

PR #1038, opened 2025-08-08
Get rocm lib paths through generic hipconfig command

### LINKED ISSUE #1020: Library path to AMD libraries may not be correct in certain cases
Hi,

I've done some tests on a machine with an AMD GPU.
However, when running `make`, it fails when trying to look for `libamdhip64`.

### The problem in my case

See this line in `cudacpp.mk`:
```makefile
$(FC) -o $@ $(BUILDDIR)/fcheck_sa_fortran.o $(BUILDDIR)/fsampler_$(GPUSUFFIX).o $(LIBFLAGS) -lgfortran -L$(LIBDIR) -l$(MG5AMC_GPULIB) $(gpu_objects_exe) -lstdc++ -L$(shell cd -L $(shell dirname $(shell $(GPUCC) -print-prog-name=clang))/../..; pwd)/lib -lamdhip64
```
In my case `shell dirname $(shell $(GPUCC) -print-prog-name=clang)` yields `/opt/rocm-6.2.2/lib/llvm/bin`, then it goes up 2 levels and appends `/lib`, resulting in the final path `/opt/rocm-6.2.2/lib/lib`, which is not correct in my case (should be `/opt/rocm-6.2.2/lib`).

### Possible solution to make it dynamic

I found the command `hipconfig` that displays a bunch of information about the installation.
One could use then:
```bash
hipconfig --rocmpath
```
to get exactly `/opt/rocm-6.2.2`, and then append `/lib`.
Now, this would be the dynamic approach, but it requires `hipconfig` to be available. Is this the case for every HIP-based GPU? In such case, I can submit a PR fixing the makefiles.

Qubitol:
This fixes #1020 
Please, @roiser have a look at this, I tried [your fix](https://github.com/madgraph5/madgraph4gpu/pull/1034/files) using `HIP_HOME` but it seems it doesn't work in the machine I used (NGT cluster).


roiser (approved):
yes this looks more stable than HIPHOME, please go ahead

PR #1040, opened 2025-08-19
Remove patches and patchMad.sh

Qubitol:
Implement the patches already in the code:
- either already in upstream code (for `driver.f` and matrix file);
- or in CUDACPP code (for `makefile`).

For the `makefile` in particular, I used an overlay approach, in which I put the patches inside an additional `cudacpp_overlay.mk` file, and modified the compilation commands in this way:
```
make -f makefile -f cudacpp_overlay.mk ...
```
so that the second file imported would override any recipe defined in `makefile`.

See also related [mg5amcnlo#191](https://github.com/mg5amcnlo/mg5amcnlo/pull/191).


Qubitol:
Instead of implementing the additions in upstream, use code generation and string placeholders in `iolibs/template_files/madevent_driver.f`, so that the changes are managed by the plugin itself, through overriding of `write_driver` in `FortranExporterBridge`.
Placeholders are called: `CUDACPP_EXTRA_HEADER`, `CUDACPP_EXTRA_INITIALISE`, `CUDACPP_EXTRA_FINALISE`.

PR #1046, opened 2025-10-03
MadtRex implementation

zeniheisser:
Cleaned up MadtRex branch, for cudacpp integration equivalent to last one but now with refactored Rex and teaRex implementations


zeniheisser:
Fixed remaining issues wrt CI, merging

PR #1049, opened 2025-10-06
Kernel splitting ihel1/2/3: helicity streams, color sum kernel, color sum BLAS

valassi:
Hi @oliviermattelaer, as discussed recently and as per my presentation at the MG5AMC meeting last Friday.

This is the PR for my kernel splitting changes that I recommend merging:
- ihel1: helicity streams
- ihel2: color sum kernels
- ihel3: color sum in BLAS

I have prepared a paper that will be shortly in arxiv with all the details.

Until yesterday, it would have been possible to merge this automatically, as I had merged the latest upstream in my developments. Yesterday there were some new changes merged (for tREX I think), so this will need some massaging. I can do that later on, or let me know how you want to proceed.

Thanks, Andrea


oliviermattelaer:
Thanks a lot Andrea,

Conflict was not relevant (a comment) so I just fixed it (I guess at least).

So I will start the review!

Thanks a lot,

Olivier

epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/model_handling.py
```diff
@@ -1521,6 +1530,17 @@ def edit_processidfile(self):
...
         ff.close()
 
+    # AV - new method
+    def edit_colorsum(self):
+        """Generate color_sum.cc"""
+        ###misc.sprint('Entering PLUGIN_OneProcessExporter.edit_colorsum')
+        template = open(pjoin(self.template_path,'gpu','color_sum.cc'),'r').read()
+        replace_dict = {}
+        # Extract color matrix again (this was also in get_matrix_single_process called within get_all_sigmaKin_lines)
+        replace_dict['color_matrix_lines'] = self.get_color_matrix_lines(self.matrix_elements[0])
```
oliviermattelaer, line 1540: This is a comment to pint-point, the previous point of attention,
this is where we need to check if the format is still correct.
(if it is we likely want to create an issue to update it, if not then no choice than fixing it)

PR #1058, opened 2025-10-21
Add MadtRex tests to CI

Qubitol:
Add a new step in `testsuite_oneprocess` workflow to check MadtRex according to validation tests running the reweighting on MC samples generated using the process folders obtained during the `codegen` step in the CI itself.
Notice that the references files have been generated using vanilla MadGraph and not CUDACPP (like in this case) - except for `nobm_pp_ttW` that has been regenerated with CUDACPP - so we increased the maximum threshold for the results difference to 5%.


Qubitol:
After an exchange with @zeniheisser, we decided that it may be worth to generate the event samples in advance, so that tests go a bit faster. This would imply:
- generate and save the LHEF as some tests assets
- compile MadtRex
- running the executable with some cmd line args

This can be done in another workflow file separated from `testsuite_oneprocess.yml`, but still called by `testsuite_allprocesses.yml`.

zeniheisser:
would be nice to have to avoid something like https://github.com/madgraph5/madgraph4gpu/pull/1090 :))))

Qubitol:
Yes, thanks @zeniheisser for bringing this up, and thanks for fixing MadtRex interface after the latest PRs.

The situation is: given we are moving fast to MadGraph7, and given both the CI and the interface to MadtRex there is still not finalised, I wanted to wait and implement everything only once there.
However, I'm open to finalise this if needed.

zeniheisser (approved):

Qubitol:
Tests are passing.
There are two main points associate to this PR.

## The reweighting procedure needs LHE event files
In order to avoid to regenerate them during each CI (some processes take hours to finish), they have been pre-generated and saved into `epochX/cudacpp/CODEGEN/PLUGIN/CUDACPP_SA_OUTPUT/test/MadtRex_baseline`, together with the validation file containing the reweigthing results to be checked within the CI.
This means both files should be regenerated if needed, and the script `.github/workflows/generate_madtrex_assets.py` does exactly that.

## One of the main issues was the model used to generate the process
Given some of them are downloadable online and should be converted automatically to be usable within MadGraph (possible to do this by doing `set auto_convert_model T` in the main prompt), this should be done in the CI as well.
This gave origin to the step `init_madgraph_configuration_and_model` in `.github/workflows/testsuite_oneprocess.yml` where we run MadGraph for two reasons:
- to write the default configuration files
- to set model auto-conversion.

In this way, the model can then be reimported from the reweighting command, launched from the `bin/madevent.py` of each process.
This command failed multiple times, and I understood that the main issue seemed to be the bytecode compiled by Python and stored inside the `<mg5amcnlo_path>/models/<model>/__pycache__`.
If that folder is removed, then no more errors are present, the model is reimported, and then the reweight procedure doesn't have any issues into using it.

PR #1064, opened 2025-11-11
Fix arm support for Linux

Qubitol:
Apparently, arm was working only in case of Mac (where `uname -p` gives `arm`) and not in case of most Linux distributions (where `uname -p` is `aarch64`).
I fixed that: CUDACPP can now run on arm, with support also for neon (equivalent to `cppsse4` and picked up using the same backend name).
In the makefiles, support for higher vectorisation is not present, but this can be something that we could think of adding in the next future.

I added tests on Linux arm as well.
I needed to regenerate the processes to update the makefiles.


Qubitol:
Latest changes involves using `__ARM_NEON`, which is automatically checked by the compiler in case of ARM SIMD.

valassi (approved):
Hi @Qubitol thanks for all the work on this! The tests are now finally passing in the CI. I suggest that we merge this.

PR #1066, opened 2025-11-18
Kernel splitting ihel4-ihel6: Feynman diagram groups

### LINKED ISSUE #310: Split the sigmakin kernel into smaller kernels

### LINKED ISSUE #348: Reduce build times in CUDA and C++ for complex processes (split kernels and more)
This is a followup of #346: on ggttggg it is clear that build times start becoming very long again (20 minutes or more, mainly in CUDA, but also in clang/C++ the situation looks bad).

The issue is clearly related to inlining of FFV functions (hence to their templating in PR #328) and more generally to LTO/RDC/inlining optimizations over very large code bases (#229 et al).

Removing inlining by hand is an option, but small tests I have done in the past were really bad for performance.

The only viable solution is most likely splitting kernels (#310), not only for CUDA but also for C++. Once we have more than 1000 Feynman diagrams as in ggttggg, it makes no sense to do any optimizations across a single calculate_wavefunctions method with O(1k-10k) FFV calls. It looks better, even just for C++ and for build times, to split this into O(1k) functions, one per diagram.

valassi:
Hi @oliviermattelaer as we discussed, this is the PR that complete my work on kernel splitting.

It extends and replaces PR #1050 that I will now close. It is described in v2 of my paper that will be in arxiv in a couple of days.

Essentially this makes it possible to define Feynman diagram groups, in separate source code files, and launch them either as kernels (DCDIAG=0) or as device functions in a single kernel (DCDIAG=1). Its main interest is very complex processes. I was able to run gg->ttgggg (2->6) on CPU and GPU and gg->ttggggg (2->7) on CPU. For our standard candles like gg_ttggg (2->5) it has the same performance as the current master when generation is configured to have a single diagram group (e.g. 2000 diagrams per group).

With respect to PR #1050 it also contains a full implementation of CUDA graphs to orchestrate diagram kernels. But this is actually not very useful as the main problem with small diagram kernels is access to GPU global memory and not kernel launch overhead. This is also described in the upcoming paper. So to ease maintenance I can remove CUDA graphs if you prefer.

The diagram per group generation is now configured in the generate script through a hack. But this should probably be in the runcard, I'd like to discuss the details.

Concerning BLAS, this is still optional as in present master. Note that in gg>ttggg now BLAS is faster than kernels for FPTYPE=d,f and just as fast for m. But it is still slower for simpler processes like gg_tt, so I think it should stay optional.

Let's discuss when the arxiv is out, it will be easier to discuss the details. I keep this in WIP for now.

Thanks!
Andrea


valassi:
Hi Olivier, Daniele, I mark you as reviewers. Let me know if you want to discuss this. Thanks
Andrea

PR #1073, opened 2025-11-24
Tests and minor performance improvements for FPTYPE=m color sums on CPU

valassi:
WIP patch for issue #1072 

I moved deltaMEs to fptype2 from fptype, but I do not think this will improve the situation - probably fpvmerge is the culprit


valassi:
Hi Olivier, Daniele, I mark you as reviewers. Let me know if you want to discuss this. Thanks
Andrea

PR #1083, opened 2026-03-17
int overflow

roiser:
When pushing the gridsize beyond max(int) the code was failing in few places where we expect int, those have been moved to size_t now.


Qubitol (approved):
I'd say we try with 3.6.6, to check whether the tests are successful and then we are good to merge.

PR #1092, opened 2026-07-01
Regenerate subprocesses for MadGraph version v3.7.1

Qubitol:
Preparation for release v1.01.02.

Check that tests are passing.


Qubitol:
Tests fail because the model can not be downloaded apparently.
Will re-run the failed jobs.

